PR #99, opened 2014-08-06
Implement an intersection mapper class

andlaus:
this will become required (or at least very useful) when attempting to implement transmissibility multipliers for `Dune::CpGrid`. Note that the `IntersectionMapper` class is in no way specific to `Dune::CpGrid`.


atgeirr:
While lots of interesting topics have been discussed in this thread, this PR has not been merged yet. Assigning that decision to @dr-robertk since he is a Dune expert.

dr-robertk:
I have implemented Intersection id's for CpGrid and PolyhedralGrid in PR #183 as it will become part of the DUNE interface in DUNE 3.0. This PR can be closed.

PR #118, opened 2015-01-12
Fix constructor of PersistentContainer for DUNE-grid 2.3 and later.

dr-robertk:
This class does currently not compile with dune-grid-2.3.1 and later. This PR fixes this issue.

### DIFF
--- a/dune/grid/cpgrid/PersistentContainer.hpp
+++ b/dune/grid/cpgrid/PersistentContainer.hpp
@@ -23,13 +23,17 @@ namespace Dune
 #else
   template< class Data, class Allocator >
   class PersistentContainer< CpGrid, Data, Allocator >
-  : public PersistentContainerVector< CpGrid, 
+  : public PersistentContainerVector< CpGrid,
                                       typename CpGrid::Traits::LeafIndexSet,
                                       std::vector<Data,Allocator> >
 #endif
   {
   public:
     typedef CpGrid  GridType;
+#if DUNE_VERSION_NEWER(DUNE_GRID, 2, 3)
+    typedef typename std::vector<Data>::allocator_type Allocator;
+#endif
+
   private:
 #if DUNE_VERSION_NEWER(DUNE_GRID, 2, 3)
     typedef PersistentContainerVector< GridType, typename GridType::Traits::LeafIndexSet, std::vector<Data> > BaseType;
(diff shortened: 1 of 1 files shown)


bska:
That's strange.  I've been using this code with Dune grid 2.3.1 for quite some time and I have never seen a build failure (in this part).  Have we not been exercising the branch?

dr-robertk:
Have you used PersistentContainer explicitly? When you compare the code you can already see that this cannot compile.

bska:
> Have you used `PersistentContainer` explicitly?

No, I have not.  That's why I'm asking if we've ever exercised that branch or if we've been shipping broken software since commit 4bf7be6 (PR #42).

dr-robertk:
Other than that everything at least compiles.

bska:
> Other than that everything at least compiles.

Sorry, I don't understand.  Could you add a bit of context to the statement?

dr-robertk:
I mean that besides the issue mentioned above I have not experienced other problems. 
Also, PersistentContainer is a feature that is not used all that much for some reason. That is why this has not been discovered yet. It can be used to attach arbitrary data to the grid. In case of CpGrid is simply used the IndexSet and a std::vector. For other grids this might be different since the IndexSet is not persistent (in terms of grid change).

bska:
> I mean that besides the issue [of this PR] I have not experienced other problems.

So, the practical upshot of this discussion is that we have indeed not been exercising the particular code in mainline dune-cornerpoint and have been shipping broken software since August 2013.  As I was the one to merge PR #42 which introduced this branch interface, I'll have to accept responsibility for missing code review.

dr-robertk:
If you want to put it that way, yes. On the other hand nobody seemed to bother, so no big deal. Anyway, this patch fixes the issue.

bska:
> nobody seemed to bother, so no big deal.

I suppose.  On the other hand, the `1.1 -> Data()` transformation could possibly have been made in a separate commit since that fixes a separate issue (namely generalising the `PersistentContainer` to handle data types other than `double`).

bska:
Oh, by the way, the removal of the constant `1.1` in the `# !DUNE_VERSION_NEWER` branch is wrong.  The Dune 2.2 `PersistentContainerVector` constructor interface is
```C++
PersistentContainerVector( const GridType& grid, const int codim,
                           const Index& index,
                           const double overEstimate,
                           const typename Vector::allocator_type &allocator );
```
so the third parameter must be a `double` rather than a `LeafIndexSet`.

dr-robertk:
Ops. That was an accidental change. Thanks for double checking.

dr-robertk:
Fixed and rebased to make my mistake disappear... ;-)

bska:
> Fixed and rebased to make my mistake disappear... ;-)

Cool.  I've run a few build tests just to make sure that we don't break anything.  I'll merge into master.  Thanks for diagnosing the problem and producing a fix.

dr-robertk:
Thanks

PR #119, opened 2015-01-12
[bugfix] Only access face_to_cell information when it exists.

blattms:
Due to a missing a check it sometimes happened that information
about a second cell attached to the face accesses even though there
was only one cell there. This commit fixes this.


bska:
I took me a while to recall why we need this precaution but now I see.  The `face_to_cell_` mapping contains only the active cell in the case of a boundary connection.

PR #120, opened 2015-01-13
Adds access to the face tag.

blattms:
When calculating transmissibilities we need to access the cartesian face tag
(I-, I+, J-, J+, K-, K+). While this possible (up to a constant) using
boundaryId when no unique boundary ids are used, we add a method that returns
the same value as for UnstructuredGrid under any circumstances.

Note that boundaryId still behaves the same as before!

### DIFF
--- a/dune/grid/CpGrid.hpp
+++ b/dune/grid/CpGrid.hpp
@@ -832,35 +832,40 @@ namespace Dune
         // Extra
         int boundaryId(int face) const
         {
-            int ret = 0;
             cpgrid::EntityRep<1> f(face, true);
             if (current_view_data_->face_to_cell_[f].size() == 1) {
                 if (current_view_data_->uniqueBoundaryIds()) {
                     // Use the unique boundary ids.
-                    ret = current_view_data_->unique_boundary_ids_[f];
+                    return current_view_data_->unique_boundary_ids_[f];
                 } else {
                     // Use the face tag based ids, i.e. 1-6 for i-, i+, j-, j+, k-, k+.
-                    const bool normal_is_in = 
-                        !(current_view_data_->face_to_cell_[f][0].orientation());
-                    enum face_tag tag = current_view_data_->face_tag_[f];
-                    switch (tag) {
-                    case LEFT:
-                        //                   LEFT : RIGHT
-                        ret = normal_is_in ? 1    : 2; // min(I) : max(I)
-                        break;
-                    case BACK:
-                        //                   BACK : FRONT
-                        ret = normal_is_in ? 3    : 4; // min(J) : max(J)
-                        break;
-                    case TOP:
-                        // Note: TOP at min(K) as 'z' measures *depth*.
-                        //                   TOP  : BOTTOM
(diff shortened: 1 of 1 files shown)


bska:
I'm afraid it isn't quite this simple.  The input to method `boundaryId()` is a face index.  In other words, `face` is an integer in the range `[0 .. numFaces() - 1]` and it is essential that it be a face index in order to have a defined meaning of the expression
```C++
current_view_data_->face_to_cell_[f].size()
```
Consequently, the proposed implementation of `faceTag()` cannot use an input index that is anything other than a face index.  On the other hand, the face tag in this context is really a property of a **pair** of indices, in particular one cell index and one face index.  Specifically, the tag in this context denotes the cardinal direction of a face as viewed from a particular cell.

Therefore, the proposed `faceTag()` should probably be implemented something along the lines of
```C++
int
faceTag(const int cell, const int face) const
{
    assert (0 <= cell);  assert (cell < numCells());
    assert (0 <= face);  assert (face < numFaces());

    typedef cpgrid::OrientedEntityTable<1,0>::row_type F2C;

    const cpgrid::EntityRep<1> f(face, true);
    const F2C&     f2c = current_view_data->face_to_cell_[f];
    const face_tag tag = current_view_data->face_tag_[f];

    assert ((f2c.size() == 1) || (f2c.size() == 2));

    const bool normal_is_in =
        ((f2c.size() == 1)
         ? ! f2c[0].orientation() // Single cell => boundary
         : (f2c[0] != cell));     // Two cells => interior

    switch (tag) {
    case LEFT:
        //                    LEFT : RIGHT
        return normal_is_in ? 0    : 1; // min(I) : max(I)

    case BACK:
        //                    BACK : FRONT
        return normal_is_in ? 1    : 2; // min(J) : max(J)

    case TOP:
        // Note: TOP at min(K) as 'z' measures *depth*.
        //                    TOP  : BOTTOM
        return normal_is_in ? 4    : 5; // min(K) : max(K)
    }
}
```

Moreover, when you remove the `ret` quantity from `boundaryId()` you also make the function undefined for non-boundary `face`s.  At the very least you should handle non-boundary face input index in some way.  We used to return zero and the following alternative implementation continues that tradition.
```C++
int
boundaryId(const int face) const
{
    assert (0 <= face);  assert (face < numFaces());

    typedef cpgrid::OrientedEntityTable<1,0>::row_type F2C;

    const cpgrid::EntityRep<1> f(face, true);
    const F2C& f2c = current_view_data->face_to_cell_[f];

    int ret = 0;  // Zero for non-boundary connections
    if (f2c.size() == 1) {
        if (uniqueBoundaryIds()) {
            ret = current_view_data_->unique_boundary_ids_[f];
        }
        else {
            // Use Cardinal directions
            ret = faceTag(f2c[0], face) + 1;
        }
    }
    
    return ret;
}       
```
The final point is minor.  I'd prefer if the colons were aligned vertically in `faceTag()`.  The comments appear to have been shifted one character position left during the refactoring.

blattms:
Thanks. That is indeed an over-enthusiastic oversight. I will into it again.

blattms:
Closing as superseded

PR #121, opened 2015-01-13
Adds access to the face tag

blattms:
When calculating transmissibilities we need to access the cartesian face tag (I-, I+, J-, J+, K-, K+). This is done by a new method faceTag that takes an iterator over the cell faces as its argument. It is used to deduce the cell and face index and then compute the value as proposed by Bard.

This PR is needed to implement similar functionality to OPM/opm-core#720 for CpGrid in opm-autodiff.
A PR for opm-autodiff follows closely.

Supersedes #120.

### DIFF
--- a/dune/grid/CpGrid.hpp
+++ b/dune/grid/CpGrid.hpp
@@ -864,6 +864,12 @@ namespace Dune
             return ret;
         }
 
+        /// \brief Get the cartesian tag associated with a face tag.
+        ///
+        /// The tag tells us in which direction the face would point
+        /// in the underlying cartesian grid.
+        /// \param An iterator that points to the face and was obtained
+        /// by iterating over Opm::UgGridHelpers::cell2Faces(grid).
         template<class Cell2FacesRowIterator>
         int
         faceTag(const Cell2FacesRowIterator& cell_face) const
--- a/dune/grid/CpGrid.hpp
+++ b/dune/grid/CpGrid.hpp
@@ -835,6 +835,11 @@ namespace Dune
         // Extra
         int boundaryId(int face) const
         {
+            // Note that this relies on the following implementation detail:
+            // The grid is always construct such that the faces where
+            // orientation() returns true are oriented along the positive IJK
+            // direction. Oriented means that the first cell attached to face
+            // has the lower index.
             int ret = 0;
             cpgrid::EntityRep<1> f(face, true);
             if (current_view_data_->face_to_cell_[f].size() == 1) {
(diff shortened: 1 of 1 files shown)


dune/grid/CpGrid.hpp
```diff
@@ -863,6 +864,46 @@ namespace Dune
...
+            case BACK:
+                //                    BACK : FRONT
+                return normal_is_in ? 1    : 2; // min(J) : max(J)
+
+            case TOP:
+                // Note: TOP at min(K) as 'z' measures *depth*.
+                //                    TOP  : BOTTOM
+                return normal_is_in ? 4    : 5; // min(K) : max(K)
+            default:
+                OPM_THROW(std::runtime_error, "Unhandeled face tag. This should never happen!");
```
atgeirr: Minor typo: Unhandled. Also, I would call this a logic_error, not a runtime_error, since it will always be due to programmer and not user/input error.

blattms: Thanks for the reasonable comments. I have incorporated them and
added some documentation to the method.

blattms:
I will fixup the last commit with the first one (the error is just too embarrassing), please tell to do this before the merge.

bska:
> please tell to do this before the merge.

Just do a `rebase -i` on top of master to squash the commits.  Let me know when it's done then I'll fire off a build and run test before merging.

blattms:
Rebased and ready for testing and merging.

dune/grid/CpGrid.hpp
```diff
@@ -863,6 +864,52 @@ namespace Dune
...
+
+            const bool normal_is_in =
+                ((f2c.size() == 1)
+                 ? ! f2c[0].orientation() // Single cell => boundary
+                 : (f2c[0].index() != cell));     // Two cells => interior
+
+            switch (tag) {
+            case LEFT:
+                //                    LEFT : RIGHT
+                return normal_is_in ? 0    : 1; // min(I) : max(I)
```
atgeirr: These statements depend on an implementation detail (that is also used in the boundaryId() method): that all faces are created oriented in the positive IJK directions. The same is done by the code that creates face tags for UnstructuredGrid (cornerpoint_grid.c of opm-core), but there is a comment there about this.

I think this PR is fine otherwise, but I'd like to see a comment about this to avoid future developers misunderstanding things.

atgeirr: Also, at the same time please add the same comment to boundaryId() to avoid similar future bugs there.

blattms: Added.

blattms: ditto for boundaryId.

atgeirr:
Thanks a lot!

PR #122, opened 2015-01-19
Adapt to API change for MinpvProcessor.

atgeirr:
Depends on OPM/opm-core#725.


bska:
Thanks.  I'll merge this into master.

PR #123, opened 2015-01-29
Provide access to the underlying parallel index set and remote indices.

blattms:
For this we now explicitly save the remote index information. This change
is needed to easily setup the parallel information needed by the dune-istl
solvers.


atgeirr:
Looks fine, will merge.

PR #124, opened 2015-02-05
CpGrid: Use shared_ptr to store the grid views and allow shallow copies

blattms:
In sim_fibo_ad_cp we need to access the global and the distributed
view at once. This can be done most elegantly by making a shallow
copy of the CpGrid (both use the same pointers to the underlying views)
and switching the view there. Therefore this commit uses shared_ptr to
hold the global and distributed view. Thus we do not even need a destructor.

### DIFF
--- a/dune/grid/CpGrid.hpp
+++ b/dune/grid/CpGrid.hpp
@@ -204,8 +204,6 @@ namespace Dune
 
 	/// Default constructor
 	CpGrid();
-
-        ~CpGrid();
  
         /// Initialize the grid from parameters.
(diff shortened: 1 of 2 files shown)


atgeirr:
Looks good, will merge. There is just one fly in the ointment: CpGrid used to be a value-semantics class, but that is no longer true. I guess this happened when the pointers changed here were first introduced, but I did not realize the subtle change of semantics at the time. I guess it's too late though, or would it be easy to write a proper (deep) copy constructor?

Another minor question: was there a particular reason to not use a smart pointer in the first place, or is it just a matter of habit?

PR #125, opened 2015-02-05
Allow for adding multiple overlap layers during loadbalance.

blattms:
Surpsingly an overlap layer of one cell does not seem to be
sufficient for opm-autodiff. Therefore with this commit we allow
multiple overlap layers and use 2 as the safe default.

### DIFF
--- a/dune/grid/CpGrid.hpp
+++ b/dune/grid/CpGrid.hpp
@@ -560,8 +560,9 @@ namespace Dune
         // loadbalance is not part of the grid interface therefore we skip it.
 
         /// \brief Distributes this grid over the available nodes in a distributed machine
+        /// \param The number of layers of cells of the overlap region (default: 1).
         /// \warning May only be called once.
-        bool loadBalance(int overlapLayers=2)
+        bool loadBalance(int overlapLayers=1)
         {
             return scatterGrid(overlapLayers);
         }
--- a/dune/grid/CpGrid.hpp
+++ b/dune/grid/CpGrid.hpp
@@ -569,10 +569,12 @@ namespace Dune
         
         /// \brief Distributes this grid and data over the available nodes in a distributed machine.
         /// \param data A data handle describing how to distribute attached data.
+        /// \param overlapLayers The number of layers of overlap cells to be added
+        ///        (default: 1)
         /// \tparam DataHandle The type implementing DUNE's DataHandle interface.
         /// \warning May only be called once.
         template<class DataHandle>
-        bool loadBalance(DataHandle& data, int overlapLayers=2)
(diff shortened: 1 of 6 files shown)


blattms:
Depending on whether #125 or #124 will be merged first, I will need to rebase the other one. Otherwise there will be conflicts.

atgeirr:
That is surprising indeed. Can you explain why?

blattms:
No, currently not. I wish I could.

atgeirr:
Referring to our discussion in the hangout, I'll hold off merging this.

blattms:
Rebased on top of master and changed the default number of layers of overlap cells. It is now 1 as expected. Nevertheless one can use more if necessary with this PR

dune/grid/CpGrid.hpp
```diff
@@ -571,9 +572,9 @@ namespace Dune
         /// \tparam DataHandle The type implementing DUNE's DataHandle interface.
         /// \warning May only be called once.
         template<class DataHandle>
-        bool loadBalance(DataHandle& data)
+        bool loadBalance(DataHandle& data, int overlapLayers=2)
```
atgeirr: Different default from the other overload is bad. Should be 1?

dune/grid/common/GridPartitioning.cpp
```diff
@@ -234,12 +234,78 @@ namespace Dune
 	ensureConnectedPartitions(grid, num_part, cell_part, recursive);
     }
 
-    void addOverlapLayer(const CpGrid& grid, const std::vector<int>& cell_part, 
-                         std::vector<std::set<int> >& cell_overlap, int mypart, bool all)
+/// \brief Adds cells to the overlap that just share a point with an owner cell.
```
atgeirr: I am a little skeptical about adding corner-neighbours. I would assume that only face-neighbours are needed? On the other hand I guess it will not inflict any harm other than a small amount of extra computation, and it may make it easier to try other numerical schemes than FV later on. Are there Dune interface requirements that makes it necessary to include the corner-neighbours?

blattms:
While you do not need them for FV you will need them for e.g. bilinear finite elements. (Or for your eikonal-whatever-stuff).

dune/grid/CpGrid.hpp
blattms: > Different default from the other overload is bad. Should be 1?

Yes, it should. Just fixed it  and added some documentation for the
parameter.

atgeirr:
> While you do not need them for FV you will need them for e.g. bilinear finite elements. (Or for your eikonal-whatever-stuff).

I am fine with this argument. I was just curious if the Dune interface required this.

blattms:
Thanks for merging.

PR #126, opened 2015-02-12
Adapt to minpv api changes of opm-parser.

qilicun:
see https://github.com/OPM/opm-parser/pull/416/


atgeirr:
As with the companion PR in opm-core, @joakim-hove you have license to merge.

PR #128, opened 2015-02-20
Ensure stable topology ops

### LINKED ISSUE #83: dune-cornerpoint can't load the Norne grid
Before #81 this was hidden by the fact that EclipseGridParser could not load the deck:

```
and@heuristix:~/src/ewoms|master > cat tutorial/tutorial1.cc
#include "config.h"
#include <dune/grid/CpGrid.hpp>
#include <opm/parser/eclipse/Parser/Parser.hpp>
#include <opm/parser/eclipse/Deck/Deck.hpp>
#include <dune/common/mpihelper.hh>

int main(int argc, char **argv)
{
    Opm::ParserPtr parser(new Opm::Parser);
    Opm::DeckConstPtr deck(parser->parseFile(argv[1]));

    Dune::CpGrid grid;
    grid.processEclipseFormat(deck, /*zTolerance=*/0, /*isPeriodic=*/false);

    return 0;
}
and@heuristix:~/src/ewoms|master > ./bin/tutorial1 /home/and/Norne/Norne_ATW_2013/NORNE_ATW2013.DATA
tutorial1: /home/and/src/dune-cornerpoint/dune/grid/cpgrid/readEclipseFormat.cpp:631: void Dune::<anonymous namespace>::buildTopo(const processed_grid &, std::vector<int> &, cpgrid::OrientedEntityTable<0, 1> &, cpgrid::OrientedEntityTable<1, 0> &, Opm::SparseTable<int> &, std::vector<array<int, 8> > &, std::vector<int> &): Assertion `f2c == f2c_again' failed.
[heuristix:26953] *** Process received signal ***
[heuristix:26953] Signal: Aborted (6)
[heuristix:26953] Signal code:  (-6)
[heuristix:26953] [ 0] /lib64/libc.so.6(+0x358d0) [0x7f2d04b2a8d0]
[heuristix:26953] [ 1] /lib64/libc.so.6(gsignal+0x39) [0x7f2d04b2a849]
[heuristix:26953] [ 2] /lib64/libc.so.6(abort+0x148) [0x7f2d04b2bcd8]
[heuristix:26953] [ 3] /lib64/libc.so.6(+0x2e616) [0x7f2d04b23616]
[heuristix:26953] [ 4] /lib64/libc.so.6(+0x2e6c2) [0x7f2d04b236c2]
[heuristix:26953] [ 5] ./bin/tutorial1() [0x4d8045]
[heuristix:26953] [ 6] ./bin/tutorial1(_ZN4Dune6cpgrid10CpGridData20processEclipseFormatERK6grdecldbb+0xf7) [0x4d7477]
[heuristix:26953] [ 7] ./bin/tutorial1(_ZN4Dune6cpgrid10CpGridData20processEclipseFormatESt10shared_ptrIKN3Opm4DeckEEdbbb+0x122f) [0x4d61ff]
[heuristix:26953] [ 8] ./bin/tutorial1(_ZN4Dune6CpGrid20processEclipseFormatESt10shared_ptrIKN3Opm4DeckEEdbbb+0x88) [0x4cca38]
[heuristix:26953] [ 9] ./bin/tutorial1(main+0x174) [0x4c3674]
[heuristix:26953] [10] /lib64/libc.so.6(__libc_start_main+0xf5) [0x7f2d04b16be5]
[heuristix:26953] [11] ./bin/tutorial1() [0x4c3425]
[heuristix:26953] *** End of error message ***
Aborted
```

atgeirr:
Sorting faces' cell neighbours by index ensures that the makeInverseRelation() function yields stable results, i.e. inverting twice yields the same table back.

The code without this PR is not wrong, and yields equivalent but not identical topology tables. The cost of the extra sorting is less than 0.1 second on the Norne grid.

This fixes #83.

### DIFF
--- a/dune/grid/cpgrid/OrientedEntityTable.hpp
+++ b/dune/grid/cpgrid/OrientedEntityTable.hpp
@@ -73,6 +73,8 @@ namespace Dune
 	    }
 	    using R::size;
 	    using R::empty;
+	    using R::begin;
+	    using R::end;
 	    /// @brief Random access operator.
 	    /// @param subindex Column index.
 	    /// @return Entity representation.
@@ -160,6 +162,36 @@ namespace Dune
 	    bool operator==(const OrientedEntityTable& other) const
 	    {
 		return super_t::operator==(other);
+	    }
+
+            /** @brief Prints the relation matrix corresponding to the table, sparse format.
+
+             Let the entities of codimensions f and t be given by
+             the sets \f$E^f = { e^f_i } \f$ and \f$E^t = { e^t_j }\f$.
+             A relation matrix R is defined by
+	     \f{eqnarray*}{
+	       R_{ij} &=& 0  \mbox{ if } e^f_i \mbox{ and } e^t_j \mbox{ are not neighbours }\\
+               R_{ij} &=& 1  \mbox{ if they are neighbours with same orientation }\\
+	       R_{ij} &=& -1 \mbox{ if they are neighbours with opposite orientation.}
+	     \f}
+             The output is written one entry to each line, in the format:
+                 row   column    entry (either 1 or -1)
+             The row and column numbers start from zero, so if using octave or
+             matlab you should add 1 to those columns after loading, before calling spconvert().
+
+	     @param os   The output stream.
+	    */
+	    void printSparseRelationMatrix(std::ostream& os) const
+	    {
+		for (int i = 0; i < size(); ++i) {
+		    const FromType from_ent(i, true);
+		    const row_type r = operator[](from_ent);
+                    const int rsize = r.size();
+                    for (int j = 0; j < rsize; ++j) {
+                        os << i << ' ' << r[j].index() << ' ' << (r[j].orientation() ? 1 : -1) << '\n';
+                    }
+		}
+                os << std::flush;
 	    }
 
  	    /** @brief Prints the relation matrix corresponding to the table.
--- a/dune/grid/cpgrid/OrientedEntityTable.hpp
+++ b/dune/grid/cpgrid/OrientedEntityTable.hpp
@@ -96,8 +96,8 @@ namespace Dune
 	/// The purpose of this class is to hide the intricacies of
 	/// handling orientations from the client code, otherwise a
 	/// straight Opm::SparseTable would do.
-	/// @tparam codim_from Codimension of ???
-	/// @tparam codim_to Codimension of ???
+	/// @tparam codim_from Codimension of domain of relation mapping
+	/// @tparam codim_to Codimension of range of relation mapping
 	template <int codim_from, int codim_to>
 	class OrientedEntityTable : private Opm::SparseTable< EntityRep<codim_to> >
 	{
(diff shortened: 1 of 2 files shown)


blattms:
After merging will the ordering be the same as for UnstructuredGrid?

atgeirr:
> After merging will the ordering be the same as for UnstructuredGrid?

Good point. It depends on what you mean: cell and face orders are unchanged, only the order of the (typically two) cell-neighbours of some of the faces. Since UnstructuredGrid (or to be specific, the create_grid_cornerpoint() function) just grabs the pointer from the output of process_grdecl() that mapping (face->cell) will be different (but equivalent). I am pretty sure that the cell->face mapping becomes identical between the two though.

dune/grid/cpgrid/OrientedEntityTable.hpp
```diff
@@ -162,6 +164,36 @@ namespace Dune
 		return super_t::operator==(other);
 	    }
 
+            /** @brief Prints the relation matrix corresponding to the table, sparse format.
+
+             Let the entities of codimensions f and t be given by
+             the sets \f$E^f = { e^f_i } \f$ and \f$E^t = { e^t_j }\f$.
+             A relation matrix R is defined by
+	     \f{eqnarray*}{
```
bska: Does Doxygen not support anything more well-behaved than `eqnarray`?

If I were writing this in a regular document I'd use something like
```latex
\usepackage{amsmath}
% ...
\begin{document}
% ...
\[
R_{ij} =
\begin{cases}
  \phantom{-} 0, &\quad \text{if } e_i^f \text{ and } e_j^f \text{ are not neighbours}, \\
  \phantom{-} 1, &\quad \text{if } e_i^f \text{ and } e_j^f \text{ are neighbours with same orientation}, \\
           -  1, &\quad \text{if } e_i^f \text{ and } e_j^f \text{ are neighbours with opposite orientation}.
\end{cases}
\]
```
Is that not possible in Doxygen?

atgeirr: Probably, but I do not know for sure. The doc in question is partially copied from the pre-existing non-sparse version.

bska: > Probably, but I do not know for sure

I don't know what the "probably" refers to, so I can't reply intelligently.  However, our [`cmake/Templates/Doxyfile`](https://github.com/OPM/dune-cornerpoint/blob/master/cmake/Templates/Doxyfile#L1147) does include the request
```
# The EXTRA_PACKAGES tag can be to specify one or more names of LaTeX
# packages that should be included in the LaTeX output.

EXTRA_PACKAGES         = amsmath
```
so using the previously suggested `cases` construction should be possible.  Would you mind trying that?

In general `eqnarray` should not be used at all in any documents, but it seems impossible to eliminate due to old habits.

atgeirr:
I have improved docs as requested. @blattms can you tell us if this would negatively affect your work in any way? Otherwise I'd like to have this merged.

blattms:
On Mon, Mar 02, 2015 at 02:07:31AM -0800, @atgeirr wrote:
> I have improved docs as requested. @blattms can you tell us if this would negatively affect your work in any way? Otherwise I'd like to have this merged.

Merging should be fine.

bska:
> Merging should be fine.

While I'd prefer that we didn't introduce any **new** tab-vs-space issues into this file (there are already quite enough), I'll just do a final build test and merge if it goes through.

bska:
Builds and all tests run on GCC 4.4/CentOS 5.11.  I'll merge into master.

PR #129, opened 2015-02-23
Moves Opm::UgGridHelpers from opm-autodiff to dune-cornerpoint.

blattms:
For historic (or no apparent) reason the free function grid interface
was added to opm-autodiff. As it depends on whether or not dune-cornerpoint
is present this seems rather unnatural. Therefore this commit adds the
functionality unconditionally to dune-cornerpoint.

While doing that we changed the namespace of the helper classes from
Opm::AutoDiff to Dune::cpgrid.

Note that there is a corresponding commit in opm-autodiff.

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -32,6 +32,7 @@ list (APPEND MAIN_SOURCE_FILES
 	dune/grid/cpgrid/Intersection.cpp
 	dune/grid/cpgrid/CpGridData.cpp
 	dune/grid/cpgrid/CpGrid.cpp
+	dune/grid/cpgrid/GridHelpers.cpp
 	dune/grid/cpgrid/PartitionTypeIndicator.cpp
 	dune/grid/cpgrid/readEclipseFormat.cpp
 	dune/grid/cpgrid/readSintefLegacyFormat.cpp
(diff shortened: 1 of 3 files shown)


atgeirr:
Thanks, I agree that it should be moved here. Is the Dune::cpgrid namespace new? I see that some things are in Opm::UgGridHelpers, that is the namespace you are supposed to use when writing code against this interface right?

blattms:
It is not new. It has been used for internal stuff (entity, etc.) of CpGrid. Therefore it seemed natural to us it. In opm-autiff this was in Opm::AutoDiffGrid.

Opm::UgGridHelpers is where the free grid function interface resides. Yes, that one should be used by users writing generic grid algorithms that should work with both CpGrid and UnstructuredGrid.

blattms:
BTW there are still some functions in Opm::AutoDiffGrid that do not rely on Eigen. I think these should be moved to Opm::UgGridHelpers (either opm-core or dune-cornerpoint) in separate PRs. The long term goal is to habe only the stuff that depends on Eigen in opm-autodiff

atgeirr:
Thanks, merging now. Sorry for delay, but found it safer to merge just after than just before the weekend.

PR #134, opened 2015-04-10
Sync build system

atgeirr:
This duplicates all build system changes from opm-core. Hopefully the last time we have to do this.


bska:

PR #136, opened 2015-04-20
Cleanup and fix communication interface

blattms:
This bundles a bug-fix and some cleanups in the communication interface. The bug is also fixed upstream and we just work around it in OPM.

I guess this could also go to the release.

### DIFF
--- a/dune/grid/CpGrid.hpp
+++ b/dune/grid/CpGrid.hpp
@@ -1037,6 +1037,16 @@ namespace Dune
             return current_view_data_->cell_remote_indices_;
         }
 
+        const ParallelIndexSet& getCellIndexSet() const
+        {
+            return current_view_data_->cell_indexset_;
+        }
+
+        const RemoteIndices& getCellRemoteIndices() const
+        {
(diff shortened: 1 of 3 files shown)


atgeirr:
I'd prefer to wait merging this until after the release (which I now think will actually happen this week), since I'm not too comfortable reviewing the parallel parts (I do not currently build with MPI). Is that OK? I think that the parallel support in Flow is very experimental for this release, but it could be argued that the CpGrid class' parallel support is not.

bska:
While you're doing clean-up I'd like to know if there a technical reason for writing
```C++
if (container.size() == 0)
```
rather than
```C++
if (container.empty())
```
I personally find the latter more to the point and easier to read than the former.

blattms:
Sure! Do as you please.

blattms:
I guess my bad habits die late. You are right of course that empty()
is much better. I changed it. Thanks for the hint!

atgeirr:
Since we already had a bugfix that needed to go in, I'll merge this as well and update the release branch.

PR #137, opened 2015-04-20
[bugfix] Use resize to change the size of cell_to_point_.

blattms:
Previously we just used reserve, but never actually changed the
size. Nevertheless we used operater[] to set and get values up to index
of the reserved size-1. Semantically this is a bug that fortunately
never results in a memory problem. Nevertheless g++ detects this
problem whith -D_GLIBCXX_DEBUG and aborts.

With this commit we now use resize() to change the size. This should definitely
go into the release.


bska:
Thanks!

I'll merge into master, and let @atgeirr handle the merge to the release branch.

PR #139, opened 2015-04-24
Adds support for initial loadbalancing using ZOLTAN

blattms:
This PR now uses ZOLTAN to repartition the grid if it is available. If not, we fall back to the old behaviour. 
It uses the grid information to build the corresponding graph infrastructure on a root process, then applies the parallel graph partitioning of ZOLTAN to it and  broadcasts the information about where each cell will be owned to all processes. According to this the parallel load balanced grid is set up by the old routines.

Note: This PR incorporates OPM/opm-core#787. There it should only be merged after propagating OPM/opm-core#787 to the other modules and rebasing this PR.

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -40,6 +40,7 @@ list (APPEND MAIN_SOURCE_FILES
 	dune/grid/common/GeometryHelpers.cpp
 	dune/grid/common/GridPartitioning.cpp
 	dune/grid/common/ZoltanGraphFunctions.cpp
+	dune/grid/common/ZoltanPartition.cpp
 	)
 
 # originally generated with the command:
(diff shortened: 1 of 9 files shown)


blattms:
With OPM/opm-cmake#6 merged we now have a working cmake find module for ZOLTAN and I have rebased this branch on top of the current master.

atgeirr:
I am not familiar enough with Zoltan to review this properly, so I leave it to @dr-robertk or @bska to review. (If you do not know it either I propose just merging this and verifying that it does not hurt anything at least.)

dr-robertk:
Looks good and seems to work.

atgeirr:
Thanks for the patch @blattms and thanks for reviewing @dr-robertk.

PR #143, opened 2015-05-22
Removes leftover cmake modules

blattms:
Somehow @joakim-hove must have missed FindZoltan and FindPETSc in PR #132


joakim-hove:
Thanks - merging.

PR #149, opened 2015-06-17
move MoveBuffer to namespace mover to avoid warning.

dr-robertk:
This PR removes a warning about an anonymous namespace.


atgeirr:
Looks fine, will merge. I also notice that you make some inheritance relationships public. I must admit I never noticed the private inheritance in the first place, but think that in most cases it is better to avoid it, so that is good.

PR #151, opened 2015-07-03
Polyhedral grid.

dr-robertk:
This PR provides an implementation of the DUNE grid interface using OPM's UnstructuredGrid data structure.


CMakeLists_files.cmake
```diff
@@ -64,6 +64,7 @@ list (APPEND TEST_SOURCE_FILES
 	tests/cpgrid/orientedentitytable_test.cpp
```
atgeirr: There are lots more files added, so this list must be updated.

bska: > There are lots more files added, so this list must be updated.

I was confused by that too, but if you look closely there is a separate `CMakeLists.txt` in the `dune/grid/polyhedralgrid` directory that **does** list all additional files.  I have not looked into how the two systems interact and I'd appreciate some insight from a build-system expert.

dune/grid/common/CartesianIndexMapper.hpp
```diff
@@ -0,0 +1,48 @@
...
+#include <dune/common/exceptions.hh>
+
+namespace Dune
+{
+    template< class Grid >
+    class CartesianIndexMapper
+    {
+    public:
+        static const int dimension = Grid :: dimension ;
+        CartesianIndexMapper( const Grid& )
```
atgeirr: This constructor should be made explicit (also in specialisations).

CMakeLists_files.cmake
dr-robertk: I will move this to the general file to keep the package logic.

bska:
Just a general question: What, if any, are the requirements in terms of Dune release versions, of the `polyhedralgrid`?

dr-robertk:
@bska: dune 2.3.1 or later.

dr-robertk:
@bska: It sctually might also compile and work with 2.2, but I haven't tested that.

bska:
I asked

> What, if any, are the requirements in terms of Dune release versions, of the polyhedralgrid?

@dr-robertk replied

> dune 2.3.1 or later. It actually might also compile and work with 2.2, but I haven't tested that.

Okay, I'll launch a build test targeting 2.2 and report back.

dune/grid/polyhedralgrid/grid.hh
```diff
@@ -0,0 +1,1207 @@
...
+      return 0;
+    }
+
+    template <int codim>
+    typename Codim<codim>::EntitySeed
+    subEntitySeed( const typename Codim<0>::EntitySeed& elemSeed, const int i ) const
+    {
+      assert( i>= 0 && i<subEntities( elemSeed, codim ) );
+      typedef typename Codim<codim>::EntitySeed  EntitySeed;
+      if( codim == 0 )
```
atgeirr: Use braces with `if` statements.

dune/grid/polyhedralgrid/grid.hh
```diff
@@ -0,0 +1,1207 @@
...
+      {
+        return EntitySeed( grid_.cell_faces[ grid_.cell_facepos[ elemSeed.index() ] + i ] );
+      }
+      else if ( codim == dim )
+      {
+        return EntitySeed( cellVertices_[ elemSeed.index() ][ i ] );
+      }
+      else
+      {
+        DUNE_THROW(InvalidStateException,"codimension not availalbe");
```
atgeirr: Typo.

atgeirr: Also: why is InvalidStateException used? Would it not be better to use some invalid argument exception type? I also noticed that for some of the other methods such as `corner()` no exception is thrown, instead the origin is returned. Should that be changed to throwing for consistency?

dune/grid/polyhedralgrid/grid.hh
```diff
@@ -0,0 +1,1207 @@
...
+      }
+    }
+
+    const std::vector< GeometryType > &geomTypes ( int codim ) const
+    {
+      static std::vector< GeometryType > emptyDummy;
+      if (0 <= codim && codim < geomTypes_.size())
+        return geomTypes_[codim];
+
+      return emptyDummy;
```
atgeirr: Is this not an error? Why return anything if an invalid codim is specified?

dune/grid/polyhedralgrid/grid.hh
```diff
@@ -0,0 +1,1207 @@
...
+    }
+
+    int indexInInside( const typename Codim<0>::EntitySeed& seed, const int i ) const
+    {
+      if( ! grid_.cell_facetag )
```
atgeirr: Braces.

dune/grid/polyhedralgrid/grid.hh
```diff
@@ -0,0 +1,1207 @@
...
+        return i;
+      else
+      {
+        // assert( i>= 0 && i<subEntities( EntitySeed( seed.index() ) ) );
+        return grid_.cell_facetag[ grid_.cell_facepos[ seed.index() ] + i ] ;
```
atgeirr: The way I read this, for face `i` of a cell, you return just `i` unless it has a logical cartesian or cornerpoint structure, in which case you return the canonical (0-5) index for the face. So for a cornerpoint grid, at a fault (say in the J+ direction) where the cell has 2 neighbours (and therefore two adjacent faces), a total of 7 faces, the values returned would be:
```
i   indexInInside(i)
0     0
1     1
2     2
3     3
4     3
5     4
6     5
```
If I understand the dune grid interface correctly, the indices returned from indexInInside() should match up with the codim 1 entities of the grid. This means that for the logical-cartesian case (detected by having grid_.cell_facetag being non-null) the code here assumes that the canonical logical cartesian faces are exactly the codim 1 entities of the grid. Whereas in the general case, the codim 1 entities are assumed to be identical to the faces of the UnstructuredGrid. However, this is not reflected in the `subEntities()` and `subEntitySeed()` methods, where the general case seems to be assumed.

The `CpGrid` class handles this by simply saying there are no codim 1 entities. It still returns something (the logical cartesian index) from `Intersection::indexInInside()` though, which may be said to break the interface since there are no subentities of codim 1 to index into.

dune/grid/polyhedralgrid/grid.hh
```diff
@@ -0,0 +1,1207 @@
...
+      }
+    }
+
+    typename Codim<0>::EntitySeed
+    neighbor( const typename Codim<0>::EntitySeed& seed, const int i ) const
+    {
+      typedef typename Codim<0>::EntitySeed EntitySeed;
+      const int face = 2 * this->template subEntitySeed<1>( seed, i ).index();
```
atgeirr: This is misleading. The face index is what you have before you multiply by 2. I suggest moving the multiplication to lines 897 and 899, then it will look like tons of other code using the UnstructuredGrid. Also: braces for the `if` below.

dune/grid/polyhedralgrid/grid.hh
```diff
@@ -0,0 +1,1207 @@
...
+
+    int
+    indexInOutside( const typename Codim<0>::EntitySeed& seed, const int i ) const
+    {
+      typedef typename Codim<0>::EntitySeed EntitySeed;
+      EntitySeed nb = neighbor( seed, i );
+      const int faces = subEntities( seed, 1 );
+      for( int face = 0; face<faces; ++ face )
+      {
+        if( neighbor( nb, face ).equals(seed) )
```
atgeirr: Conceptual issues with indexInInside() also apply here. Also: braces around 'if' body.

dune/grid/polyhedralgrid/grid.hh
```diff
@@ -0,0 +1,1207 @@
...
+    }
+
+    template <class EntitySeed>
+    GlobalCoordinate
+    outerNormal( const EntitySeed& seed, const int i ) const
+    {
+      const int face  = this->template subEntitySeed<1>( seed, i ).index();
+      const int normalIdx = face * GlobalCoordinate :: dimension ;
+      GlobalCoordinate normal = copyToGlobalCoordinate( grid_.face_normals + normalIdx );
+      const int nb = grid_.face_cells[ face ];
```
atgeirr: Here you need to use `2*face`. Also, braces for the `if`. (I will now stop nagging about this, and assume you will fix all remaining un-nagged-about cases...)

dune/grid/polyhedralgrid/grid.hh
```diff
@@ -0,0 +1,1207 @@
...
+        return GlobalCoordinate( 0 );
+      }
+    }
+
+    GlobalCoordinate copyToGlobalCoordinate( const double* coords ) const
+    {
+      GlobalCoordinate coordinate;
+      for( int i=0; i<GlobalCoordinate::dimension; ++i )
+        coordinate[ i ] = coords[ i ];
+      return std::move( coordinate );
```
atgeirr: Since GlobalCoordinate is a Dune::FieldVector (it is, right?) I do not think move is going to help since a move is the same as a copy for that class. I do not think it hurts either (by preventing the return value optimization, which I think *is* important here) although I am not 100% sure.

dune/grid/polyhedralgrid/grid.hh
```diff
@@ -0,0 +1,1207 @@
...
+
+          cellVertices_[ c ].resize( cell_pts.size() );
+          std::copy(cell_pts.begin(), cell_pts.end(), cellVertices_[ c ].begin() );
+        }
+      }
+
+      geomTypes_.resize(dim + 1);
+      GeometryType tmp;
+      for (int codim = 0; codim <= dim; ++codim) {
+        tmp.makeCube(dim - codim);
```
atgeirr: This would be the same as CpGrid (except for also adding cubes for codims 1 and 2) and does not address the logical cartesian vs. general grid issue discussed above.

dune/grid/polyhedralgrid/grid.hh
```diff
@@ -0,0 +1,1207 @@
...
+        geomTypes_[codim].push_back(tmp);
+      }
+    }
+
+  protected:
+    std::shared_ptr< UnstructuredGridType > gridPtr_;
```
atgeirr: Good to be rid of the naked pointer, but did you try unique_ptr? A minor difference is that the deleter type is part of the template type of unique_ptr, unlike for shared_ptr which uses some type erasure trick.

atgeirr:
I have now reviewed the core of the implementation: the functions that do most of the mapping from UnstructuredGrid fields to the Dune grid interface. I found only one clear bug (line 926) but the question of codim 1 entities must be clarified before we merge this.

dune/grid/polyhedralgrid/grid.hh
andlaus: > If I understand the dune grid interface correctly, the indices returned from indexInInside() should match up with the codim 1 entities of the grid. This means that for the logical-cartesian case (detected by having grid_.cell_facetag being non-null) the code here assumes that the canonical logical cartesian faces are exactly the codim 1 entities of the grid.

we have struggled here a bit too, because the dune grid interface is a bit ambiguous here: it is not really clear what a codim 1 entity is in the case of a non-conforming grid. the only semantics of the `indexInside()` method that makes sense in the context of such grids is to return the index of the face in the reference element, which coincides with the "face tag".

dune/grid/polyhedralgrid/grid.hh
andlaus: as an (hacky) alternative. one can statically cast `coords` to `GlobalCoordinate*`

dr-robertk:
@atgeirr: The indexInInside issue is not so easy to handle. The current state is that all faces are treated as entities. This is a legacy from UnstructuredGrid because it would be impossible to reconstruct the "entities". I think it's ok in this case. Also, for real unstructured grids it is totally ok to make that assumption. The only problem arises in the corner-point case where all of a sudden a reference element is needed (because of the Cartesian logic from ECL to access data).  

We could also remove the codim 1 entities to avoid the conceptual problem. For a hexahedral grid with non conforming connections I don't see how the proper sub entities could be constructed from the information available in UnstructuredGrid.

dune/grid/polyhedralgrid/grid.hh
dr-robertk: What is your suggestion?

atgeirr: Just "return coordinate;".

Atgeirr

9. juli 2015 kl. 11:13 skrev dr-robertk <notifications@github.com>:

atgeirr:
About the indexInInside() issue: I think that this grid would be most useful if it provided codim 1 entities in the general case, and those entities were the same as the UnstructuredGrid faces. The cost of that would either be that the grid would not provide reference elements of a known GeometryType, even for the logical cartesian case, or that several functions would need two implementations (and a conditional), one for the general and one for the logical cartesian case.

The mapping to logical cartesian indices could still be provided I think.

bska:
I said
> I'll launch a build test targeting 2.2 and report back.

No go (sequential, Dune 2.2.1, GCC 4.8.4).
```
${ROOT}/dune-cornerpoint/dune/grid/polyhedralgrid/grid.hh:16:59: fatal error: dune/common/parallel/collectivecommunication.hh: No such file or directory
#include <dune/common/parallel/collectivecommunication.hh>
                                                          ^
compilation terminated.
```

Targeting Dune 2.3.1 is fine although `make test` fails
```
8/8 Testing: polyhedralgrid
8/8 Test: polyhedralgrid
Command: "/home/bska/work/opm/build/opm/dbg/seq/dune-2.3.1/dune-cornerpoint/bin/polyhedralgrid_test"
Directory: /data/projects/opm/build/opm/dbg/seq/dune-2.3.1/dune-cornerpoint/tests
"polyhedralgrid" start time: Jul 09 17:08 CEST
Output:
----------------------------------------------------------
Warning: codimension 2 not available in PolyhedralGrid
polyhedralgrid_test: /home/bska/work/opm/inst/dune/2.3.1/dbg/seq/include/dune/grid/test/gridcheck.cc:860: void gridcheck(Grid&) [with Grid = Dune::PolyhedralGrid<3, 3>]: Assertion `g.size(i)>0' failed.
<end of output>
Test time =   0.05 sec
----------------------------------------------------------
Test Failed.
"polyhedralgrid" end time: Jul 09 17:08 CEST
"polyhedralgrid" time elapsed: 00:00:00
----------------------------------------------------------
```

dr-robertk:
@bska: Should compile and run now with dune 2.2

dr-robertk:
@atgeirr: I introduced the geometry type none, when the face_tag in the UnstructuredGrid struct is NULL. Unfortunately, in dune-grid and dune-geometry there are some situations that have never been used with geometry type none. So guess what: it fails. That will need some time to fix. So for now we can live with the current solution.

dune/grid/polyhedralgrid/grid.hh
```diff
@@ -0,0 +1,1287 @@
...
+          }
+
+          cellVertices_[ c ].resize( cell_pts.size() );
+          std::copy(cell_pts.begin(), cell_pts.end(), cellVertices_[ c ].begin() );
+        }
+        // if no face_tag is available we assume that no reference element can be
+        // assigned to the elements
+        geomTypes_.resize(dim + 1);
+        GeometryType tmp;
+        for (int codim = 0; codim <= dim; ++codim)
```
atgeirr: The following code looks to me like it always creates None for the 0-entities, irrespective of face_tag or not. Is that intended?

atgeirr:
Sounds to me like you chose the most comprehensive (and complex) way: treating cells in logical cartesian grids as Cube and using None in the general case, with branching in methods to do the right thing depending on this. 

The PR could still contain bugs that could have been found by further review, but at this point I think testing it in practice will be a quicker way to identify problems. Apart from my question above concerning `geomTypes_`, I think I can merge this, if you think it is ready.

dune/grid/polyhedralgrid/grid.hh
dr-robertk: No, when face_tag is available then cube is used, otherwise none.

dr-robertk:
@atgeirr: I want to add the specialization of GridHelpers and then it should be ready.

blattms:
Can't we indicate that there are no entities of codim 1 or 2 with
Capabilities::asEntity<Grid,codim>::v? But I guess for
UnstructuredGrid this cannot be decided during compile time.

dune/grid/polyhedralgrid/grid.hh
atgeirr: I see now, I did not not notice that we were in an if-else.

blattms: But for a cornerpoint grid face_tag is available, right? Nevertheless
a face might look quite different from a cube.

dr-robertk:
@blattms: codim 2 (edges) are not available since there is not data for that in UnstructuredGrid.
For codim 1 it's a bit tricky and I guess we have to sort things out along the ride.

dr-robertk:
@atgeirr: I have added the GridHelpers to make PolyhedralGrid run with flow. It seems to work fine. 
There is a minor fix necessary in opm-autodiff. When this PR is merged I will add a PR in autodiff. 
This is simply the specialization of ADFaceCellTraits.

bska:
> Should compile and run now with dune 2.2

I do get a lot further, but not quite over the finish line I'm afraid:
```
${ROOT}/dune-cornerpoint/tests/polyhedralgrid_test.cc:14:39: fatal error: dune/grid/test/gridcheck.cc: No such file or directory
 #include <dune/grid/test/gridcheck.cc>
```
Note: This is with an **installed** version of Dune 2.2.1.  While `gridcheck.cc` is definitely part of the 2.2.1 distribution, the file does not appear to be among the products that gets published during `make install`.

I guess at this point we choose whether or not we want to support Dune 2.2.x (at least **installed** versions of Dune 2.2.1).  If we *do* continue to support 2.2.x we should build `tests/polyhedralgrid_test.cc` only when we're targeting 2.3.x or later.

dr-robertk:
@bska: Fixed. Should work now.

bska:
> Fixed. Should work now.

It does.  Thanks a lot.

By the way, I agree with @atgeirr's sentiment that we should probably expose this to further testing so I'm going to merge this into master.

bska:
> I'm going to merge this into master.

At least when you consider it to be ready, @dr-robertk.  Do you?

dr-robertk:
I think it's ready to merge.

bska:
> I think it's ready to merge.

Okay, then I'll merge.  Thanks a lot.

PR #153, opened 2015-07-10
Make header file list reflect reality

bska:
Basically, sort the header list and chase renamed/removed files.


bska:
I'll merge this myself since this fixes a build problem.

PR #154, opened 2015-07-13
PolyhedralGrid: enable Minpv keyword when constructing grid.

dr-robertk:
This PR enables the MINPV keyword in constructing a grid from deck. This is needed when used with Norne.


atgeirr:
Looks totally reasonable, will merge, thanks!

Just curious: why was it disabled initially?

PR #156, opened 2015-07-15
PolyhedralGrid: disable grid check for all dune versions < 2.4

dr-robertk:
The gridcheck for checking the dune-grid interface implementation has a small bug when used with grid that do not provide sizes for all codimensions. I have fixed this in the current 2.4 version. 
For all prior versions we simply disable the gridcheck.


atgeirr:
Thanks! Merging.

PR #157, opened 2015-07-23
Works around internal compiler error for polyhedralgrid with g++-4.4.

blattms:
Not sure whether we still support gcc-4.4, but: 

It seems like gcc-4.4 (Debian wheezy) has problems detecting the correct
type when using ```const auto& var= *(dune-grid-iterator)```. At least for version
"g++ (Debian 4.7.2-5) 4.7.2" I get an internal compiler error:
```sh
dune-cornerpoint/tests/polyhedralgrid_test.cc:49: internal compiler error: in type_unification_real, at cp/pt.c:12624
Please submit a full bug report,
with preprocessed source if appropriate.
See <file:///usr/share/doc/gcc-4.4/README.Bugs> for instructions.
```
Therefore this patch prevents dereferncing the iterators and uses operator->
instead.

### DIFF
--- a/tests/polyhedralgrid_test.cc
+++ b/tests/polyhedralgrid_test.cc
@@ -46,8 +46,7 @@ void testGrid( const GridView& gridView )
     ElemIterator elemIt = gridView.template begin<0>();
     ElemIterator elemEndIt = gridView.template end<0>();
     for (; elemIt != elemEndIt; ++elemIt) {
-        const auto& elem = *elemIt;
-        const Geometry& elemGeom = elem.geometry();
+        const Geometry& elemGeom = elemIt->geometry();
         if (std::abs(elemGeom.volume() - 1.0) > 1e-8)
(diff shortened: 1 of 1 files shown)


tests/polyhedralgrid_test.cc
```diff
@@ -61,27 +60,26 @@ void testGrid( const GridView& gridView )
 
 
         int numIs = 0;
-        IsIt isIt = gridView.ibegin(elem);
-        IsIt isEndIt = gridView.iend(elem);
+        IsIt isIt = gridView.ibegin(*elemIt);
```
atgeirr: Since this line apparently works, it cannot just be the dereferencing that makes things go wrong, but using it with `auto`, or perhaps assigning it to a reference. I can only speculate (assuming I do not want to start learning gcc internals)...

atgeirr:
Thanks for the patch. While we have decided to leave 4.4 behind and settle on gcc 4.8 or equivalent, I think there is not much point in *not* merging a workaround like this if it makes life easier for someone, so I'll merge it.

PR #161, opened 2015-08-11
Silence warnings

atgeirr:
With this, dune-cornerpoint compiles without warnings on my system.

@blattms: when you have time, I'd appreciate if you could verify that we should be using the MPI-2 version of the errhandler functions (see the distribution_test.cpp change).


blattms:
I did look at this. Looks good. Thanks a lot for the patch and the merge.

PR #163, opened 2015-09-01
Add possibility to use well information when partitioning.

blattms:
With this commit one can pass eclipse information to the ```CpGrid::loadbalance``` routine.
If this is done then a graph resembling the well information is used in addition. This graph
has connections for all cells that are perforated by a well (i.e. the cell
is part of the competion set) with all cells that are perforated by the same well.
Then this connectivity is added to the one resembling the grid (cell are vertices,
edges are the faces). The connections of the wells get a very high weight, and the
ones of the grid a low one. This combined graph is then provided to ZOLTAN for
loadbalancing.

### DIFF
--- a/dune/grid/CpGrid.hpp
+++ b/dune/grid/CpGrid.hpp
@@ -568,6 +568,12 @@ namespace Dune
 
         /// \brief Distributes this grid over the available nodes in a distributed machine
         /// \param ecl Pointer to the eclipse state information. Default: null
+        ///            If this is not null then complete well information of
+        ///            of the last scheduler step of the eclipse state will be
+        ///            used to make sure that all the possible completion cells
+        ///            of each well are stored on one process. This done by
+        ///            adding an edge with a very high edge weight for all
+        ///            possible pairs of cells in the completion set of a well.
         /// \param The number of layers of cells of the overlap region (default: 1).
         /// \warning May only be called once.
         bool loadBalance(Opm::EclipseStateConstPtr ecl=Opm::EclipseStateConstPtr(),
@@ -579,6 +585,12 @@ namespace Dune
         /// \brief Distributes this grid and data over the available nodes in a distributed machine.
         /// \param data A data handle describing how to distribute attached data.
         /// \param ecl Pointer to the eclipse state information. Default: null
+        ///            If this is not null then complete well information of
+        ///            of the last scheduler step of the eclipse state will be
+        ///            used to make sure that all the possible completion cells
+        ///            of each well are stored on one process. This done by
+        ///            adding an edge with a very high edge weight for all
+        ///            possible pairs of cells in the completion set of a well.
         /// \param overlapLayers The number of layers of overlap cells to be added
         ///        (default: 1)
         /// \tparam DataHandle The type implementing DUNE's DataHandle interface.
@@ -1057,7 +1069,14 @@ namespace Dune
 #endif
 
     private:
-        /// Scatter a global grid to all processors.
+        /// \brief Scatter a global grid to all processors.
+        /// \param ecl Pointer to the eclipse state information. Default: null
+        ///            If this is not null then complete well information of
+        ///            of the last scheduler step of the eclipse state will be
+        ///            used to make sure that all the possible completion cells
+        ///            of each well are stored on one process. This done by
+        ///            adding an edge with a very high edge weight for all
+        ///            possible pairs of cells in the completion set of a well.
         bool scatterGrid(Opm::EclipseStateConstPtr ecl, int overlapLayers);
 
         /** @brief The data stored in the grid.
--- a/dune/grid/CpGrid.hpp
+++ b/dune/grid/CpGrid.hpp
@@ -46,6 +46,8 @@
 // Warning suppression for Dune includes.
 #include <opm/core/utility/platform_dependent/disable_warnings.h>
 
+#include <opm/parser/eclipse/EclipseState/EclipseState.hpp>
+
 #include <dune/grid/common/capabilities.hh>
 #include <dune/grid/common/grid.hh>
 #include <dune/grid/common/gridenums.hh>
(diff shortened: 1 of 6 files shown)


dune/grid/common/ZoltanGraphFunctions.cpp
```diff
@@ -188,6 +230,130 @@ void getCpGridEdgeList(void *cpGridPointer, int sizeGID, int sizeLID,
...
+}
+
+CombinedGridWellGraph::CombinedGridWellGraph(const CpGrid& grid,
+                      const Opm::EclipseStateConstPtr eclipseState)
+    : grid_(grid)
+{
+    wellsGraph_.resize(grid.numCells());
+    // create compressed lookup from cartesian.
+    std::vector<Opm::WellConstPtr> wells  = eclipseState->getSchedule()->getWells();
+    std::vector<int> cartesian_to_compressed(grid.numCells(), -1);
```
bska: That's almost certainly a programming error.  Unless something's changed very recently, the `numCells` method returns the number of active (compressed) cells but you need the number of global cells here (i.e., the product of the `cartDims()`).

blattms: I love reviews! Thanks for noticing. Will fix ASAP.

dune/grid/common/ZoltanGraphFunctions.hpp
```diff
@@ -97,12 +98,81 @@ inline int getNullNumCells(void* cpGridPointer, int* err)
...
+        return wellsGraph_;
+    }
+    
+private:
+    void addCompletionSetToGraph(std::set<int>& well_indices)
+    {
+        for( auto well_idx = well_indices.begin(); well_idx != well_indices.end();
+             ++well_idx)
+        {
+            auto well_idx2 = well_idx;
```
bska: Is it okay if the matrix is upper triangular or do you need all pairs?

dune/grid/CpGrid.hpp
```diff
@@ -565,23 +567,28 @@ namespace Dune
         // loadbalance is not part of the grid interface therefore we skip it.
 
         /// \brief Distributes this grid over the available nodes in a distributed machine
+        /// \param ecl Pointer to the eclipse state information. Default: null
```
atgeirr: Although I know from discussions and the PR intro what this argument is used for, it is not clear in isolation. Could you add a sentence stating that if an eclipse stat is given, well information is extracted and used?

blattms: I added some more documentation about the eclipse state information
passed to loadbalance.
-- 
Join us at the DUNE User Meeting Sept. 28-29, 2015, Heidelberg, Germany
http://users.dune-project.org/projects/dune-user-meeting-2015/wiki

dune/grid/common/ZoltanPartition.cpp
```diff
@@ -53,11 +54,24 @@ std::vector<int> zoltanGraphPartitionGridOnRoot(const CpGrid& cpgrid,
...
+    Zoltan_Set_Param(zz, "OBJ_WEIGHT_DIM", "0");
     Zoltan_Set_Param(zz, "PHG_EDGE_SIZE_THRESHOLD", ".35");  /* 0-remove all, 1-remove none */
 
     bool pretendEmptyGrid = cc.rank()!=root;
+    std::shared_ptr<CombinedGridWellGraph> grid_and_wells;
 
-    Dune::cpgrid::setCpGridZoltanGraphFunctions(zz, cpgrid, pretendEmptyGrid);
+    if( eclipseState )
+    {
+        Zoltan_Set_Param(zz,"EDGE_WEIGHT_DIM","1");
```
atgeirr: Not being familiar with Zoltan details, I presume this means that each graph edge has a single-number weight, and that we do not use this (no weighting) in the default (no eclipseState) case?

Would this be the place to put transmissibilities for weights?

dune/grid/common/ZoltanPartition.cpp
```diff
@@ -81,6 +95,30 @@ std::vector<int> zoltanGraphPartitionGridOnRoot(const CpGrid& cpgrid,
...
+            for( auto vertex : well )
+            {
+                if( part != parts[vertex] )
+                {
+                    cells_on_other.insert(std::make_pair(vertex, parts[vertex]));
+                }
+            }
+            if ( cells_on_other.size() )
+            {
+                OPM_THROW(std::domain_error, "Well is distributed between proceses, which should not be the case!");
```
atgeirr: Typo: proceses -> processes.

dune/grid/common/ZoltanPartition.hpp
```diff
@@ -31,17 +31,20 @@ namespace cpgrid
 /// \brief Partition a CpGrid using Zoltan
 ///
 /// This function will extract Zoltan's graph information
-/// form the grid and use it to partition it.
+/// form the grid, and the wells and use it to partition the grid.
```
atgeirr: Typo: form -> from.

dune/grid/common/ZoltanGraphFunctions.hpp
```diff
@@ -97,12 +98,81 @@ inline int getNullNumCells(void* cpGridPointer, int* err)
...
+/// the faces of the grid and the ones that represent the connections within the
+/// wells. If a well has completions
+/// on cell i and cell j, then there is an edge from i to j and j to i in the graph.
+/// Even for shut wells the connections will exist.
+class CombinedGridWellGraph
+{
+public:
+    typedef std::vector<std::set<int> > GraphType;
+
+    /// \brief Create a grapg representing a grid together with the wells.
```
atgeirr: Typo: grapg -> graph.

atgeirr:
This looks fine! I have not checked the Zoltan-related details (that would require me to know a lot more about that library) though. Ping me when the typos are fixed and I'll merge.

dune/grid/common/ZoltanPartition.cpp
blattms: Yes, using transmissibilities as edge weights could make a lot of sense.

bska: > Yes, using transmissibilities as edge weights could make a lot of sense.

Absolutely.  In the context of METIS I have seen that using the weighting function `log(1/T)` with special treatment of connections for which `T=0` tends to give reasonable partitions.  What functions to use does of course depend greatly on what objective function the software uses in the partitioning.

blattms:
Seems like there are still problems with Norne. I need to take another look. It might be that Zoltan decides to split up some wells nevertheless.

atgeirr:
> Seems like there are still problems with Norne. I need to take another look. It might be that Zoltan decides to split up some wells nevertheless.

Still, I'll merge this now, as it is a big step in the right direction.

dune/grid/common/ZoltanPartition.cpp
blattms: >  I have seen that using the weighting function log(1/T) with special treatment of connections for which T=0 tends to give reasonable partitions

@bska Could you comment on why log(1/T)? If I look at `DerivedGeology::trans_` then the entries seem to be in the range 10^-14 to 10^-12 for Norne.  The formulation would mean that lower transmissibilities produce higher weights. But in my humble understanding we get higher flow over faces with high transmissibility. What am I missing?

blattms: I just checked the transmissibilities for Norne are in the range [0, 4.23e^-9]

atgeirr: That is in the input, right? Then it is in millidarcy.

bska: > Could you comment on why `log(1/T)`?

Sure.  It is indeed a bit counterintuitive but the objective of METIS is to minimise communication cost associated with each edge in the partitioned graph.  As `T` is a measure of conductivity, `1/T` is a measure of resistance.  Taking logarithms furthermore ensures that the wildly varying field becomes a bit smoother.

PR #165, opened 2015-09-03
[bugfix] Fixes const reference to temporary in Cell2FacesRow.

blattms:
Cell2FacesRow store a temporary the a row of Opm::SparseTable.
Unfortunately that is a temprory object, a boost::iterator range.
This invalidates any iterators over this row once there is no
more reference to it. Code like
```C++
auto cell2Faces = UgGridHelpers::cell2Faces(G);
for (auto fi=cell2Faces[*ci].begin(),
         fe=cell2Faces[*ci].end();
       fi != fe;
       ++fi)
{
   std::cout<<*fi<<std::endl;
}
```
might result in hard to debug segmentation faults as accessing the row
in fi uses the reference tp a temporary object that does not exist any
more.

With this commit we store copies of the boost iterator in both the row
and its iterator. Therefore with this commit these two also get reassignable.
Copying an iterator should be fast anyway and not reassignable iterators seem
unnatural.

Closes OPM/opm-autodiff#457

### DIFF
--- a/dune/grid/cpgrid/GridHelpers.hpp
+++ b/dune/grid/cpgrid/GridHelpers.hpp
@@ -233,17 +233,17 @@ public:
         public IndexIterator
     {
     public:
-        iterator(const Dune::cpgrid::OrientedEntityTable<0,1>::row_type* row,
+        iterator(const Dune::cpgrid::OrientedEntityTable<0,1>::row_type& row,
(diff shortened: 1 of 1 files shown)


blattms:
BTW on my system this is need for running norne with an optimized flow_cp. See issue OPM/opm-autodiff#457.

PR #167, opened 2015-09-08
Postprocess loadbalanced partioning for wells

blattms:
After loadbalancing we make another iteration over the wells to check that they are not cut by the process borders. If that is the case we assign the perforated cells manually to the process with the most completions 
of this cell.

### DIFF
--- a/dune/grid/common/ZoltanGraphFunctions.cpp
+++ b/dune/grid/common/ZoltanGraphFunctions.cpp
@@ -319,9 +319,15 @@ void getCpGridWellsEdgeList(void *graphPointer, int sizeGID, int sizeLID,
 }
 
 CombinedGridWellGraph::CombinedGridWellGraph(const CpGrid& grid,
-                      const Opm::EclipseStateConstPtr eclipseState)
+                                             const Opm::EclipseStateConstPtr eclipseState,
+                                             bool pretendEmptyGrid)
     : grid_(grid)
 {
+    if ( pretendEmptyGrid )
+    {
+        // wellsGraph not needed
+        return;
+    }
     wellsGraph_.resize(grid.numCells());
     const auto& cpgdim = grid.logicalCartesianSize();
     // create compressed lookup from cartesian.
@@ -347,9 +353,7 @@ CombinedGridWellGraph::CombinedGridWellGraph(const CpGrid& grid,
             assert(compressed_idx>=0);
             well_indices.insert(compressed_idx);
         }
-            
         addCompletionSetToGraph(well_indices);
-            
     }
 }
 
@@ -373,7 +377,6 @@ void setCpGridZoltanGraphFunctions(Zoltan_Struct *zz, const Dune::CpGrid& grid,
     }
 }
 
-    
 void setCpGridZoltanGraphFunctions(Zoltan_Struct *zz,
                                    const CombinedGridWellGraph& graph,
                                    bool pretendNull)
@@ -394,7 +397,7 @@ void setCpGridZoltanGraphFunctions(Zoltan_Struct *zz,
         Zoltan_Set_Num_Edges_Multi_Fn(zz, getCpGridWellsNumEdgesList, graphPointer);
         Zoltan_Set_Edge_List_Multi_Fn(zz, getCpGridWellsEdgeList, graphPointer);
     }
-}   
+}
 } // end namespace cpgrid
 } // end namespace Dune
 #endif // HAVE_ZOLTAN
--- a/dune/grid/common/ZoltanGraphFunctions.cpp
+++ b/dune/grid/common/ZoltanGraphFunctions.cpp
@@ -321,7 +321,7 @@ void getCpGridWellsEdgeList(void *graphPointer, int sizeGID, int sizeLID,
 CombinedGridWellGraph::CombinedGridWellGraph(const CpGrid& grid,
                                              const Opm::EclipseStateConstPtr eclipseState,
                                              bool pretendEmptyGrid)
-    : grid_(grid)
+    : grid_(grid), eclipseState_(eclipseState)
 {
     if ( pretendEmptyGrid )
     {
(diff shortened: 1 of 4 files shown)


blattms:
BTW up  to now Zoltan did never split a well between partitions. Whenever I thought it turned out to be a bug in the checks on my end.

dune/grid/common/ZoltanPartition.cpp
```diff
@@ -95,15 +95,20 @@ std::vector<int> zoltanGraphPartitionGridOnRoot(const CpGrid& cpgrid,
...
-    //#ifndef NDEBUG
-    if( eclipseState )
+    if( eclipseState && ! pretendEmptyGrid )
     {
+        grid_and_wells->postProcessPartitioningForWells(parts);
+#ifndef NDEBUG
         int index = 0;
         for( auto well : grid_and_wells->getWellsGraph() )
         {
+            if( ! well.size() )
```
atgeirr: Is this block necessary? I'd think that for an empty `well` the code would work as is?

dune/grid/cpgrid/CpGrid.cpp
```diff
@@ -139,6 +139,8 @@ bool CpGrid::scatterGrid(Opm::EclipseStateConstPtr ecl, int overlapLayers)
         distributed_data_.reset(new cpgrid::CpGridData(new_comm));
         distributed_data_->distributeGlobalGrid(*this,*this->current_view_data_, cell_part,
                                                 overlapLayers);
+        std::cout<<"After loadbalancing process "<<my_num<<" has "<<
```
atgeirr: Nitpick: whitespace around operators.

dune/grid/common/ZoltanGraphFunctions.cpp
```diff
@@ -347,9 +353,69 @@ CombinedGridWellGraph::CombinedGridWellGraph(const CpGrid& grid,
...
+    std::vector<int> cartesian_to_compressed(cpgdim[0]*cpgdim[1]*cpgdim[2], -1);
+    for( int i=0; i < grid_.numCells(); ++i )
+    {
+        cartesian_to_compressed[grid_.globalCell()[i]] = i;
+    }
+    int last_time_step = eclipseState_->getSchedule()->getTimeMap()->size()-1;
+    // Check that all completions of a well have ended up on one process.
+    // If that is not the case for well then move them manually to the
+    // process that already has the most completions on it.
+    for (auto wellIter= wells.begin(); wellIter != wells.end(); ++wellIter) {
```
atgeirr: Nitpick: space around ops here, line 388, 395, 408, 409.

dune/grid/common/ZoltanGraphFunctions.cpp
```diff
@@ -347,9 +353,69 @@ CombinedGridWellGraph::CombinedGridWellGraph(const CpGrid& grid,
...
+        }
+        std::map<int,std::size_t> no_completions_on_proc;
+        for (size_t c=0; c<completionSet->size(); c++) {
+            Opm::CompletionConstPtr completion = completionSet->get(c);
+            int i = completion->getI();
+            int j = completion->getJ();
+            int k = completion->getK();
+            int cart_grid_idx = i + cpgdim[0]*(j + cpgdim[1]*k);
+            int compressed_idx = cartesian_to_compressed[cart_grid_idx];
+            assert(compressed_idx>=0);
```
atgeirr: This implies that we require all completions to be in active cells. I am not sure if we require this in the WellsManager any more? Should we require it?

dune/grid/common/ZoltanPartition.cpp
blattms: You are probably right. I will change and test it.

dune/grid/common/ZoltanGraphFunctions.cpp
bska: > I am not sure if we require [that all declared completions be in active cells]

We don't.

atgeirr: But in that case would not the assert fire? We should rather `continue` than assert I think (also in the loop further down that is similar).

bska: > But would not the assert fire [when presented with an completion outside the active cells]?

It would indeed.  Skipping that completion (i.e., `continue`) is probably the correct choice here.

blattms: I beg to differ. I think that we do require that completions are in active cells in https://github.com/OPM/opm-core/blob/3261ec3281cb43b64ff4ff5f51cfcdf9f564147/opm/core/wells/WellsManager_impl.hpp#L164 by throwing if we cannot find a compressed index for the cartesian one. Or am I missing something?

Nevertheless we need to throw an exception here. (Maybe you did mean that anyway.)

bska: > I think that we do require that completions are in active cells in [parts of the code]

That's a bug, then.  It is absolutely possible for the completion set (i.e., keyword `COMPDAT`) to identify cells that don't exist once the grid topology is created (e.g., after pinch-out).

blattms: > > But would not the assert fire [when presented with an completion outside the active cells]?
> 
> It would indeed.  Skipping that completion (i.e., `continue`) is probably the correct choice here.

If we settle on continue, then we should probably change it in the
WellsManager, too. Currently, an exception is thrown there in this case.

blattms: Thanks for the hint and explanation. I have changed all the asserts in
this PR to ignores using a continue statement in the loops.

dune/grid/common/ZoltanGraphFunctions.cpp
```diff
@@ -344,12 +350,83 @@ CombinedGridWellGraph::CombinedGridWellGraph(const CpGrid& grid,
...
+                      << new_owner << std::endl;
+             for ( size_t c = 0; c < completionSet->size(); c++ )
+             {
+                 Opm::CompletionConstPtr completion = completionSet->get(c);
+                 int i = completion->getI();
+                 int j = completion->getJ();
+                 int k = completion->getK();
+                 int cart_grid_idx = i + cpgdim[0]*(j + cpgdim[1]*k);
+                 int compressed_idx = cartesian_to_compressed[cart_grid_idx];
+                 if ( compressed_idx >= 0 ) // ignore completions in inactive cells
```
bska: This should probably say `compressed_idx < 0`.

atgeirr:
I'd like to merge this, but the compressed_idx >= 0 remains to be fixed.

blattms:
The ill-posed condition on the `compressed_idx` should be fixed by  1760899
Thanks again for the nice review.

atgeirr:
Thanks!

PR #168, opened 2015-09-10
Provide zcorn-averaged cell depth for the cp grid.

osae:
This PR is associated with PR opm-core #862.

### DIFF
--- a/dune/grid/CpGrid.hpp
+++ b/dune/grid/CpGrid.hpp
@@ -753,7 +753,18 @@ namespace Dune
         {
             return current_view_data_->face_to_point_[face][local_index];
         }
-
+        /// \brief Get vertical position of cell center ("zcorn" average).
+        /// \brief cell_index The index of the specific cell.
+        double cellCenterDepth(int cell_index)
+        {
+            double zz = 0.0;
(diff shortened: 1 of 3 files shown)


blattms:
There sould be no need to recompute as his information should be readily available:
```c++
return current_view_data_->geomVector<0>()[Entity<0>(*this,cell_index)[nd-1];
```
or
```c++
return this->cellCentroid(cell_index)[nd-1]
```

blattms:
Sorry, I did not read OPM/opm-core#862 first. My suggestion result in the difference that rose the invention of this function. Can someone enlighten me what the difference is? Do UG and CpGrid even use the  same centroid computation?

atgeirr:
> Do UG and CpGrid even use the same centroid computation?

No they don't, and small differences in centroids, areas and volumes could be a significant factor in why the results are different.

blattms:
Agreed. That might explain the difference observed in OPM/opm-autodff#467

atgeirr:
I am currently running a build test on this and the opm-core PR, and will merge if that succeeds. I would appreciate a minor change here though: a copy of the explanatory comment (present in the UnstructuredGrid version) would be good to have here, too.

osae:
OK, I'll copy the comment.

PR #170, opened 2015-09-14
Point-2-Point communicator.

dr-robertk:
This PR adds a point-2-point communicator for Dune/OPM. This will be used in an upcoming PR for the parallel output. 

@atgeirr: I have tried to include as many white space violations of the coding style as possible ;-).

### DIFF
--- a/dune/grid/common/p2pcommunicator.hh
+++ b/dune/grid/common/p2pcommunicator.hh
@@ -51,7 +51,7 @@ public:
     }
 
     /** \brief clear the buffer */
-    void clear() { BufferType().swap( buffer_ ); resetReadPosition(); }
+    void clear() { buffer_.clear(); resetReadPosition(); }
     /** \brief reset read position of buffer to beginning */
     void resetReadPosition() { pos_ = 0 ; }
     /** \brief return size of buffer */
--- a/dune/grid/common/p2pcommunicator.hh
+++ b/dune/grid/common/p2pcommunicator.hh
@@ -163,42 +163,46 @@ public:
     inline void insertRequest( const std::set< int >& sendLinks, const std::set< int >& recvLinks );
 
   public:
-    // return number of processes we will send data to
+    /** \brief return number of processes we will send data to */
     inline int sendLinks () const { return _sendLinkage.size(); }
-    // return number of processes we will receive data from
+    /** \brief return number of processes we will receive data from */
     inline int recvLinks () const { return _recvLinkage.size(); }
 
-    // return vector containing possible recv buffer sizes
+    /** \brief return vector containing possible recv buffer sizes */
     const vector_t& recvBufferSizes() const { return _recvBufferSizes; }
 
-    // use assert here, since this part also affects some communications methods in dune-fem
+    /** \brief return send link number for a given send rank number */
     inline int sendLink (const int rank) const
     {
       assert (_sendLinkage.end () != _sendLinkage.find (rank)) ;
       return (* _sendLinkage.find (rank)).second ;
     }
 
+    /** \brief return recv link number for a given recv rank number */
     inline int recvLink (const int rank) const
     {
       assert (_recvLinkage.end () != _recvLinkage.find (rank)) ;
       return (* _recvLinkage.find (rank)).second ;
     }
 
-    // use assert here, since this part also affects some communications methods in dune-fem
+    /** \brief return vector containing all process numbers we will send to */
     const std::vector< int > &sendDest   () const { return _sendDest; }
+    /** \brief return vector containing all process numbers we will receive from */
     const std::vector< int > &recvSource () const { return _recvSource; }
 
-    // remove stored linkage
+    /** \brief remove stored linkage */
     inline void removeLinkage () ;
 
-    // exchange message buffers with peers
-    std::vector< MessageBufferType > exchange (const std::vector< MessageBufferType > &) const;
+    /** \brief exchange message buffers with peers defined by inserted linkage */
+    virtual std::vector< MessageBufferType > exchange (const std::vector< MessageBufferType > &) const;
 
-    // exchange object stream and immediately unpack, when data was received
-    void exchange ( DataHandleInterface& ) const;
+    /** \brief exchange data with peers, handle defines pack and unpack of data */
+    virtual void exchange ( DataHandleInterface& ) const;
 
-    // exchange data and reuse buffer sizes from the previous run
-    void exchangeCached ( DataHandleInterface& ) const;
+    /** \brief exchange data with peers, handle defines pack and unpack of data,
+     *  if receive buffers are known from previous run and have not changed
+     *  communication could be faster */
+    virtual void exchangeCached ( DataHandleInterface& ) const;
   };
 
 } // namespace Dune
--- a/dune/grid/common/p2pcommunicator.hh
+++ b/dune/grid/common/p2pcommunicator.hh
@@ -109,11 +109,11 @@ public:
     typedef std::map< int, int > linkage_t;
     typedef std::vector< int >   vector_t;
 
-    linkage_t  _sendLinkage ;
-    linkage_t  _recvLinkage ;
+    linkage_t  sendLinkage_ ;
+    linkage_t  recvLinkage_ ;
 
-    vector_t   _sendDest ;
-    vector_t   _recvSource ;
+    vector_t   sendDest_ ;
+    vector_t   recvSource_ ;
 
     mutable vector_t   _recvBufferSizes;
     mutable bool       _recvBufferSizesComputed;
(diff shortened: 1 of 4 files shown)


atgeirr:
I assume this is essentially cloned from ALUGrid functionality as we discussed earlier? If so, can I also assume it has been well tested? I am not quite sure how to review this, since I am not an MPI expert. I will take a look though.

> I have tried to include as many white space violations of the coding style as possible

Always a rebel...

dune/grid/common/p2pcommunicator.hh
```diff
@@ -0,0 +1,240 @@
...
+    /** \brief constructor taking memory reserve estimation factor (default is 1.1, i.e. 10% over estimation )
+     */
+    SimpleMessageBuffer( const double factor = 1.1 )
+      : buffer_(), factor_( factor )
+    {
+      resetReadPosition();
+    }
+
+    /** \brief clear the buffer */
+    void clear() { BufferType().swap( buffer_ ); resetReadPosition(); }
```
atgeirr: Why is the "hard" clear necessary for buffer_? I would think a plain buffer_.clear() would be enough?

dr-robertk:
The implementation of the  MPI API has indeed been well tested. The SimpleMessageBuffer is new, but the test shows that it works.

dune/grid/common/p2pcommunicator.hh
dr-robertk: Probably. Just a matter of taste. Not very important.

dune/grid/common/p2pcommunicator.hh
```diff
@@ -0,0 +1,240 @@
...
+      size_t pos  = buffer_.size();
+      const size_t sizeNeeded = pos + tsize ;
+      // reserve with some 10% overestimation
+      if( buffer_.capacity() < sizeNeeded )
+      {
+        reserve( size_t(factor_ * sizeNeeded) ) ;
+      }
+      // resize to size need to store value
+      buffer_.resize( sizeNeeded );
+      for(unsigned int i=0; i<tsize; ++i )
```
atgeirr: Could replace this by std::copy().

dune/grid/common/p2pcommunicator.hh
```diff
@@ -0,0 +1,240 @@
...
+
+    /** \brief read value from buffer, value must implement the operator= correctly (i.e. no internal pointers etc.) */
+    template <class T>
+    void read( T& value ) const
+    {
+      // read bytes from stream and store in value
+      const size_t tsize = sizeof( T );
+      union { T value; char bytes[ tsize ]; } convert;
+      assert( pos_ + tsize <= buffer_.size() );
+      for( unsigned int i=0; i<tsize; ++i )
```
atgeirr: As above. Also: why not use size_t for `i` when that is used for `tsize`?

dune/grid/common/p2pcommunicator.hh
```diff
@@ -0,0 +1,240 @@
...
+  /** \brief Point-2-Point communicator for exchange messages between processes */
+  template <class MsgBuffer, class Comm = MPIHelper::MPICommunicator >
+  class Point2PointCommunicator : public CollectiveCommunication< Comm >
+  {
+  protected:
+    typedef CollectiveCommunication< Comm > BaseType;
+    typedef Point2PointCommunicator< MsgBuffer, Comm > ThisType;
+
+    // starting message tag
+    static const int messagetag = 234;
```
atgeirr: Is 234 an MPI code or an arbitrary number?

dune/grid/common/p2pcommunicator.hh
```diff
@@ -0,0 +1,240 @@
...
+    };
+
+  public :
+    // default constructor
+    Point2PointCommunicator() : BaseType() { removeLinkage(); }
+    // constructor taking CollectiveCommunication
+    Point2PointCommunicator( const BaseType& comm ) : BaseType( comm ) { removeLinkage(); }
+
+    // return new tag number for the exchange messages
+    static int getMessageTag( const unsigned int icrement )
```
atgeirr: Typo?

bska:
I'll just state once that the kind of type-punning used in the `read` and `write` functions always makes me uneasy.  Type-punning through a union is supported and portable in C99 and C11 but not in C89.  C++ introduces the notion of an **active** member of a `union` and basically states that reading a union through a member other than the one used to store it is implementation defined.

```C++
template <class T>
void write( const T& value )
{
    // union to access bytes in value
    const size_t tsize = sizeof( T );
    union { T value; char bytes[ tsize ]; } convert;
    convert.value = value;
    // ...
    for(unsigned int i=0; i<tsize; ++i )
    {
        buffer_[ pos++ ] = convert.bytes[ i ] ;
    }
}
```

atgeirr:
> reading a union through a member other than the one used to store it is implementation defined.

Would a reinterpret_cast<char*> behave better (or at least more standard-guaranteed) for this purpose?

bska:
>> reading a union through a member other than the one used to store it is implementation defined.
>
> Would a `reinterpret_cast<char*> behave better (or at least more standard-guaranteed) for this purpose?

Yes, at least as long as `T` is a POD.  I think I'd prefer
```C++
reinterpret_cast<unsigned char*>
```
though.

dune/grid/common/p2pcommunicator.hh
dr-robertk: It's and arbitrary number away from standard MPI message tasks which should be between 0 and 32768.

atgeirr:
Googled this and found no clear recommendations. Apparently there is undefined behaviour in all the methods. Makes you wonder how memcpy can be implemented in a standard-conforming way, maybe it cannot? Or I just did not find a good resource.

dr-robertk:
Actually, reinterpret_cast is used in the ALUGrid implementation and I though that was a bit fishy. 
So I'll change that then.

bska:
> Makes you wonder how `memcpy` can be implemented in a standard-conforming way, maybe it cannot?

It cannot.  Function `memcpy()` is a portable interface into a non-portable feature.  There is almost always some assembly implementation that takes care of marshalling the bits once it gets close to the hardware.

dr-robertk:
I addressed the above comments.

dune/grid/common/p2pcommunicator.hh
```diff
@@ -0,0 +1,233 @@
...
+        buffer_.reserve( size );
+    }
+
+    /** \brief resize buffer to 'size' entries  */
+    void resize( const size_t size )
+    {
+        buffer_.resize( size );
+    }
+
+    /** \brief write value to buffer, value must implement the operator= correctly (i.e. no internal pointers etc.) */
```
atgeirr: Final point (I promise): this comment (and below) are now misleading. I guess it should say that T must be a POD?

dr-robertk: That's actually not entirely correct. As long as the operator= is implemented correctly and the types contained in value contain only types that only contain types that at some point are PODs it's ok. But, constructors and destructors are allowed.

atgeirr: Maybe "standard layout" is the term I am looking for? Anyway, I'll not let it hold back. Merging.

bska:
This does not build out of the box in an environment that targets Dune 2.2.x.  I can, however, restore the build in that environment, at least in a sequential configuration (#171), if I apply the change below to
```
dune/grid/common/p2pcommunicator.hh
```
(i.e., look for Dune MPI headers in `dune/common` rather than `dune/common/parallel` when targeting 2.2.x).  The `p2pcommunicator_test` also runs successfully in this environment.

Is Dune 2.2.x support worth the added maintenance cost or should we simply disable the point-to-point communicator unless we're targeting Dune 2.3.x or later?
```diff
diff --git a/dune/grid/common/p2pcommunicator.hh b/dune/grid/common/p2pcommunicator.hh
index fd95dad..79e13ca 100644
--- a/dune/grid/common/p2pcommunicator.hh
+++ b/dune/grid/common/p2pcommunicator.hh
@@ -24,13 +24,29 @@
 #include <set>
 #include <map>
 
+#include <dune/common/version.hh>
+
+#if DUNE_VERSION_NEWER(DUNE_COMMON, 2, 3)
+
 #include <dune/common/parallel/mpihelper.hh>
 #include <dune/common/parallel/collectivecommunication.hh>
 
 // the following implementation is only available in case MPI is available
 #if HAVE_MPI
 #include <dune/common/parallel/mpicollectivecommunication.hh>
-#endif
+#endif // HAVE_MPI
+
+#else // DUNE_VERSION_NEWER(DUNE_COMMON, 2, 3)
+
+#include <dune/common/mpihelper.hh>
+#include <dune/common/collectivecommunication.hh>
+
+// the following implementation is only available in case MPI is available
+#if HAVE_MPI
+#include <dune/common/mpicollectivecommunication.hh>
+#endif // HAVE_MPI
+
+#endif // DUNE_VERSION_NEWER(DUNE_COMMON, 2, 3)
```

PR #172, opened 2015-09-15
Point2PointCommunicator: make code compile with ancient dune releases.

dr-robertk:
This PR fixes the issues brought up when using ancient dune releases.

### DIFF
--- a/dune/grid/common/p2pcommunicator.hh
+++ b/dune/grid/common/p2pcommunicator.hh
@@ -24,6 +24,9 @@
 #include <set>
 #include <map>
 
+#include <dune/common/version.hh>
+
+#if DUNE_VERSION_NEWER(DUNE_COMMON,2,3)
 #include <dune/common/parallel/mpihelper.hh>
 #include <dune/common/parallel/collectivecommunication.hh>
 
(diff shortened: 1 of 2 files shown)


dune/grid/common/p2pcommunicator.hh
```diff
@@ -223,6 +220,32 @@ public:
...
+    // return new tag number for the exchange messages
+    static int getMessageTag( const unsigned int increment )
+    {
+      static int tag = messagetag + 2 ;
+      // increase tag counter
+      const int retTag = tag;
+      tag += increment ;
+      // the MPI standard guaratees only up to 2^15-1
+      // this needs to be revised for the all-to-all communication
+      if( tag < 0 ) // >= 32767 )
```
bska: You can't say this.  There is nothing in the C++ standard that guarantees that signed integers wrap on overflow.  If you're really concerned about integer overflow, you need to say something along the lines of
```C++
if (tag > std::numeric_limits<int>::max() - increment) {
    // Reset to initial value
    tag = messagetag + 2;
}
else {
    tag += increment;
}
```

dr-robertk: The only thing I concerned is that this thing is positive and somewhat unique. The rest I don't care. The int is only there, because that's what MPI wants.

bska: > The int is only there, because that's what MPI wants.

Good.  You still can't use the check `tag < 0` because even the very expression `x + y` is undefined--rendering the program undefined too--if `x` and `y` are positive (signed) `int`s and the result of the addition overflows.

dr-robertk:
@bska: Satisfied now? You solution would have also violated the MPI standard. So now we follow that 
very closely and we also don't have to deal with overflow.

bska:
> @bska: Satisfied now?

Yes, much better.  I **suppose** you could write 32767 as `(1 << 15) - 1`, but whether or not that makes the statement more readable is in the eye of the beholder.

> Your solution would have also violated the MPI standard

Sure, but yours violated the C++ standard and that's much more serious.

atgeirr:
Compiles and test passes. Thanks! Merging.

PR #173, opened 2015-09-15
Silence warnings in p2pcommunicator_test.cc.

atgeirr:
Just silence a few warnings. Robert, could you check that the change to line 155 is what was intended in the first place?

I also noticed that this test does not use the same test framework (boost.test) as the others, but instead does `#undef NDEBUG` and uses `assert`. Is this made so on purpose, or did it just happen to be an old test we imported?


dr-robertk:
It should be ok, but I would rather put { } around the instance in line 149,150.

atgeirr:
> It should be ok, but I would rather put { } around the instance in line 149,150.

Done.

PR #177, opened 2015-09-17
Remove warnings.

dr-robertk:
This PR removes a couple of warnings due to unused variables or a newer DUNE version.

### DIFF
--- a/dune/grid/common/ZoltanGraphFunctions.cpp
+++ b/dune/grid/common/ZoltanGraphFunctions.cpp
@@ -178,7 +178,9 @@ void getCpGridEdgeList(void *cpGridPointer, int sizeGID, int sizeLID,
         *err = ZOLTAN_FATAL;
         return;
     }
+#ifndef NDEBUG
     int oldidx = 0;
+#endif
     int idx = 0;
 
     for( int cell = 0; cell < numCells;  cell++ )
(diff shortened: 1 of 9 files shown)


atgeirr:
Thanks! There are a few more I get though, could you also take them while at it?
```
/Users/atgeirr/opm-build-release/dune-cornerpoint/tests/p2pcommunicator_test.cc:71:24: warning: unused parameter 'link' [-Wunused-parameter]
  void pack( const int link, MessageBufferType& buffer )
                       ^
/Users/atgeirr/opm-build-release/dune-cornerpoint/tests/p2pcommunicator_test.cc:79:26: warning: unused parameter 'link' [-Wunused-parameter]
  void unpack( const int link, MessageBufferType& buffer )

/Users/atgeirr/opm-build-release/dune-cornerpoint/dune/grid/common/p2pcommunicator_impl.hh:488:68: warning: unused parameter 'buffer' [-Wunused-parameter]
    bool receivedMessage( MPI_Request& request, MessageBufferType& buffer )
```
I was curious about why I did not see the last one in dune-cornerpoint (only from opm-autodiff), and discovered that p2pcommunicator.hh is included inside a "warning suppression section". I think that should be removed since it is not an external header we cannot change.

dr-robertk:
Done.

atgeirr:
> Done

Almost... there is one in the p2pcommunicator_impl.hpp as well (see above).

dr-robertk:
Done.

PR #182, opened 2015-10-13
make it compile on GCC 4.4

andlaus:
again, the changes required to the actual library are pretty
trivial. the test function for the PolyhedralGrid is more interesting,
though: it causes an internal compiler error on GCC 4.4.

### DIFF
--- a/dune/grid/common/p2pcommunicator.hh
+++ b/dune/grid/common/p2pcommunicator.hh
@@ -52,7 +52,7 @@ namespace Dune
     typedef std::vector< char >  BufferType;
 
     mutable BufferType buffer_;
-    const double factor_;
+    double factor_;
     mutable size_t pos_;
 public:
     /** \brief constructor taking memory reserve estimation factor (default is 1.1, i.e. 10% over estimation )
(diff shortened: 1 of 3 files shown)


dune/grid/common/p2pcommunicator.hh
```diff
@@ -52,7 +52,7 @@ namespace Dune
     typedef std::vector< char >  BufferType;
 
     mutable BufferType buffer_;
-    const double factor_;
```
atgeirr: I am surprised that this fails to compile. Are you sure this change is needed?

andlaus: no, I'm not sure. maybe it could be fixed another way. The problem is that GCC 4.4 currently spits out the following warning:

```
/data/and/ctest/gcc-44/src/dune-cornerpoint/dune/grid/common/p2pcommunicator.hh:51: error: non-static const member ‘const double Dune::SimpleMessageBuffer::factor_’, can't use default assignment operator
```

(note that I also think this is a weird one, probably it's a GCC 4.4 bug.)

joakim-hove:
> (note that I also think this is a weird one, probably it's a GCC 4.4 bug.)

Guys - this is in my opinion *insane*  - we have agreed that we are targeting version 4.8 of gcc, and then you are considering/discussing patches required to work around possible bugs in gcc 4.4 -  I just don't get it?

atgeirr:
I have come to the conclusion that I will not merge this, I discussed what I'd consider in OPM/opm-core#903.

PR #183, opened 2015-10-19
Make Dune::gridcheck compile with CpGrid. 

dr-robertk:
I have updated CpGrid to comply with the interface changes that were made in Dune 2.4.

### DIFF
--- a/dune/grid/CpGrid.hpp
+++ b/dune/grid/CpGrid.hpp
@@ -574,7 +574,22 @@ namespace Dune
         /// \brief returns the number of boundary segments within the macro grid
         unsigned int numBoundarySegments() const
         {
-            return 0;
+            if( uniqueBoundaryIds() )
+            {
+                return current_view_data_->unique_boundary_ids_.size();
+            }
+            else
+            {
+                unsigned int numBndSegs = 0;
+                const int num_faces = numFaces();
+                for (int i = 0; i < num_faces; ++i) {
+                    cpgrid::EntityRep<1> face(i, true);
+                    if (current_view_data_->face_to_cell_[face].size() == 1) {
+                        ++numBndSegs;
+                    }
+                }
+                return numBndSegs;
+            }
         }
 
         // loadbalance is not part of the grid interface therefore we skip it.
--- a/dune/grid/CpGrid.hpp
+++ b/dune/grid/CpGrid.hpp
@@ -571,6 +571,12 @@ namespace Dune
             return 0;
         }
 
+        /// \brief returns the number of boundary segments within the macro grid
+        unsigned int numBoundarySegments() const
+        {
+            return 0;
+        }
+
         // loadbalance is not part of the grid interface therefore we skip it.
 
         /// \brief Distributes this grid over the available nodes in a distributed machine
--- a/dune/grid/CpGrid.hpp
+++ b/dune/grid/CpGrid.hpp
@@ -496,6 +496,13 @@ namespace Dune
           return leafIndexSet().geomTypes( codim );
         }
 
+        /// given an EntitySeed (or EntityPointer) return an entity object
+        template <int codim>
+        cpgrid::Entity<codim> entity( const cpgrid::EntityPointer< codim >& seed ) const
+        {
+            return cpgrid::Entity<codim>( *seed );
+        }
+
         /*  No refinement implemented. GridDefaultImplementation's methods will be used.
 
         /// \brief Mark entity for refinement
--- a/dune/grid/cpgrid/Entity.hpp
+++ b/dune/grid/cpgrid/Entity.hpp
@@ -274,10 +274,12 @@ namespace Dune
                 return *this;
             }
 
+            /// isValid method for EntitySeed
+            /// \return return true if seed is pointing to a valid entity
+            bool isValid () const;
+
         protected:
             const CpGridData* pgrid_;
-
-            bool valid() const;
         };
 
 
@@ -322,14 +324,14 @@ namespace Dune
             /// Const member by pointer operator.
             const Entity* operator->() const
             {
-                assert(Entity::valid());
+                assert(Entity::isValid());
                 return (this);
             }
 
             /// Const dereferencing operator.
             const Entity& operator*() const
             {
-                assert(Entity::valid());
+                assert(Entity::isValid());
                 return (*this);
             }
 
@@ -447,9 +449,9 @@ bool Entity<codim>::hasBoundaryIntersections() const
 }
 
 template <int codim>
-bool Entity<codim>::valid() const
+bool Entity<codim>::isValid() const
 {
-    return EntityRep<codim>::index() < pgrid_->size(codim);
+    return pgrid_ ? EntityRep<codim>::index() < pgrid_->size(codim) : false;
 }
 
 }}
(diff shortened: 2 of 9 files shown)


dune/grid/cpgrid/Entity.hpp
```diff
@@ -331,14 +348,14 @@ namespace Dune
 //          }
 
             /// Const member by pointer operator.
-            Entity* operator->() const
+            const Entity* operator->() const
             {
                 assert(Entity::valid());
                 return const_cast<EntityPointer*>(this); // const_cast-hack added because of error in vtkwriter.hh
```
atgeirr: If this hack is not needed anymore (I don't remember the vtkwriter bug but it has probably been fixed) we should remove the const_cast, at the same time you change the method signature. Same for the `operator*()`.

dune/grid/cpgrid/Iterators.hpp
```diff
@@ -113,6 +113,32 @@ namespace Dune
     } // namespace cpgrid
 } // namespace Dune
 
+namespace std
+{
+  template< int codim, Dune::PartitionIteratorType pitype >
```
atgeirr: Sorry for being annoying, but I'd like two formatting changes:
 - Use 4-indentation since that is what the rest of the file is like.
 - Use `typedef value_type* pointer`, the asterisk should be considered part of the pointer type, same goes for the reference `&`, in both places below.

tests/grid_test.cc
```diff
@@ -96,17 +97,10 @@ void testGrid( const GridView& gridView )
...
-    Opm::ParseMode parseMode;
-    const auto deck = parser.parseString(deckString , parseMode);
-
-    std::vector<double> porv;
-    typedef Dune::PolyhedralGrid< 3, 3 > Grid;
-    typedef Grid::LeafGridView GridView;
-    Grid grid(deck, porv);
-
+    typedef typename Grid::LeafGridView GridView;
 #if DUNE_VERSION_NEWER(DUNE_GRID,3,0)
```
atgeirr: I need to ask (again, sorry): is DUNE_VERSION_NEWER really NEWER_OR_EQUAL_TO? That is, this will trigger for version 3.0? Also, I assume that 3.0 is the current git master version, so this will not trigger for the 2.4 release. Is that intended, or is this something remaining from a time when we thought there would be no 2.4 release?

atgeirr:
It will be good to have these compilation tests active again for CpGrid, thanks for the fix! A few minor issues and question noted in the diff.

dune/grid/cpgrid/Entity.hpp
dr-robertk: I thought about it. Maybe I should not change the signature just before the release. 
However, CpGrid is quite out of date when it comes to recent dune interface changes. 
For dune 2.4 we can return an Entity here. I think the operator-> has vanish.

tests/grid_test.cc
dr-robertk: Ja, exactly. DUNE_VERSION_NEWER(3,0) means >= 3.0.

dr-robertk:
@atgeirr: I have double checked with opm-porsol and opm-upscaling and the dune 2.2 release. Everything compiles.

dr-robertk:
@atgeirr: the grid test revealed that indexInOutside in CpGrid is implemented wrong. The test check does the following: get the indexInOutside for an Intersection, then go to the neighbor, iterate over the Intersections and when you find the one with neighbor being the element you came from than make sure that indexInInside is the same as the previously obtained indexInOutside. Should this be fixed before the release? Didn't seem to bother anybody so far.

tests/grid_test.cc
atgeirr: If the indexInInside() issue can easily be fixed (without breaking the concept of the class...) please do so. If this is not a quick fix we should discuss what do do.

I still wonder if the 3.0 should be a 2.4 in the #if above?

dr-robertk:
My fixes of the gridcheck did not make it into 2.4. Therefore, 3.0 is the correct target. 
Of course, I'm testing with 2.4 since I locally patches the two lines in the gridcheck. But still, 3.0 is the correct number.

atgeirr:
> My fixes of the gridcheck did not make it into 2.4. Therefore, 3.0 is the correct target. 
> Of course, I'm testing with 2.4 since I locally patches the two lines in the gridcheck. But still, 3.0 is the correct number.

Thanks for the explanation!

atgeirr:
> The test check does the following: get the indexInOutside for an Intersection, then go to the neighbor, iterate over the Intersections and when you find the one with neighbor being the element you came from than make sure that indexInInside is the same as the previously obtained indexInOutside. Should this be fixed before the release? Didn't seem to bother anybody so far.

It worries me a little though. If nobody is bothered, perhaps we should disable the methods until we figure out the bug? I looked at them BTW, and could not figure out why it should be wrong. Do you know?

dr-robertk:
@atgeirr: I checked and the indexInOutside works correct. The reason for the test failing is the "incorrect" number of subEntities. This was previously implemented correctly but then set to zero (see Entity.hpp). Also, the methods geometryInInside and geometryInOutside should return an empty geometry instead of an exception. But that is just a matter of taste. The test works now also with CpGrid if I return 6 in the method subEntities for cc == 1.

dr-robertk:
I found two more missing methods. Is there a simple way to implement numBoundarySegments?

atgeirr:
> I found two more missing methods. Is there a simple way to implement numBoundarySegments?

I am not sure what it is supposed to do, so I don't know. I have some trouble finding the docs for the boundary segments, too. Was that the thing that replaced boundary ids?

dr-robertk:
I implemented a correct version of numBoundarySegments that makes the test work.

atgeirr:
> I implemented a correct version of numBoundarySegments that makes the test work.

Great! Is there more coming or do you consider it merge-ready?

dune/grid/CpGrid.hpp
```diff
@@ -564,6 +571,27 @@ namespace Dune
             return 0;
         }
 
+        /// \brief returns the number of boundary segments within the macro grid
+        unsigned int numBoundarySegments() const
```
atgeirr: It could be a surprise that this in O(n) in the number of faces, but I'll not worry about that now. I assume Dune make no run-time complexity guarantees for methods?

dr-robertk: If this is an issue we can store this number, since the grid is fixed anyway. For now I just want this to be a correct implementation.

dr-robertk:
I have one more thing, that takes care of the other PR but I'll include it here to save some time.

dune/grid/CpGrid.hpp
atgeirr: It's fine with me. BTW is the parallel case handled correctly? Not sure what is expected of the grid in that case, and it looks to me like the implementation here may count faces that are interior to the global grid, and on the boundary of the local grid.

dr-robertk:
I have implemented id's for Intersections since this will be included in the DUNE grid interface very soon. This also makes the IntersectionMapper obsolete. For both, CpGrid and PolyhedralGrid, the Intersection id is of course simply the face number.

dr-robertk:
The whole numBoundarySegments concept is bogus. It does not work in parallel, that is why we should actually not worry about this to much.

atgeirr:
> The whole numBoundarySegments concept is bogus

I sense a thorny Dune debate here... If so, I'm fine with merging as-is.

PR #188, opened 2015-10-23
Bump version and unify format of dune.module.

atgeirr:
After this, all code modules have the same formatting in dune.module, and the correct version name (for the master branch): "2016.04-pre".


atgeirr:
> parallelgrid and ALU3d should be removed.

Done. Merging.

PR #190, opened 2015-10-26
This completes the transition of identifying overlap cells with copy …

blattms:
…indices in dune-istl.

It was needed for the parallel AMG to behave as expected when
agglomerating data to fewer processes. In this case the overlap in the
dune-grid sense has to marked as copy in the dune-istl index sets.
This was already done for the index sets, but the manually crafted
remote index list unsed in CpGrid did not reflect the change and
still marked cells not interrior as overlap and not copy. While this
was not a problem for the communication within the grid interface, it
broke the communication within dune-istl. The latter is used in
opm-autodiff and opm-core and there is poses a serious problem.

This commit fixes the setup of the remote index lists.


blattms:
This is also a bugfix and it might make sense to included in the release.

atgeirr:
Thanks for the patch! I'll merge.

PR #191, opened 2015-11-05
Add Dune::MPIHelper::instance() at start of main().

atgeirr:
Since the CpGrid class makes MPI calls if HAVE_MPI is set, we must ensure that MPI is initialised. The MPIHelper can be used unconditionally since it will be a dummy without MPI (at least I think that is correct).

### DIFF
--- a/examples/grdecl2vtu.cpp
+++ b/examples/grdecl2vtu.cpp
@@ -113,6 +113,7 @@ void condWriteIntegerField(std::vector<double> & fieldvector,
 int main(int argc, char** argv)
 try
 {
+    Dune::MPIHelper::instance(argc,argv); // Dummy if no MPI.
 
     CpGrid grid;
 


atgeirr:
Is there a reason why this cannot be merged?

blattms:
On Fri, Nov 27, 2015 at 12:16:28AM -0800, Atgeirr Fl� Rasmussen wrote:
> Is there a reason why this cannot be merged?

Please change the line to
```
Dune::MPIHelper& helper = Dune::MPIHelper::instance(argc, argv);
```
and move it outside of the try clause to make sure that its Destructor
(and thus `MPI_Finalize`) is called last in the program.

atgeirr:
OK, just two questions:
 - I was a little unsure as to why it makes a difference to assign the return value (since it is not used), but realised that not assigning it will call the destructor immediately, is that what you thought about?
 - I thought it would be cleaned up last anyway since it is the first local variable, even if inside the try block. For example flow.cpp uses the same pattern, does that also need to be fixed?

andlaus:
> Please change the line to
 ```
Dune::MPIHelper& helper = Dune::MPIHelper::instance(argc, argv); 
```
and move it outside of the try clause to make sure that its Destructor (and thus `MPI_Finalize`) is called last in the program.

I'm a bit confused about how this could possibly make any difference: the _reference_ to the object returned by Dune::MPIHelper::instance is destroyed at the end of the try scope, but not the object itself; the object is a global variable ("singleton") which is always destroyed at program termination, regardless of the lifetime of the references or pointers to it. that's also why singletons should not be used if the order of destruction matters: if e.g. another singleton needs the MPIHelper in its destructor and the MPIHelper is destroyed before the other singleton, there will be a segfault on exit. (but on other machines/compilers/compiler flags it may work because the deconstruction ordering is different.)

(I just read @atgeirr's comment, I agree with him.)

blattms:
On Fri, Nov 27, 2015 at 04:51:37AM -0800, Atgeirr Fl� Rasmussen wrote:
> OK, just two questions:
>  - I was a little unsure as to why it makes a difference to assign the return value (since it is not used), but realised that not assigning it will call the destructor immediately, is that what you thought about?
>  - I thought it would be cleaned up last anyway since it is the first local variable, even if inside the try block. For example flow.cpp uses the same pattern, does that also need to be fixed?

Thanks for the questions. They made me realze that I  was talking complete non-sense and your original version is perfectly ok as
- MPI_Helper is a singleton (static local variable of type MPIHelper in instance method) and therefore gets cleaned up when the program exits.
- Holding a mutable reference to an object does not change the lifetime it.

I am sorry for the confusion. I must have been struggling with fixtures in boost::unit_test too much.

atgeirr:
Ok thanks, I'll self-merge this then.

PR #195, opened 2015-11-24
Honor transmissibilities when loadbalancing as edge weights

blattms:
It tells Zoltan to cut at faces with low transmissibilities rather than with high ones. For the latter we expect that the is better for the linear solvers to have them in one subdomain.

Note that this PR is mainly for information purposes and for people that want to try out the parallel version of flow (PR upcoming).
In the long run it might to turn out that we must change the weights. Neverthless this PR in itself is harmless and does not change sequential runs.

### DIFF
--- a/dune/grid/common/ZoltanGraphFunctions.hpp
+++ b/dune/grid/common/ZoltanGraphFunctions.hpp
@@ -129,6 +129,7 @@ public:
     /// \param pretendEmptyGrid True if we should pretend the grid and wells are empty.
     CombinedGridWellGraph(const Dune::CpGrid& grid,
                           Opm::EclipseStateConstPtr eclipseState,
+                          const double* transmissibilities,
                           bool pretendEmptyGrid);
 
     /// \brief Access the grid.
@@ -144,7 +145,11 @@ public:
     /// \brief Post process partitioning to ensure a well is completely on one process.
     /// \param[inout] parts The assigned partition numbers for each vertex.
     void postProcessPartitioningForWells(std::vector<int>& parts);
-    
+
+    double transmissibility(int face_index) const
+    {
+        return transmissibilities_ ? (1.0e18*transmissibilities_[face_index]) : 1;
+    }
 private:
     void addCompletionSetToGraph(std::set<int>& well_indices)
     {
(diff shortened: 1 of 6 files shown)


dune/grid/common/ZoltanGraphFunctions.hpp
```diff
@@ -144,7 +145,11 @@ class CombinedGridWellGraph
     /// \brief Post process partitioning to ensure a well is completely on one process.
     /// \param[inout] parts The assigned partition numbers for each vertex.
     void postProcessPartitioningForWells(std::vector<int>& parts);
-    
+
+    double transmissibility(int face_index) const
+    {
+        return transmissibilities_ ? (1.0e18*transmissibilities_[face_index]) : 1;
```
atgeirr: As I understand the current code, the weights are 1 for the faces, which will be replaced by this expression. The well weights are `std::numeric_limits<float>::max`, which makes me wonder: how is this dealt with in terms of scaling? Will it notice at all if one transmissibility is twice as large as another? Is the 1e18 factor something you have experimented with?

andlaus:
the idea is good, but a question springs to my mind: isn't it better to use the face area as edge weights? They are already included in the transmissibilities, but has the advantage that thin cells (which lead to very large transmissibilities) do not stick together too much....

atgeirr:
> has the advantage that thin cells (which lead to very large transmissibilities) do not stick together too much

Why would thin cells (assuming you mean in the z-direction) have large transmissibilities? Also, if the transmissibilities are large they *should* stick together.

bska:
> Why would thin cells (assuming you mean in the z-direction) have large transmissibilities?

Transmissibility scales with connection area.

atgeirr:
> Transmissibility scales with connection area.

Yes, I assumed he meant that they would stick in the thin dimension, but I guess I misunderstood: @andlaus did you mean to say that a (z-)stack of thin cells would stick together too much?

andlaus:
> Yes, I assumed he meant that they would stick in the thin dimension, but I guess I misunderstood: @andlaus did you mean to say that a (z-)stack of thin cells would stick together too much?

yes: my point was that transmissibilities are proportional to both the phase area and the distance of the cell centers to the phase centers and that I thus considered it possible that thin cells make the partitioning worse than it would be if just the face area was considered. (note that I'm actually not too sure what works better, it could be transmissibilities.)

PR #198, opened 2016-01-12
Adjust to recent changes in dune-grid.

dr-robertk:
This Pr fixes some minor issues which occur after some recent changes in dune-grid.

### DIFF
--- a/dune/grid/cpgrid/Indexsets.hpp
+++ b/dune/grid/cpgrid/Indexsets.hpp
@@ -36,7 +36,6 @@ along with OPM.  If not, see <http://www.gnu.org/licenses/>.
 #ifndef OPM_INDEXSETS_HEADER
 #define OPM_INDEXSETS_HEADER
 
-#include <dune/common/nullptr.hh>
 #include <dune/geometry/type.hh>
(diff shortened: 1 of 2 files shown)


dune/grid/cpgrid/Indexsets.hpp
```diff
@@ -36,7 +36,6 @@ along with OPM.  If not, see <http://www.gnu.org/licenses/>.
 #ifndef OPM_INDEXSETS_HEADER
 #define OPM_INDEXSETS_HEADER
 
-#include <dune/common/nullptr.hh>
```
bska: Is this related to changes in recent versions of Dune or simply pruning an unused header?

dr-robertk: This header is deprecated but also not needed since all newer compilers
support C++11.

bska: > This header is deprecated but also not needed since all newer compilers support C++11.

Okay, thank you for the clarification.

bska:
Looks good overall. Everything builds and all tests run when targeting older Dune releases (2.3.1 and 2.4.0 tested), so I'll merge into master.  Thanks a lot.

PR #203, opened 2016-02-04
PolyhedralGrid: Add default constructor for Entities.

dr-robertk:
This is needed to compile with DUNE >= 2.4. Surprising that this has not been discovered earlier.

### DIFF
--- a/dune/grid/polyhedralgrid/entity.hh
+++ b/dune/grid/polyhedralgrid/entity.hh
@@ -163,11 +163,11 @@ namespace Dune
     typedef typename Base :: EntitySeed EntitySeed;
     using Base :: codimension ;
 
-    explicit PolyhedralGridEntity ()
+    PolyhedralGridEntity ()
     : Base()
     {}
 
-    PolyhedralGridEntity ( ExtraData data )
+    explicit PolyhedralGridEntity ( ExtraData data )
     : Base( data )
     {}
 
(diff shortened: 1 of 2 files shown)


andlaus:
seems good and it fixes the build error which I've stumbled over with `PolyhedralGrid` recently. Please merge!

dune/grid/polyhedralgrid/entity.hh
```diff
@@ -157,7 +163,11 @@ namespace Dune
     typedef typename Base :: EntitySeed EntitySeed;
     using Base :: codimension ;
 
-    explicit PolyhedralGridEntity ( ExtraData data )
+    explicit PolyhedralGridEntity ()
```
bska: Default constructors probably don't need to be `explicit`.

dune/grid/polyhedralgrid/entity.hh
```diff
@@ -157,7 +163,11 @@ namespace Dune
...
+    : Base()
+    {}
+
+    PolyhedralGridEntity ( ExtraData data )
```
bska: Single-parameter constructors probably should be `explicit`.

dr-robertk:
Fixed explicit keyword mixup.

bska:
> Fixed explicit keyword mixup.

Thanks.  This is good to go.

PR #204, opened 2016-02-08
Travis support

jepebe:
Added support for TravisCI.
Renamed README to README.md and added build status image.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,43 @@
+language: cpp
+
+compiler:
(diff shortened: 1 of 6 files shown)


bska:
Overall I guess this is good.  I personally prefer to avoid `pushd` and `popd` in scripts favoring forking off subshells instead when needed, but I won't quibble over this now.

I'll merge into master.  Thanks a lot for the contribution.

PR #207, opened 2016-03-30
Fixes needed before opm output change

atgeirr:
Mostly avoiding the EclipseGridInspector, so that we don't have to make this module depend on opm-output.


atgeirr:
I intend to merge this, and the three opm-output related PRs tomorrow.

PR #210, opened 2016-04-06
Module renaming

atgeirr:
This updates code, Travis and Jenkins scripts to refer to opm-grid instead of dune-cornerpoint.

Requires OPM/opm-common#104.


atgeirr:
I see that this is not mergeable for some reason, I will try to figure out why and fix it.

PR #213, opened 2016-04-20
Using getInputGrid API from Parser to get ACTNUM info

pgdr:
Using new EclipseState.getInputGrid API in Parser to get complete ACTNUM info for EclipseGrid.

See OPM/opm-parser#769

### DIFF
--- a/dune/grid/cpgrid/readEclipseFormat.cpp
+++ b/dune/grid/cpgrid/readEclipseFormat.cpp
@@ -40,13 +40,16 @@
 
 #include "CpGridData.hpp"
 
+#include <dune/grid/common/GeometryHelpers.hpp>
+
 #include <opm/core/grid/cpgpreprocess/preprocess.h>
 #include <opm/core/grid/MinpvProcessor.hpp>
-#include <dune/grid/common/GeometryHelpers.hpp>
 #include <opm/core/utility/StopWatch.hpp>
+
+#include <opm/parser/eclipse/Deck/Deck.hpp>
+#include <opm/parser/eclipse/EclipseState/EclipseState.hpp>
 #include <opm/parser/eclipse/Parser/Parser.hpp>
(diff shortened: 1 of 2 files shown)


joakim-hove:
jenkins build this opm-parser=769 please

atgeirr:
I am not quite satisfied with this, as it makes the EclipseState over again. One issue is the extra time, another is that the default ParseContext is used (for Flow we use a non-default one). I'm looking at this right now to find a better solution.

pgdr:
I must admit that I don't know where you are referring to now, is it `readEclipseFormat.cpp` or `grid.hh`?

joakim-hove:
> I'm looking at this right now to find a better solution.

Well - in my opinion the obvious solution is to pass an `EclipseGrid` instance to the `CpGridData::processEclipseFormat()` and `createGrid()`methods - and remove the `Deck`argument? That way we shave off one `EclipseGrid`creation compared to the current situation.

pgdr:
... and in both cases, the API takes as input a `deck` and is oblivious to the current `ParseContext`.  If we create an `EclipseState` outside these functions, the best might be to change the API to take the state as an input?

atgeirr:
Mostly readEclipseFormat. But in general, I think that we can save this. The problem is:
 - The overload with the Deck argument is very convenient and used from multiple places.
 - That overload must create its own EclipseState with default ParseContext
 - In Flow (and possibly other programs) we already created the EclipseState, with the correct ParseContext.

The solution is actually quite simple: do not use the Deck overload from flow. This means changing opm-simulators, but that will be to the better. I will submit a PR to your branch there that should take care of the problem. With that, this can be merged.

atgeirr:
(I now read your comments)

We may change the API, but the current patch is helpful for those that do not need to go that far down.

I think there is already a quite complex chain of classes we have to deal with: create a Parser, create a Deck, create an EclipseState, get the EclipseGrid.

PR #215, opened 2016-04-21
Fix grid creation from incomplete decks.

atgeirr:
This is necessary for working with .grdecl files directly. That ability was temporarily lost by the merging of #213.

@joakim-hove I would appreciate if you could review and merge this (if ok).


joakim-hove:
This was how I perceived it; and since there is actually a test in opm-grid I am confident to merge this now!

pgdr:
Are there tests for this?  If not, could you add a test together with this PR?

Does DeckKeyword.getIntData give correct ACTNUM info (that is, even if BOXed)?

atgeirr:
> Does DeckKeyword.getIntData give correct ACTNUM info (that is, even if BOXed)?

No, but this can probably not be fixed for incomplete decks (i.e. grdecl files) in any case?

PR #218, opened 2016-04-30
Fixes computation of facetags

blattms:
For Norne the faceTag computed with UnstructuredGrid differed from the one with CpGrid. That cause for this was that we assumed that the cell at whose faces we look always comes first in faceCells. But actually these are ordered by their cell index. Therefore we need to search for the right location and check the cells orientation.

With these changes we get much more similar results for Norne with CpGrid than before. Albeit there is still a small difference to using UG.

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -60,6 +60,7 @@ list (APPEND TEST_SOURCE_FILES
   tests/cpgrid/distribution_test.cpp
 	tests/cpgrid/entityrep_test.cpp
 	tests/cpgrid/entity_test.cpp
+	tests/cpgrid/facetag_test.cpp
 	tests/cpgrid/geometry_test.cpp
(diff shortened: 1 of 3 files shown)


atgeirr:
I am happy to see this bug fixed. Tested with Norne on my system, differences in the results between CpGrid and UnstructuredGrid versions (flow_mpi and flow) are now insignificant. I especially appreciate the addition of a test (although it looks like it returns true for the original version of the code?), good work!

blattms:
On Sun, May 01, 2016 at 11:27:18PM -0700, Atgeirr Flø Rasmussen wrote:
> I especially appreciate the addition of a test (although it looks like it returns true for the original version of the code?), good work!

You are welcome. I did not want to break stuff. I still feel a bit
uneasy about the fact the problem only occured with Norne.

PR #220, opened 2016-06-10
update jenkins build scripts

akva2:
opm-output is now an upstream and not a downstream


atgeirr:
This looks good, does it not? I'll merge.

PR #225, opened 2016-06-22
fixing compilation due to depreciated WellPtr.

GitPaean:


joakim-hove:
Again an example that MPI is not fully integrated in build testing++. Thank you for fix.

PR #229, opened 2016-08-24
Update Intersection.cpp

birgerh:
When running in parallel, the method Intersection::neighbor() has to return "false" for intersections which do not have a neighbor in the current process. However, currently it seems to return "true" for intersections on the outside boundary of overlapping cells. We stumbled on this problem when we tried to run a Dumux simulation using cpgrid in parallel. The suggested change solved our problem.
Could please somebody have a look if the changes should be merged?

Regards Birger

### DIFF
--- a/dune/grid/cpgrid/Intersection.cpp
+++ b/dune/grid/cpgrid/Intersection.cpp
@@ -89,7 +89,7 @@ void Intersection::update()
                     cells_of_face[0].index()==std::numeric_limits<int>::max() ||
                     cells_of_face[1].index()==std::numeric_limits<int>::max();
                 if (has_no_nbcell) {
-                    nbcell_ = index_; // self is invalid value
+                    nbcell_ = std::numeric_limits<int>::max(); // neighbor is not within this process
                 } else {
                     assert(cells_of_face.size() == 2);
                     if (cells_of_face[0].index() == index_) {


atgeirr:
I am not an expert in the parallel dune grid interface, so I would need a little time to check this out. I hope @dr-robertk or @blattms can review the change?

blattms:
Will take a look today / tomorrow. (Vacation is over).

@birgerh Would be possible to have a small test case that fails without this PR in parallel?

birgerh:
You can use the test test_cc2pcornerpoint in dumux/test/porousmediaflow/2p/implicit. However, the grid should be larger, so that it runs on more than one level. You can use for example the attached grid, set the coarsenTarget of the AMG to 8 and run it on 8 processes.
[8x4x2.GRDECL.txt](https://github.com/OPM/opm-grid/files/451329/8x4x2.GRDECL.txt)

blattms:
Well that is not exactly what I meant. I meant a basic grid test (using just DUNE grid functionality) that triggers the error. That one could move up to dune-grid and serve as an additional grid test for future implementations and save other future users the debugging time that you invested. BTW thanks a lot for finding this bug.

birgerh:
Sorry, I am not aware how a grid test should look like. However, I have reduced my test case to a minimum and got rid of all Dumux functionalities:

```
#include "config.h"
#include <dune/grid/CpGrid.hpp>

int main (int argc, char** argv) try
{
	Dune::CpGrid grid;
	std::array<int, 3> dims={{8, 4, 2}};
	std::array<double, 3> size={{ 8.0, 4.0, 2.0}};
	grid.createCartesian(dims, size);
	typedef Dune::CpGrid::LeafGridView GridView;
	GridView gridView(grid.leafGridView());
	enum{dimWorld = GridView::dimensionworld};
	typedef typename GridView::ctype CoordScalar;
	typedef Dune::FieldVector<CoordScalar,dimWorld> GlobalPosition;
    typedef GridView::Codim<0>::Iterator ElementIterator;
    typedef typename GridView::IntersectionIterator IntersectionIterator;

	grid.loadBalance();
	ElementIterator endEIt = gridView.end<0>();
	for (ElementIterator eIt = gridView.begin<0>(); eIt != endEIt; ++eIt) {
	    IntersectionIterator isEndIt = gridView.iend(eIt);
	    for (IntersectionIterator isIt = gridView.ibegin(eIt); isIt != isEndIt; ++isIt)
	        {
	    	    if (isIt->neighbor())
	    	    {
	    	    	GlobalPosition distVec = eIt->geometry().center() -
	    	    			isIt->outside()->geometry().center();
	    	    	if (distVec.two_norm2()<1e-8) {
	    	    		DUNE_THROW(Dune::GridError, "Coordinates of an element and its neighbor are identical, "
	    	    				"please check the methods Intersection::neighbor() and Intersection::outside()");
	    	    	}
	    	    }
	        }
	    }

  // everything okay
  return 0;

}
catch (Dune::Exception &e) {
  std::cerr << e << std::endl;
  return 1;
} catch (...) {
  std::cerr << "Generic exception!" << std::endl;
  return 2;
}
```

This test will throw an exception when running in parallel. I hope this helps a little?

blattms:
Great. That is exactly what I was looking for.

birgerh:
Probably the number of grid cells could be better. At the moment the test throws the grid error only when it is running on 8 processes. A larger grid, e.g. 5x5x10, would be more flexible with respect to the number of processes.

atgeirr:
With the test provided, were you able to review the patch @blattms? Also, it  would be good to have the test in the module and run automatically.

blattms:
The patch is fine and should be merged. It fixes the error for the test that birger provided. You can find his test in PR #239. Please merge that one afterwards.

PR #232, opened 2016-09-13
Loadbalance returns set of defunct wells on process

blattms:
Previously, we tried to compute this information in the WellsManager of opm-core. In part this was not done correctly and in the situation were the eclipse file contains completion cells of a well that are not
contained in the global grid this is not even possible.

With this commit we change the loadbalancing routines such that they return the set of names of wells that are not associated with the process. This will then be passed by the simulator to the WellsManager. uing OPM/opm-core#1062.

I habe not really tested it, but it should be save to merge this PR even without the accompanying ones in opm-core, and opm-simulator

### DIFF
--- a/dune/grid/common/ZoltanPartition.cpp
+++ b/dune/grid/common/ZoltanPartition.cpp
@@ -27,11 +27,12 @@ namespace Dune
 {
 namespace cpgrid
 {
-std::vector<int> zoltanGraphPartitionGridOnRoot(const CpGrid& cpgrid,
-                                                const Opm::EclipseStateConstPtr eclipseState,
-                                                const double* transmissibilities,
-                                                const CollectiveCommunication<MPI_Comm>& cc,
-                                                int root)
+std::pair<std::vector<int>, std::vector<int> >
+zoltanGraphPartitionGridOnRoot(const CpGrid& cpgrid,
+                               const Opm::EclipseStateConstPtr eclipseState,
+                               const double* transmissibilities,
+                               const CollectiveCommunication<MPI_Comm>& cc,
+                               int root)
 {
     int rc = ZOLTAN_OK - 1;
     float ver = 0;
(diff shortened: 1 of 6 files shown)


bska:
jenkins build this opm-core=1062 please

dune/grid/common/ZoltanPartition.cpp
```diff
@@ -122,11 +128,62 @@ std::vector<int> zoltanGraphPartitionGridOnRoot(const CpGrid& cpgrid,
...
+        std::vector<MPI_Request> reqs(cc.size(), MPI_REQUEST_NULL);
+        my_well_indices = wells_on_proc[root];
+        for ( int i=0; i < cc.size(); ++i )
+        {
+            if(i==root)
+            {
+                continue;
+            }
+            MPI_Isend(wells_on_proc[i].data(), wells_on_proc[i].size(),
+                      MPI_INT, i, 267553, cc, &reqs[i]);
```
atgeirr: What is the magic number 267533?

dune/grid/common/ZoltanPartition.cpp
```diff
@@ -122,11 +128,62 @@ std::vector<int> zoltanGraphPartitionGridOnRoot(const CpGrid& cpgrid,
...
+    // free space allocated for zoltan.
     Zoltan_LB_Free_Part(&exportGlobalGids, &exportLocalGids, &exportProcs, &exportToPart);
     Zoltan_LB_Free_Part(&importGlobalGids, &importLocalGids, &importProcs, &importToPart);
     Zoltan_Destroy(&zz);
-    return parts;
+
+    cc.broadcast(&parts[0], parts.size(), root);
+    std::vector<int> my_well_indices;
+
+    if( !pretendEmptyGrid)
```
atgeirr: Could you explain what this section is doing?

dune/grid/common/ZoltanPartition.cpp
blattms: It is a tag to identify the communication (matching send and receives) to prevent mixups.

dune/grid/common/ZoltanPartition.cpp
blattms: I assume your are talking about the if statement. `!pretendEmptyGrid` is only true on the root process that has the global grid during loadbalancing. I gues a more speaking name would be appreciated for readability? `hasGlobalGrid` or `isRootProcess`?
Anyway that process sends all the well information to the other processes.

dune/grid/common/ZoltanPartition.cpp
atgeirr: Then I think it should be named since it is used in two places, also a name would make it clear that it is essentially a random unique number.

dune/grid/common/ZoltanPartition.cpp
blattms: I correct myself. All the processes know the global grid but except one of them all others pretend to not have this information. Thus before load balancing for the load balancer one processes has the whole grid and all the others have an empty partition of the whole grid.

blattms: Added a commit that uses more speaking variable names.

dune/grid/common/ZoltanPartition.cpp
blattms: done with cd80060

PR #233, opened 2016-09-16
Really make Zoltan quiet.

blattms:
Setting the debug level to zero and then increasing three lines down makes ZOLTAN shout out all parameters used. This commit remove the second setting and results in  debug level 0 (quiet).


atgeirr:
Good, merging!

PR #234, opened 2016-09-16
Use manual main method to prevent segmentation faults for parallel boost::unit_test.

blattms:
For the parallel test using MPI experienced weired segmentation occurred
during program exit in some scenarios. Running make test worked. Running
a single test with mpirun worked. Running a single test without mpirun
caused a segmentation fault after all tests ran and the program exited.
Something like "memory access violation at address: 0x00000008: no mapping at fault address".

I guess the problem was that MPI_Finalize was not the last thing that ran
in the program.

This commit changes all the tests such that they have their own main method
with MPI_Init first and MPI_Finalize last. The unit tests are started
manually in between. This resolved the problems on my machine.

### DIFF
--- a/tests/cpgrid/distribution_test.cpp
+++ b/tests/cpgrid/distribution_test.cpp
@@ -6,6 +6,7 @@
 #define BOOST_TEST_DYN_LINK
 #endif
 #define BOOST_TEST_MODULE DistributedCpGridTests
+#define BOOST_TEST_NO_MAIN
 #include <boost/test/unit_test.hpp>
 
 #include <dune/grid/CpGrid.hpp>
@@ -329,4 +330,18 @@ BOOST_AUTO_TEST_CASE(distribute)
         grid.gatherData(gather_gid_set_data);
 
     }
+}
+
+
+bool
+init_unit_test_func()
+{
+    return true;
+}
+
+int main(int argc, char** argv)
+{
+    Dune::MPIHelper::instance(argc, argv);
+    boost::unit_test::unit_test_main(&init_unit_test_func,
+                                     argc, argv);
 }
--- a/tests/cpgrid/entity_test.cpp
+++ b/tests/cpgrid/entity_test.cpp
@@ -41,6 +41,7 @@
 
 
 #define BOOST_TEST_MODULE EntityTests
+#define BOOST_TEST_NO_MAIN
 #include <boost/test/unit_test.hpp>
 #include <sstream>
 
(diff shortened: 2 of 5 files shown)


blattms:
@akva2 I guess this should fix [this problem](http://80.240.129.131:8080/job/opm-grid-PR-builder/8/testReport/junit/%28root%29/mpi/partition_iterator/)

blattms:
I guess merging this would prevent me and others from getting emails from the jenkins server about unstable builds. http://80.240.129.131:8080/job/opm-grid/23/

akva2:
you were emailed to markus at dr-blatt.de

akva2:
oh sorry, misread.

akva2:
jenkins build this please

akva2:
@blattms still not entirely fixed apparently.

akva2:
right. it kind of makes sense. it enters the function zoltanGraphPartitionGridOnRoot() with 'eclipseState == nullptr' and https://github.com/OPM/opm-grid/blob/master/dune/grid/common/ZoltanPartition.cpp#L68 evaluates to true. 

the if() at https://github.com/OPM/opm-grid/blob/master/dune/grid/common/ZoltanPartition.cpp#L110
does not pass, and thus https://github.com/OPM/opm-grid/blob/master/dune/grid/common/ZoltanPartition.cpp#L112 is not executed leading to ```wells_on_proc`` being empty.

then https://github.com/OPM/opm-grid/blob/master/dune/grid/common/ZoltanPartition.cpp#L144 is entered, leading to the unconditional deref at https://github.com/OPM/opm-grid/blob/master/dune/grid/common/ZoltanPartition.cpp#L147 causing a segfault and/or random behavior (depending on build configuration).

the unconditional deref at https://github.com/OPM/opm-grid/blob/master/dune/grid/common/ZoltanPartition.cpp#L172 makes me wonder if we should even be here with a null eclipseState.

in either case my conclusion is logic error.

blattms:
I stumbled over this when working on #235. Will check whether it is really fixed by my changes.

blattms:
jenkins build this please.

akva2:
please trigger jenkins again, you weren't on opm-grid list for some reason (only opm-simulators).

blattms:
Don't have to. Checks are running as I rebased and pushed --force.

akva2:
travis doesn't build with mpi afaik so that wouldn't help. but i managed to sneak in your name on the jenkins whitelist before the poll cycle ran.

blattms:
jenkins build this please

atgeirr:
Seems all tests are green: will you merge if you deem it ready, @akva2?

Issue #235, opened 2016-09-20
Compile failure with MPI but no ZOLTAN


atgeirr:
I get the following error:
```
/Users/atgeirr/opm-build-debug/opm-grid/dune/grid/cpgrid/CpGrid.cpp:129:40: error: use of undeclared identifier 'part_and_wells'
    return std::make_pair(true, get<1>(part_and_wells));
```
The part_and_wells variable is only defined in the HAVE_ZOLTAN part of the code.

GitPaean:
I did not see this when compiling with the master branch yesterday and today.

GitPaean:
Sorry, I misread your title and did not see `NO ZOLTAN`.

blattms:
Thanks for the report. I am on it.

PR #236, opened 2016-09-21
Fix compile failure for mpi without zoltan

blattms:
Up to now everything worked whenever ZOLTAN was installed and eclipse state was passed to the load balancing method. Unfortunately, I missed checking the parallel code without MPI. With this PR that code works now too and for the first time calculates the well partitioning, too. It should be fully functional but load balancing might be pure as it works on the cartesian grid.

As a side effect we also fix an issue when loadbalancing is used without an eclipse state. That problem was discussed by @akva2 in #234

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -116,6 +116,7 @@ list (APPEND PUBLIC_HEADER_FILES
 	dune/grid/cpgrid/PartitionTypeIndicator.hpp
 	dune/grid/cpgrid/PersistentContainer.hpp
 	dune/grid/common/CartesianIndexMapper.hpp
+	dune/grid/common/WellConnections.hpp
 	dune/grid/common/ZoltanGraphFunctions.hpp
 	dune/grid/common/ZoltanPartition.hpp
 	dune/grid/polyhedralgrid/capabilities.hh
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -39,6 +39,7 @@ list (APPEND MAIN_SOURCE_FILES
 	dune/grid/cpgrid/writeSintefLegacyFormat.cpp
 	dune/grid/common/GeometryHelpers.cpp
 	dune/grid/common/GridPartitioning.cpp
+	dune/grid/common/WellConnections.cpp
 	dune/grid/common/ZoltanGraphFunctions.cpp
 	dune/grid/common/ZoltanPartition.cpp
 	)
--- a/dune/grid/common/WellConnections.cpp
+++ b/dune/grid/common/WellConnections.cpp
@@ -36,6 +36,13 @@ WellConnections::WellConnections(const Opm::EclipseStateConstPtr eclipseState,
                                  const std::array<int, 3>& cartesianSize,
                                  const std::vector<int>& cartesian_to_compressed)
 {
+    init(eclipseState, cartesianSize, cartesian_to_compressed);
+}
+
+void WellConnections::init(const Opm::EclipseStateConstPtr eclipseState,
+                           const std::array<int, 3>& cartesianSize,
+                           const std::vector<int>& cartesian_to_compressed)
+{
     std::vector<const Opm::Well*>  wells  = eclipseState->getSchedule()->getWells();
     int last_time_step = eclipseState->getSchedule()->getTimeMap()->size()-1;
     well_indices_.resize(wells.size());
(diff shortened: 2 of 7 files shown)


blattms:
Closes #235.

akva2:
jenkins build this please

akva2:
jenkins happy.

blattms:
yeah!

akva2:
now my next worry. you say the bug was only triggered if zoltan is *not* around ?

blattms:
the compile bug #235 the @atgeirr reported was there only when ZOLTAN was not around. 

The bug that you mentioned and I fixed on the way was only apparent when ZOLTAN was there as that code path used the eclipse state. Without ZOLTAN eclipse state was simply neglected before this PR.

blattms:
Well the search space is rather larger and seems to be growing.

akva2:
thanks for the explanation. i can chill then, once this lands all is well in happy-land.

CMakeLists_files.cmake
```diff
@@ -39,6 +39,7 @@ list (APPEND MAIN_SOURCE_FILES
 	dune/grid/cpgrid/writeSintefLegacyFormat.cpp
 	dune/grid/common/GeometryHelpers.cpp
 	dune/grid/common/GridPartitioning.cpp
+	dune/grid/common/WellConnections.cpp
 	dune/grid/common/ZoltanGraphFunctions.cpp
```
atgeirr: It seems the header file has not been added, only the cpp file.

dune/grid/common/WellConnections.hpp
```diff
@@ -0,0 +1,136 @@
...
+#include <opm/parser/eclipse/EclipseState/EclipseState.hpp>
+#include <opm/parser/eclipse/EclipseState/Schedule/Well.hpp>
+namespace Dune
+{
+namespace cpgrid
+{
+
+/// \brief A class calculating and representing all connections of wells.
+///
+/// Wells are identified by their position as ecported by the eclipse file.
```
atgeirr: exported, expected or reported?

dune/grid/common/WellConnections.cpp
```diff
@@ -0,0 +1,197 @@
...
+                                             [](const std::pair<int,std::size_t>& p1,
+                                                const std::pair<int,std::size_t>& p2){
+                                                 return ( p1.second > p2.second );
+                                             })->first;
+            std::cout << "Manually moving well " << well->name() << " to partition "
+                      << new_owner << std::endl;
+
+            for ( size_t c = 0; c < well_completions.size(); c++ )
+            {
+                parts[c] = new_owner;
```
atgeirr: Is this correct? Should it not rather be `parts[well_completions[c]] = new_owner`? Might be simpler if using range-for as on line 101.

dune/grid/common/WellConnections.cpp
```diff
@@ -0,0 +1,197 @@
...
+#ifdef HAVE_MPI
+std::unordered_set<std::string>
+computeDefunctWellNames(const std::vector<std::vector<int> >& wells_on_proc,
+                        const Opm::EclipseStateConstPtr eclipseState,
+                        const CollectiveCommunication<MPI_Comm>& cc,
+                        int root)
+{
+    std::vector<const Opm::Well*>  wells  = eclipseState->getSchedule()->getWells();
+    std::vector<int> my_well_indices;
+    const int well_information_tag = 267553;
```
atgeirr: If this is required to be the same tag as in the other place it was used, the name must be in a header. Otherwise (it is independent) this is fine of course.

atgeirr:
Compiles and tests fine (with no warnings even), but I think there is a missing file, and a bug as noted in the review above. Will merge when fixed.

dune/grid/common/WellConnections.hpp
blattms: Should be "exported", fixed.

dune/grid/common/WellConnections.cpp
blattms: Nice catch. Fixed as proposed.

dune/grid/common/WellConnections.cpp
blattms: There should be only this place. The other place should be gone.

blattms: > There should be only this place. The other place should be gone.

with commit 10db5e86ec

atgeirr:
Everything fixed, tests pass, merging.

Thanks for the fix!

blattms:
thanks @atgeirr. As often your precise review unconvered some serious bugs.Keep it up.

PR #238, opened 2016-09-26
Rename variable for clarity and to silence a warning.

atgeirr:
Will self-merge, as it is a trivial change.


blattms:
Thanks for the improvement. I need to get better with names.

PR #244, opened 2016-10-18
[bugfix][CpGridData] make code compile when dune-istl is not found.

dr-robertk:
When dune-istl is not found successfully, opm-grid does not compile. his fixes this issue.


atgeirr:
It had not occurred to me that istl-less dune might be a conceivable configuration, but I guess you are right, and we should not require the module.

PR #246, opened 2016-10-20
Add pointer deref forgotten in shared_ptr purge.

atgeirr:
Code in question is the MPI-but-not-Zoltan part. All derefs occur inside if-blocks guaranteeing that the pointer is non-null.

Will self-merge since it is rather trivial, and a compile error showstopper.


jokva:
Oh boy, that's a configuration too? My bad.

PR #251, opened 2016-11-16
fixing some warnings.

GitPaean:
Silencing some warnings from `mpi.h`, and a few other warning silencing.

### DIFF
--- a/dune/grid/common/ZoltanGraphFunctions.cpp
+++ b/dune/grid/common/ZoltanGraphFunctions.cpp
@@ -167,7 +167,7 @@ void getNullEdgeList(void *cpGridPointer, int sizeGID, int sizeLID,
     (void) cpGridPointer; (void) sizeGID; (void) sizeLID; (void) numCells;
     (void) globalID; (void) localID; (void) numEdges; (void) nborGID;
     (void) nborProc; (void) wgtDim; (void) ewgts;
-    err = ZOLTAN_OK;
+    *err = ZOLTAN_OK;
 }
 
 void getCpGridEdgeList(void *cpGridPointer, int sizeGID, int sizeLID,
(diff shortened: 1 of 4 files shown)


dune/grid/common/ZoltanGraphFunctions.cpp
```diff
@@ -167,7 +167,7 @@ void getNullEdgeList(void *cpGridPointer, int sizeGID, int sizeLID,
     (void) cpGridPointer; (void) sizeGID; (void) sizeLID; (void) numCells;
     (void) globalID; (void) localID; (void) numEdges; (void) nborGID;
     (void) nborProc; (void) wgtDim; (void) ewgts;
-    err = ZOLTAN_OK;
```
andlaus: hu? how could that code have ever worked?

bska: One possible mechanism is if `ZOLTAN_OK` is `#define`d as zero (can't check right now) in some header.  In that case the statement
```C++
err = ZOLTAN_OK;
```
makes `err` into a null pointer within the function body and there's no change visible in the caller.  If, moreover, the caller already initialises `*err` as `ZOLTAN_OK` before calling into the call-back then the call-back is a big no-op.

bska: > so does that boil down to: it worked because of unlucky coincidences?

Quite possibly, yes.

bska:
jenkins build this please

dune/grid/common/ZoltanGraphFunctions.cpp
blattms: Sorry, did not look further down. Fix is already there.

atgeirr:
Good! (Warnings rule!)

PR #253, opened 2016-11-22
update downstream list

akva2:
ewoms moved in the hierarchy


atgeirr:
Feel free to merge all of these when green.

akva2:
yeah, will just have to run local tests first on at least one module first :)

akva2:
local test passed fine. should be good enough, jenkins is a rather busy atm.

PR #255, opened 2016-11-29
Add grid things from opm-core

atgeirr:
This imports the grid things (with history) from opm-core. It is intended to replace #254. No paths have been changed, that is better left for a separate PR.

There is one known missing thing: some tests should also have been moved, but I forgot. I'll do that in a later PR, these ones do all the important things.


dr-robertk:
Seems fine and the opm module chain compiles including tests. The only thing I noticed was that test_ug.cpp should probably also be moved to opm-grid.

atgeirr:
> The only thing I noticed was that test_ug.cpp should probably also be moved to opm-grid.

I know, there might be others as well. I'll do them next!

PR #257, opened 2016-11-30
fix some annoying clang specific warnings

andlaus:
GCC does not complain about those missing brackets, but it accepts
their presence without lamenting about them...


atgeirr:
I'd like to merge this after #258, just in case.

PR #258, opened 2016-11-30
Add files from opm-core

atgeirr:
These files were not moved in the big grid code transfer, but belong here.


atgeirr:
A companion PR in opm-core will come shortly.

bska:
We've completed a Jenkins run of all downstream modules without issue.  Since these commits were previously approved, I think this should go in.

PR #259, opened 2016-12-28
Refactor stopwatch to use std::chrono

atgeirr:
This removes the boost dependency from this class. Split in two commits, the first refactors and the second replaces boost with std::chrono usage.

### DIFF
--- a/opm/core/utility/StopWatch.cpp
+++ b/opm/core/utility/StopWatch.cpp
@@ -44,7 +44,7 @@ namespace Opm
 
     namespace time
     {
-        using Duration = boost::posix_time::time_duration;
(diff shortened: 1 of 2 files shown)


atgeirr:
A note: due to transitive includes (probably) this change caused a compile failure in some old opm-upscaling stuff. A tiny PR for that is coming.

atgeirr:
> non-historical reason to keep that class?

I intend to move it to opm-common, to make it usable from more modules. Other than that it is mostly historical (work required to replace it may not be worth it).

atgeirr:
Thanks for the review @andlaus, merging.

andlaus:
> work required to replace it may not be worth it.

I agree.

PR #260, opened 2016-12-29
Silence unused arguments warning.

atgeirr:
Doing a sweep looking at warnings in several modules.


atgeirr:
Trivial fix, tests successful, will self-merge.

PR #262, opened 2017-01-19
Make opm-common and opm-parser optional to opm-grid.

dr-robertk:
This PR makes opm-common and opm-parser optional to opm-grid. When the opm toolchain is not present, then the DUNE build system is used to build the library. 

In case the OPM tool chain is available everything stays the same. 
The benefit is that for testing of the grids with other DUNE modules or DuMuX this is much simpler without opm-parser and the dependencies. 

This PR also fixes the build of the module with DUNE 2.5. Testing has been done with DUNE 2.4 and DUNE 2.5.


atgeirr:
I am a bit skeptical about this, as it seems to me it would prevent me from using opm-common features in this module, or at least in certain parts of it? That would be such things as OpmLog etc.

Some additional questions:
 - I assume this changes no behaviour whatsoever in the "normal" case? Can you verify that?
 - Does it work with Dune 2.3 still?
 - The Travis error seems real (complaining about missing opm-common), so there seems to be something wrong about the "can make do without opm-common" parts.

dr-robertk:
@atgeirr: Why, if you want to use features from opm-common then everything stays the same as I have pointed out. The default should be to use opm-common and opm-parser. The fallback to DUNE is just in case these modules are not available.

dr-robertk:
@atgeirr: I tested the opm-simulators with this branch. Seemed fine. But some other people should try this before merging.

dr-robertk:
@atgeirr: DUNE 2.3 should still work, I haven't tested it though.

atgeirr:
> Why, if you want to use features from opm-common then everything stays the same as I have pointed out. The default should be to use opm-common and opm-parser. The fallback to DUNE is just in case these modules are not available.

Perhaps a better question is: what functionality of which classes remain available to use in the case that I don't have opm?

dr-robertk:
@atgeirr: I guess everything except OpmLog (opm-common) and everything from opm-parser. If these things are wanted then of course these modules need to be installed.

dr-robertk:
I think you misunderstood the purpose. The purpose is not to get rid of opm-common or opm-parser dependency. The main reasons is that for someone who wants to just use the grids, not ecl stuff required, then with this patch it's a lot easier. At least that is what I found and I guess DuMuX people will confirm that.

andlaus:
I think that in principle it is good to be able to compile opm-grid opm-{common,parser}. It is unclear to me how a useful grid object can be created without the parser, though. I suppose that was also @atgeirr's main concern?

dr-robertk:
The next step was to add some of DGF features as well as a grid factory. The code already exists and works. I just need to setup a PR. So basically the opm-parser is not necessarily needed to construct at least a PolyhedralGrid. I'm not so sure about CpGrid though.

dr-robertk:
But a closer looks reveals that even CpGrid can be constructed from DGF files. So no problem.

andlaus:
ah, ok. I wasn't aware of this...

atgeirr:
> But a closer looks reveals that even CpGrid can be constructed from DGF files. So no problem.

OK, so I see that it can be usable, then. Let me think a little about it, though. It could be easy to inadvertently destroy the "can be used without opm-common" property if you are not careful. I don't have time to review right now, hope it is not time critical?

dr-robertk:
It's not critical and I think its ok since the there is no heavy development on this module anyway.

atgeirr:
@dr-robertk do you want this to be part of the release?

dr-robertk:
If the tests work it could be part of the release as it also contains some fixes for DUNE 2.5. I have not idea why the travis checks fail.

dr-robertk:
@andlaus I have compiled with dune 2.3, 2.4 and 2.5 and no problem in the default setting, that is that the opm tool chain is used. I discovered one problem when only using dune modules. The dune.module file does not contain the dune-istl dependency even though code from dune-istl is used in CpGrid.cpp.

andlaus:
> The dune.module file does not contain the dune-istl dependency even though code from dune-istl is used in CpGrid.cpp.

that's what I would call a surprise. anyway, I'd be grateful if you manage to merge this PR be today. (but I will an exception for it if merging slips to tomorrow.)

dr-robertk:
I would first run some jenkins on just about everything with this PR.

andlaus:
jenkins build this with downstream please

andlaus:
hm, something seems to be still broken. jenkins' complaint is:

```
Building main module opm-grid=origin/pr/262/merge configuration=mpi

 /---------------------------------------------------------------------------------\
 |  Could not locate the opm build macros. The opm build macros                    |
 |  are in a separate repository - instructions to proceed:                        |
 |                                                                                 |
 |    1. Clone the repository: git clone git@github.com:OPM/opm-common.git         |
 |                                                                                 |
 |    2. Run cmake in the current project with -DOPM_COMMON_ROOT=<path>/opm-common |
 |                                                                                 |
 |    3. If only using DUNE not further action needed!                             |
 |                                                                                 |
 \---------------------------------------------------------------------------------/
```

I'd like to get this into the release, but since I'm not a maintainer for opm-grid someone else must press green once Jenkins is satisfied.

dr-robertk:

dr-robertk:
jenkins build this with downstream please

dr-robertk:
Hm, don't know why it's still failing. Maybe then it should not go in.

blattms:
Maybe they are failing because Geometry::integrationElement now always return zero instead of computing the correct value? See https://github.com/dr-robertk/opm-grid/blob/ecfce2daa6dd3b5153fa88ac24698d68cef4eee1/dune/grid/cpgrid/Geometry.hpp#L217-L225

dr-robertk:
@blattms Thanks for pointing this out. That was a development leftover.

dr-robertk:
jenkins build this downstream please

dr-robertk:
Seems like this is good to go. I recommend backporting but that should be decided by the RM.

dr-robertk:

atgeirr:
Thanks for the effort, I'll merge this now (finally). I hope it is not a problem that it did not make it into the release.

PR #263, opened 2017-01-19
Provide and interface for real scattering/gathering cell data.

blattms:
The current `scatterData` method relies on the global grid being stored on all processes. The `gatherData` code is a bit complicated. In addition we might need to able to gather data arrays (and not complete `ReservoirState` together with the well information) in `flow_ebos`. Hopefully, we can shrink the code of `ParallelOutputState` with this quite a lot. The current implementation works just for data attached to cells.

### DIFF
--- a/dune/grid/CpGrid.hpp
+++ b/dune/grid/CpGrid.hpp
@@ -1194,9 +1194,32 @@ namespace Dune
 
         /// \brief Get an interface for gathering/scattering data with communication.
         ///
+        /// Scattering means sending data from the indices of the global grid on
+        /// process 0 to the distributed grid on all ranks independent of the grid.
+        /// Gathering is the other way around.
         /// The interface can be used with VariableSizeCommunicator and a custom
         /// index based data handle to scatter (forward direction of the communicator)
-        // and gather data (backward direction of the communicator).
+        /// and gather data (backward direction of the communicator).
+        /// Here is a small example that prints the received values when scattering:
+        /// \code
+        /// struct Handle{
+        ///   typedef int DataType;
+        ///   const std::vector<int>& vals;
+        ///   bool fixedsize() { return true; }
+        ///   size_t size() { return 1; }
+        ///   void gather(auto& B buf, size_t i)[ buf.write(vals[i]); }
+        ///   void scatter(auto& B buf, size_t i) {
+        ///     int val;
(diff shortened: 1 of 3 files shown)


dune/grid/CpGrid.hpp
```diff
@@ -1071,6 +1074,19 @@ namespace Dune
...
 
+        /// \brief The type of the map describing communication interfaces.
+        typedef VariableSizeCommunicator<>::InterfaceMap InterfaceMap;
+
+        /// \brief Get an interface for gathering/scattering data with communication.
+        ///
+        /// The interface can be used with VariableSizeCommunicator and a custom
+        /// index based data handle to scatter (forward direction of the communicator)
+        // and gather data (backward direction of the communicator).
+        const InterfaceMap& cellScatterGatherInterface()
```
andlaus: Maybe it is just me, but I don't really understand the concepts of `InterfaceMap` and `VariableSizeCommunicator<>`. is it possible to explain their purpose in one or two sentences?

andlaus:

andlaus:
travis fails. This looks to be a Dune 2.3 issue...

blattms:
Nope. It is my favorite "Why should anyone run/test code without MPI
support" assumption that proves wrong as always.

dune/grid/CpGrid.hpp
```diff
@@ -47,6 +47,12 @@
 // Warning suppression for Dune includes.
 #include <opm/common/utility/platform_dependent/disable_warnings.h>
 
+#include <dune/common/version.hh>
+
+#if HAVE_MPI && DUNE_VERSION_NEWER(DUNE_GRID, 2, 3)
```
andlaus: I suppose that ´DUNE_VERSION_NEWER(DUNE_GRID, 2, 3)` is superfluous because the build system hopefully already ensures this. If you like to keep it around, that's fine with me, too... (dune.module still needs to be modified to make this requirement explicit, though...)

atgeirr:
Code looks ok to me (if a little hard to understand without comments... interface docs are fine though), it compiles and tests correctly, and it does not change any existing behaviour. I've run a local test to ensure that it does not cause any failures in construction (the only place this PR could cause trouble), so I'll merge.

blattms:
Thanks!

PR #264, opened 2017-01-20
update the dune.module file

andlaus:
this explicitly sets the minimum dune version to 2.3 in the `dune.module` file.


andlaus:
note that this patch also fixes a small indentation problem which newish GCC versions warn about...

atgeirr:
> note that this patch also fixes a small indentation problem which newish GCC versions warn about...

I actually had to look at the change *many* times before I understood what was happening. Braces I tell you...

andlaus:
I think that this warning was added to GCC specifically to detect the "goto fail;" class of bugs. not sure if braces would have helped, though...

PR #266, opened 2017-01-20
BUGFIX fix areaNormal computations for faces with nodes > 4

totto82:
With these bugfixes TRANX(YZ) matches ecl excellently for all cells not effected by MINPV and PINCH for Norne and Model 2.


andlaus:
To me it seems like this does not require any changes to the simulator, right?

totto82:
This should be merged asap since it fixes a obvious bug. Especially since the upstream PR in OPM/ewoms#144 is already merged.

PR #267, opened 2017-03-10
Add Simple Facility for Repairing ZCORN Vectors

bska:
This change-set adds and uses a simple class that ensures that ZCORN values honour certain consistency relations.  In particular, we ensure that ZCORN values are always depth (increasing downwards, for increasing layer IDs) and furthermore that

 * Cell's top corner is not below lower corner on same pillar
 * Cell's bottom corner is not below cell's lower neighbour's upper corner on same pillar.

This, in turn, restores our ability to use periodic boundary conditions in the [single-phase permeability upscaling facility](https://github.com/OPM/opm-upscaling/blob/master/examples/upscale_singlephase.cpp) of module opm-upscaling on the `flaser.grdecl` case.

As an aside, we also ensure that the low-level CP grid analysis considers the relevant version of ZCORN when determining whether or not the coordinate system is left-handed.

### DIFF
--- a/dune/grid/cpgrid/processEclipseFormat.cpp
+++ b/dune/grid/cpgrid/processEclipseFormat.cpp
@@ -44,12 +44,16 @@
 
 #include <opm/core/grid/cpgpreprocess/preprocess.h>
 #include <opm/core/grid/MinpvProcessor.hpp>
+#include <opm/core/grid/RepairZCORN.hpp>
 #include <opm/core/utility/StopWatch.hpp>
 
 #include <opm/parser/eclipse/Deck/Deck.hpp>
 
+#include <cstddef>
 #include <fstream>
 #include <iostream>
+#include <initializer_list>
+#include <utility>
 
 namespace Dune
 {
@@ -99,11 +103,51 @@ namespace cpgrid
         std::vector<double> coordData;
         ecl_grid.exportCOORD(coordData);
 
-        std::vector<double> zcornData;
-        ecl_grid.exportZCORN(zcornData);
-
         std::vector<int> actnumData;
         ecl_grid.exportACTNUM(actnumData);
+
+        std::vector<double> zcornData;
+        {
+            ecl_grid.exportZCORN(zcornData);
+
+            auto repair = ::Opm::UgGridHelpers::RepairZCORN {
+                std::move(zcornData), actnumData,
+                std::vector<int>{ ecl_grid.getNX() ,
+                                  ecl_grid.getNY() ,
+                                  ecl_grid.getNZ() }
+            };
+
+            zcornData = repair.destructivelyGrabSanitizedValues();
+
+            if (repair.switchedToDepth()) {
+                std::cout << "ZCORN Values Switched from Elevation to "
+                          << "Depth (Sign Reversal)\n";
+            }
+
+            {
+                const auto& statTBB = repair.statTopBelowBottom();
+
+                if (statTBB.cells > std::size_t{0}) {
+                    std::cout << "ZCORN Changes From Top Not Below Bottom:\n"
+                              << "  - Number of Cells Changed:   "
+                              << statTBB.cells << '\n'
+                              << "  - Number of Corners Changed: "
+                              << statTBB.corners << '\n';
+                }
+            }
+
+            {
+                const auto& statBBLT = repair.statBottomBelowLowerTop();
+
+                if (statBBLT.cells > std::size_t{0}) {
+                    std::cout << "ZCORN Changes From Bottom Not Below Lower Top:\n"
+                              << "  - Number of Cells Changed:   "
+                              << statBBLT.cells << '\n'
+                              << "  - Number of Corners Changed: "
+                              << statBBLT.corners << '\n';
+                }
+            }
+        }
 
         // Make input struct for processing code.
         grdecl g;
--- a/dune/grid/cpgrid/processEclipseFormat.cpp
+++ b/dune/grid/cpgrid/processEclipseFormat.cpp
@@ -62,6 +62,10 @@ namespace Dune
     // Forward declarations.
     namespace
     {
+        std::vector<double>
+        getSanitizedZCORN(const ::Opm::EclipseGrid& ecl_grid,
+                          const ::std::vector<int>& actnum);
+
         typedef std::array<int, 3> coord_t;
         typedef std::array<double, 8> cellz_t;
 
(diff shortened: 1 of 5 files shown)


bska:

bska:
Since this PR touches some very core grid facilities, I have tested this locally both on flow simulations and "difficult" upscaling cases.  In those tests none of the ZCORN conditions fire (except for the aforementioned `flaser.grdecl`), but I'd appreciate some actual run-tests by other interested parties before this gets installed in master.

joakim-hove:
FYI interest some desperate hacking in this domain was performed here: https://github.com/OPM/opm-parser/pull/1000 - so for grids which have passed through the `EclipseGrid` object in opm-parser some of this trickery has already been performed. 

It is probably better to do it where you are doing, *might* be possible to remove this hacking in the input layer now?

bska:
> for grids which have passed through the [`EclipseGrid`](https://github.com/OPM/opm-parser/blob/master/opm/parser/eclipse/EclipseState/Grid/EclipseGrid.hpp) object in opm-parser some of this trickery has already been performed.

I do recall that work.  When upscaling the `flaser.grdecl` case that handling did not fire however because the corner-points were essentially fine--just interpreted as elevations (decreasing downwards) rather than depths (increasing downwards).  When combined with the [periodic grid extensions](https://github.com/OPM/opm-grid/blob/master/dune/grid/cpgrid/processEclipseFormat.cpp#L313-L320) and, in particular, the [vertical truncation](https://github.com/OPM/opm-grid/blob/master/dune/grid/cpgrid/processEclipseFormat.cpp#L415-L424) needed to impose periodic flow boundary conditions for upscaling we got a grid with zero active cells.  I didn't want to introduce a lot of special-case handling in a code-path that is already quite complicated so I rather set about solving a more general problem.

The downside to destructively changing ZCORN and guaranteeing that the values are depths is that an upscaling case (e.g., from the SBED tool) may become flipped with respect to its containing geometry.  I don't know if that is a real concern in a practical workflow.

> *might* be possible to remove this hacking in the input layer now?

Yes, I think that's a good idea.

dune/grid/cpgrid/processEclipseFormat.cpp
```diff
@@ -99,12 +103,52 @@ namespace cpgrid
...
         ecl_grid.exportCOORD(coordData);
 
-        std::vector<double> zcornData;
-        ecl_grid.exportZCORN(zcornData);
-
         std::vector<int> actnumData;
         ecl_grid.exportACTNUM(actnumData);
 
+        std::vector<double> zcornData;
+        {
```
atgeirr: It would be good to have this section as a separate function returning (possibly sanitized) zcornData, to keep the length of this function more managable.

dune/grid/cpgrid/processEclipseFormat.cpp
```diff
@@ -99,12 +103,52 @@ namespace cpgrid
...
+                std::move(zcornData), actnumData,
+                std::vector<int>{ ecl_grid.getNX() ,
+                                  ecl_grid.getNY() ,
+                                  ecl_grid.getNZ() }
+            };
+
+            zcornData = repair.destructivelyGrabSanitizedValues();
+
+            if (repair.switchedToDepth()) {
+                std::cout << "ZCORN Values Switched from Elevation to "
```
atgeirr: These should eventually be replaced with OpmLog::info() or OpmLog::warning(), but then the upscaling programs must also be changed to set up logging (preferably just using the OpmLog::setupSimpleDefaultLogging() function) so that can wait. We do risk getting messy output from the simulators though, if triggered.

opm/core/grid/RepairZCORN.hpp
```diff
@@ -0,0 +1,728 @@
...
+
+        /// Check whether or not all elements of a vector are equal, while
+        /// ignoring particular element value.
+        ///
+        /// \param[in] coll Sequence of elements.
+        ///
+        /// \param[in] ignore Special value that should be considered equal
+        ///    (i.e., ignored) for comparison purposes.  Typically zero.
+        ///
+        /// \param[in] lookfor Search hint for which value to look for in
```
atgeirr: I think this doc is a bit misleading. If calling allEqual(v, ignore, x) I am asking if all non-ignored elements of v are equal to x, not if they are all equal to each other. I.e. allEqual({1, 1, 1}, 0, 2) will return false, since not all are equal to 2. This is not just an optimization or a hint, so the doc must change.

opm/core/grid/RepairZCORN.hpp
```diff
@@ -0,0 +1,728 @@
...
+            for (const auto& globCell : this->active_.activeGlobal()) {
+                all_signs.push_back(this->getZCornSign(globCell));
+            }
+
+            const int ignore = 0;
+
+            // Elevation == ZCORN decreasing => all non-ignored cell signs
+            // negative.
+            return (first(all_signs, ignore) == -1)
+                && allEqual(all_signs, ignore, -1);
```
atgeirr: Cannot this return statement be replaced with just `return allEqual(all_signs, ignore, -1);`?

Also: what should happen if all_signs contains a mix of signs? It seems to me that we will then assume zcorn is depth.

opm/core/grid/RepairZCORN.hpp
```diff
@@ -0,0 +1,728 @@
...
+            ///
+            /// Assumes that lateral cell positions are valid.
+            ///
+            /// \param[in] ijk Cartesian index position of single cell.
+            ///
+            /// \return Global ID of cell immediately below \p ijk.  If no
+            ///    such cell exists, e.g., because the input cell is in the
+            ///    bottom layer of the model or all cells below it are
+            ///    explicitly deactivated, then function neighbourBelow()
+            ///    returns \code static_cast<std::size_t>(-1) \endcode.
```
atgeirr: I'm not too happy to see size_t used here. Rather use a signed number, assign -1 using a named constant and check against that constant in the calling code.

The same goes for the getGlobalIndex() method.

opm/core/grid/RepairZCORN.hpp
bska: > Cannot this return statement be replaced with just `return allEqual(all_signs, ignore, -1);`?

Yes, probably.  I'll think about it.

opm/core/grid/RepairZCORN.hpp
bska: > If calling `allEqual(v, ignore, x)` I am asking if all non-ignored elements of `v` are equal to `x`, not if they are all equal to each other.

Sure.  However, we're only using the three-parameter version of `allEqual()` if we already **know** what `first(v, ignore)` returns.  In the context of that use, we're omitting the first scan that would otherwise be needed to determine which value to look for.

dune/grid/cpgrid/processEclipseFormat.cpp
bska: > It would be good to have this section as a separate function returning (possibly sanitized) `zcornData`,

I don't disagree.  Having to pass **both** the `ecl_grid` and the `actnumData` is a bit annoying though.  Maybe we could add a function
```C++
struct CpGridInput {
    std::vector<int>    actnum;
    std::vector<double> coord;
    std::vector<double> zcorn;
};

CpGridInput getCpGridKeywords(const EclipseGrid& G);
```
to `processEclipseFormat.cpp` but that is essentially just moving the problem around.

dune/grid/cpgrid/processEclipseFormat.cpp
bska: > These should eventually be replaced with OpmLog::*()

Fine.  **After** the default of that family is to **not** print those ASCII escape sequences.

dune/grid/cpgrid/processEclipseFormat.cpp
atgeirr: I'd say just go with the simple solution, even though the conceptually overlapping arguments is a little inelegant.

dune/grid/cpgrid/processEclipseFormat.cpp
atgeirr: > Fine. After the default of that family is to not print those ASCII escape sequences.

I'll add another argument to the setupSimpleDefaultLogging() function for that. I'd like to avoid default arguments though (leaving that for the callers for now).

opm/core/grid/RepairZCORN.hpp
atgeirr: Can the check `i < numGlobalCells()` become false for any *other* reason than the above? If not I think it should be signed.

opm/core/grid/RepairZCORN.hpp
atgeirr: That is a detail of the current usage of the function, which could easily be changed in the future by someone not knowing this, so I still think the doc must change to reflect function behaviour.

bska:
Don't merge yet.  Commit 9fe9fba breaks the tests.

bska:

atgeirr:

atgeirr:
I cannot reproduce the Jenkins failure locally. @akva2 could this be related to the opm-material-related trouble, indicating that Jenkins does not clean up properly? I assumed that restarting the Jenkins process would use the current master from all repositories, except for the one being checked (here opm-grid)?

atgeirr:

atgeirr:
I'll just have some sanity-checks run on this, but will merge when those are complete.

bska:
> I'll just have some sanity-checks run on this, but will merge when those are complete.

Okay, thanks a lot!

atgeirr:
Confirmed no change for model 2 and Norne, merging.

PR #268, opened 2017-03-21
Travis: apt-get install bc

joakim-hove:


bska:

PR #271, opened 2017-04-20
Cleanup DUNE_VERSION_NEWER( 2 , 3 )

dr-robertk:
DUNE 2.3 is a hard dependency so all checks of the form DUNE_VERSION_NEWER( ..., 2 ,3 ) can be safely removed. To also remove some warning the use of Dune::tuple, which is a forward to std::tuple was removed and std::tuple is used consequently. 

This should not be backported to the current release.

### DIFF
--- a/dune/grid/CpGrid.hpp
+++ b/dune/grid/CpGrid.hpp
@@ -49,7 +49,7 @@
 
 #include <dune/common/version.hh>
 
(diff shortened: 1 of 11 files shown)


atgeirr:

atgeirr:

atgeirr:
Checks succesful, merging!

PR #273, opened 2017-04-28
Make dgfparser.hh compile with Clang.

atgeirr:
This was broken by #262.

Value returned is dummy anyway, should at least match expected type. It seemed to me that the defaultValue() method would be the simplest way to ensure this.

I intend to self-merge this after Jenkins, as it is compile-critical.


atgeirr:

PR #274, opened 2017-05-07
Use libecl in jenkins.

joakim-hove:


atgeirr:
For some reason this seems to have been fixed by pushing a commit directly rather than merging this. In any case this PR is obsolete, closing.

PR #278, opened 2017-07-19
require dune >= 2.4

akva2:
Required for correct results in opm-simulators.

### DIFF
--- a/dune.module
+++ b/dune.module
@@ -10,5 +10,5 @@ Label: 2017.10-pre
 Maintainer: atgeirr@sintef.no
 MaintainerName: Atgeirr F. Rasmussen
 Url: http://opm-project.org
-Depends: dune-common (>= 2.3) dune-geometry (>= 2.3) dune-grid (>= 2.3)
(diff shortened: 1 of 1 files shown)


andlaus:
can this (or #277) be merged soonish, please? there are several technical advantages of dune 2.4 over dune 2.3, jenkins and travis are both on 2.4 already (Jenkins even fails when using Dune 2.3) and most of the current flock of Linux distributions ship with at least Dune 2.4, i.e., you actually have to do some work to shoot yourself into your foot by using Dune 2.3.

tl;dr: I can see no real reasons to continue to support Dune 2.3, but I see quite a few technical reasons to drop it.

alfbr:
You are breaking the benchmarks and general building of opm on Debian stable and old-stable with this patch. As I have already said, we are working hard to fix it, one vacation day and counting. Please allow us the time to fix this in the right order.

akva2:
@alfbr is it okay to merge this with friends now ?

alfbr:
Yes, it should be safe to merge. Thanks for the patience.

PR #279, opened 2017-08-28
Allow compilation of parallel tests with DUNE 2.3.

blattms:
In that version checkpartition.cc uses dune_static_assert but
never includes the header dune/common/static_assert.hh that defines
it. Therefore we now do this ourselves for DUNE 2.3.

I guess some previous commit must have stripped an include of header that
implicitly included static_assert.hh. Otherwise compilation would have failed.
(Or nobody using 2.3 compiled the tests.)

Unless #278 is merged this is needed on my system. Note that a merge of this requires a change in #277.

### DIFF
--- a/tests/cpgrid/distribution_test.cpp
+++ b/tests/cpgrid/distribution_test.cpp
@@ -25,6 +25,9 @@
 #include <dune/grid/test/checkpartition.hh>
 #include <dune/grid/test/checkcommunicate.hh>
 #else
+// checkpartition.cc in dune-grid uses dune_static_assert but never includes
+// the header in version 2.3.1
+#include <dune/common/static_assert.hh>
 #include <dune/grid/test/checkpartition.cc>
 #include <dune/grid/test/checkcommunicate.cc>
 #endif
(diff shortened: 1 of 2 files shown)


blattms:
Mmh, the travis build fails due a cmake macro search path mixup:
```
CMake Error at /home/travis/build/OPM/dune-common/cmake/modules/FindUMFPack.cmake:21 (find_package):

  find_package called with invalid argument "OPTIONAL_COMPONENTS"

Call Stack (most recent call first):

  cmake/modules/DuneIstlMacros.cmake:8 (find_package)

  /home/travis/build/OPM/dune-common/cmake/modules/DuneMacros.cmake:521 (include)

  /home/travis/build/OPM/dune-common/cmake/modules/DuneMacros.cmake:664 (dune_process_dependency_macros)

  CMakeLists.txt:24 (dune_project)
```
There is a FindUMFPack.cmake in both DUNE and OPM. Unfortunately, when configuring DUNE the one of OPM is found and used and they seem to be incompatible

blattms:
jenkins build this please

blattms:
Well, it turns out that the problem was entirely caused by my messed up system. I installed  dune-grid 2.3.1 together with dune-common 2.4.1.

PR #280, opened 2017-09-18
[cleanup] remove_const --> std::remove_const.

dr-robertk:
Use remove_const from std namespace.


akva2:
@atgeirr would be nice to get this in as it clogs up logs on refreshed jenkins.

atgeirr:
Sure, going in!

PR #283, opened 2017-10-02
Make DUNE 2.5 grid check pass

blattms:
DUNE overdoes it a bit with the generic tests and tests jacobians, local, and global coordinates even for vertices. In additions it again neglects that some grids might not have codim 1 entities and just asks them for corners.

This PR makes the grid check run through again

### DIFF
--- a/dune/grid/cpgrid/Entity.hpp
+++ b/dune/grid/cpgrid/Entity.hpp
@@ -394,6 +394,21 @@ namespace Dune
 namespace Dune {
 namespace cpgrid {
 
+namespace Detail
+{
+inline unsigned int numFaces(const OrientedEntityTable<0, 1>& cell_to_face,
+                             const Entity<0>& e)
+{
+    return  cell_to_face[e].size();
+}
+
+template<int codim>
+unsigned int numFaces(const OrientedEntityTable<0, 1>&, const Entity<codim>&)
+{
+    return 0;
+}
+} // end namespace Detail
+
 template<int codim>
 unsigned int Entity<codim>::subEntities ( const unsigned int cc ) const
 {
@@ -402,7 +417,7 @@ unsigned int Entity<codim>::subEntities ( const unsigned int cc ) const
         return 1;
     } else if ( codim == 0 ){
         if ( cc == 1 ) {
-            return pgrid_->cell_to_face_[*this].size();
+            return Detail::numFaces(pgrid_->cell_to_face_, *this);
         } else if ( cc == 3 ) {
             return 8;
         }
--- a/dune/grid/cpgrid/Entity.hpp
+++ b/dune/grid/cpgrid/Entity.hpp
@@ -180,24 +180,7 @@ namespace Dune
             }
 
             /// The count of subentities of codimension cc
-            unsigned int subEntities ( const unsigned int cc ) const
-            {
-                // static_assert(codim == 0, "");
-                if (cc == 0) {
-                    return 1;
-                } else if (cc == 3) {
-                    return 8;
-                } else {
-                    return 0;
-                }
-//              if (cc == 0) {
-//                  return 1;
-//              } else if (cc == 1) {
-//                  return pgrid_->cell_to_face_[*this].size();
-//              } else {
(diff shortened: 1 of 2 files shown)


blattms:
jenkins build this please.

dune/grid/cpgrid/Entity.hpp
```diff
@@ -411,6 +394,37 @@ namespace Dune
...
+
+template<int codim>
+unsigned int Entity<codim>::subEntities ( const unsigned int cc ) const
+{
+    // static_assert(codim == 0, "");
+    if (cc == 0) {
+        return 1;
+    } else if ( codim == 0 ){
+        if ( cc == 1 ) {
+            return Detail::numFaces(pgrid_->cell_to_face_, *this);
```
atgeirr: Are you sure about this one? I would think that this promises that we have numFaces codim 1 sub-entities, whereas we actually have none (we have only intersections, not codim 1 entities).

dune/grid/cpgrid/Geometry.hpp
```diff
@@ -398,8 +398,10 @@ namespace Dune
             /// This method is meaningless for singular geometries.
             GlobalCoordinate corner(int /* cor */) const
             {
-                OPM_THROW(std::runtime_error, "Meaningless call to cpgrid::Geometry::corner(int): "
-                      "singular geometry has no corners.");
+                // Meaningless call to cpgrid::Geometry::corner(int): 
+                //"singular geometry has no corners.
+                // But the DUNE tests assume at least one corner.
+                return GlobalCoordinate( 0.0 );
```
atgeirr: In general I am not very happy about meaningless calls being allowed, so I would prefer that the tests were rewritten. I do realize we cannot do that, in particular for existing stable dune versions, so I'll accept this. However, could we at least return points that are likely to trigger other problems, such as `return GlobalCoordinate(-1e100);` or similar?

atgeirr:
I have tested this locally with Dune 2.5, and this causes another test to fail (the cpgrid geometry test) because it expects throws... The following change fixes that:
```
diff --git a/tests/cpgrid/geometry_test.cpp b/tests/cpgrid/geometry_test.cpp
index 8e51628..ad4847e 100644
--- a/tests/cpgrid/geometry_test.cpp
+++ b/tests/cpgrid/geometry_test.cpp
@@ -50,12 +50,12 @@ BOOST_AUTO_TEST_CASE(vertexgeom)
     BOOST_CHECK_EQUAL(g.corner(0), c);
     Geometry::LocalCoordinate lc(0.0);
     BOOST_CHECK_EQUAL(g.global(lc), c);
-    BOOST_CHECK_THROW(g.local(c), std::exception);
+    // BOOST_CHECK_THROW(g.local(c), std::exception);
     BOOST_CHECK_EQUAL(g.integrationElement(lc), 1.0);
     BOOST_CHECK_EQUAL(g.volume(), 1.0);
     BOOST_CHECK_EQUAL(g.center(), c);
-    BOOST_CHECK_THROW(g.jacobianTransposed(lc), std::exception);
-    BOOST_CHECK_THROW(g.jacobianInverseTransposed(lc), std::exception);
+    // BOOST_CHECK_THROW(g.jacobianTransposed(lc), std::exception);
+    // BOOST_CHECK_THROW(g.jacobianInverseTransposed(lc), std::exception);
 }
 
 
@@ -74,7 +74,7 @@ BOOST_AUTO_TEST_CASE(intersectiongeom)
     BOOST_CHECK(g.type().isNone());
     BOOST_CHECK(g.affine());
     BOOST_CHECK_EQUAL(g.corners(), 0);
-    BOOST_CHECK_THROW(g.corner(0), std::exception);
+    // BOOST_CHECK_THROW(g.corner(0), std::exception);
     Geometry::LocalCoordinate lc(0.0);
     BOOST_CHECK_THROW(g.global(lc), std::exception);
     BOOST_CHECK_THROW(g.local(c), std::exception);
```

atgeirr:
This was merged via #284. My question about 1-entities still stands though!

PR #285, opened 2017-10-05
Use wells

joakim-hove:
With this PR the load-balance & scatter operations on the grid take `std::vector< const Well*>` argument instead of an `EclipseState` argument. The direct motivation is downstream of: https://github.com/OPM/opm-parser/pull/1138

### DIFF
--- a/dune/grid/common/WellConnections.cpp
+++ b/dune/grid/common/WellConnections.cpp
@@ -31,27 +31,25 @@ namespace Dune
 {
 namespace cpgrid
 {
-WellConnections::WellConnections(const OpmEclipseStateType& eclipseState,
+WellConnections::WellConnections(const std::vector<const OpmWellType*>& wells,
                                  const std::array<int, 3>& cartesianSize,
                                  const std::vector<int>& cartesian_to_compressed)
 {
-    init(eclipseState, cartesianSize, cartesian_to_compressed);
+    init(wells, cartesianSize, cartesian_to_compressed);
 }
 
-void WellConnections::init(const OpmEclipseStateType& eclipseState,
+void WellConnections::init(const std::vector<const OpmWellType*>& wells,
                            const std::array<int, 3>& cartesianSize,
                            const std::vector<int>& cartesian_to_compressed)
 {
 #if HAVE_OPM_PARSER
-    std::vector<const Opm::Well*>  wells  = eclipseState.getSchedule().getWells();
-    int last_time_step = eclipseState.getSchedule().getTimeMap().size()-1;
     well_indices_.resize(wells.size());
 
     // We assume that we know all the wells.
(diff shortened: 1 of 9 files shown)


GitPaean:
You mean `std::vector<Well *>` ? :D

joakim-hove:
> You mean std::vector<Well *> ? :D

Dang - all these wells; yes I mean: `std::vector<const Opm::Well *>`

joakim-hove:
jenkins build this with downstreams opm-parser=1138 opm-output=207 please

joakim-hove:
jenkins build this with downstreams opm-parser=1138 opm-output=207 please

atgeirr:
The Jenkins test failed:

https://ci.opm-project.org/job/opm-grid-PR-builder/lastFailedBuild/

I have no idea why it does not show up on this page (for me at least), nor have I investigated the failure beyond noticing some zoltan-related trouble?!?

dune/grid/common/WellConnections.cpp
```diff
@@ -31,27 +31,25 @@ namespace Dune
...
-    std::vector<const Opm::Well*>  wells  = eclipseState.getSchedule().getWells();
-    int last_time_step = eclipseState.getSchedule().getTimeMap().size()-1;
     well_indices_.resize(wells.size());
 
     // We assume that we know all the wells.
     int index=0;
     for (const auto well : wells) {
         std::set<int>& well_indices = well_indices_[index];
-        const auto& completionSet = well->getCompletions(last_time_step);
+        const auto& completionSet = well->getCompletions( );
```
atgeirr: Is this equivalent? I assume that no arguments would give all completions over the whole history of the well, and last_time_step only those present at the end? If so, then this would actually be a bugfix, since if a well loses some completions before the end they should still be considered for the partitioning!

joakim-hove: They are 100% equivalent. Completions can not disappear.

atgeirr: Thanks for the clarification! I guess I was a little confused between completions appearing/disappearing, and being opened/closed. I'll restart the Jenkins test just to see if there was a temporary hiccup.

atgeirr:
jenkins build this with downstreams opm-parser=1138 opm-output=207 please

atgeirr:
This fails because of (among other things):
```
/var/lib/jenkins/workspace/opm-grid-PR-builder/deps/opm-core/opm/core/wells/WellsManager_impl.hpp:349:22: error: 'const class Opm::EclipseState' has no member named 'getSchedule'
```
If this function has been remove an opm-core PR also needs to be in the mix.

joakim-hove:
I know - PR in opm-core++ coming.

PR #289, opened 2017-11-13
changed: make libdune-uggrid-dev an alternative for libug-dev

akva2:
former is used on stretch, latter is used on xenial


atgeirr:
Since opm-grid does not depend on UG, why is it included here?

akva2:
i see libdune-grid-dev specifies it as a requirement now, at some point it didn't. so we indeed can remove these explicits now.

atgeirr:
OK, merge at will.

PR #291, opened 2017-12-06
[feature][PolyhedralGrid] Allow to use coordinate type other than double.

dr-robertk:
This PR adds a template parameter to PolyhedralGrid allowing to set the coordinate type. The default is as before double.

### DIFF
--- a/dune/grid/polyhedralgrid/cartesianindexmapper.hh
+++ b/dune/grid/polyhedralgrid/cartesianindexmapper.hh
@@ -6,10 +6,10 @@
 
 namespace Dune
 {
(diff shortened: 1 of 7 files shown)


andlaus:
That's a good feature to have IMO. One question, though: what does this change? I suppose that the `UnstructuredGrid` structure stays as it is because of its C-nature?

atgeirr:

dr-robertk:
Yes, UnstructuredGrid stays the same for now, but at least the upper layer  has the correct structure.

atgeirr:
Thanks, merging!

PR #298, opened 2018-01-08
replace Dune::array by std::array

andlaus:
Dune::array is removed in Dune 2.6. This is easy to fix because Dune::array can be swapped out for std::array.


atgeirr:

atgeirr:
All green, merging. Thanks for the patch!

PR #302, opened 2018-01-22
Add std::string specialization for read and write in p2pcommunicator

totto82:


atgeirr:
I think this looks ok, but as always one has to be careful when mixing normal overloading and template functions. Consider the following example program:
```
#include <iostream>

template <typename T>
void foo(T& t)
{
    std::cout << "I am in generic foo(T&), and t has size " << sizeof(t) << std::endl;
}

template <typename T>
void foo(const T& t)
{
    std::cout << "I am in generic foo(const T&), and t has size " << sizeof(t) << std::endl;
}

void foo(const std::string& s)
{
    std::cout << "I am in string foo(), and s has length " << s.length() << std::endl;    
}

int main()
{
    // Temporaries.
    foo(0.0);
    foo(0);
    foo(0L);
    foo(nullptr);
    foo(std::string("12345"));

    // Variables.
    auto d = 0.0;
    auto i = 0;
    auto l = 0L;
    auto p = nullptr;
    auto s = std::string("12345");
    foo(d);
    foo(i);
    foo(l);
    foo(p);
    foo(s);    
}
```

The output of the above program is:
```
I am in generic foo(const T&), and t has size 8
I am in generic foo(const T&), and t has size 4
I am in generic foo(const T&), and t has size 8
I am in generic foo(const T&), and t has size 8
I am in string foo(), and s has length 5
I am in generic foo(T&), and t has size 8
I am in generic foo(T&), and t has size 4
I am in generic foo(T&), and t has size 8
I am in generic foo(T&), and t has size 8
I am in generic foo(T&), and t has size 24
```
the surprise here is that when we have an lvalue (here: a variable we are allowed to change) the generic foo(T&) is called, even though the call to foo(const string&) would be legal.

I think that the current code is ok, since the templated write() also takes a const reference, but I would appreciate it if someone else also could verify this.

atgeirr:

atgeirr:
Tests green and no comments, I'll merge this independently.

PR #303, opened 2018-01-22
Quell warnings when building with dune 2.6

akva2:
This suppresses warnings when building with dune 2.6.

The last commit is a build fix (only triggered from simulators); it can possibly be made unconditional, i wasn't immediately sure about the dune 2.4 status.

### DIFF
--- a/dune/grid/cpgrid/Entity.hpp
+++ b/dune/grid/cpgrid/Entity.hpp
@@ -36,6 +36,7 @@
 #ifndef OPM_ENTITY_HEADER
 #define OPM_ENTITY_HEADER
 
+#include <dune/common/version.hh>
 #include <dune/geometry/type.hh>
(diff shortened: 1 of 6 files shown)


andlaus:
looks good as far as I can see.

I'm pretty sure that the `Dune::array` usage can be replaced by `std::array` unconditionally even in Dune 2.4 and I'd strongly recommend doing so. (jenkins can test this on dune 2.4, right?)

also, while doing the first pass on dune 2.6 errors in #298, I disabled the `grid_test` unit test because it made the compiler trip and I did not know what to make of this error message while downstream works fine. maybe you can have a look?

akva2:
will check if i can do the array unconditionally.

cpgrid needs some refactoring to conform to new dune grid expectations, which is the reason grid_test.cc breaks. the changes are non-trivial, the geometry and intersection stuff needs a do-over.

PR #304, opened 2018-01-24
fixed/changed: make opm-parser a optional dependency

akva2:
It's a bit unclear what was intended, but I think the optionality is wanted. I at the very least have to fix the config vars for the opm-parser buildsystem change.


atgeirr:

PR #308, opened 2018-02-08
adding functions to access and recover the current view of CpGrid

GitPaean:
currentView() and setView()

The motivation please see OPM/ewoms#283.  I do see this expose `current_view_data_` to outside. Please let me know if it is something welcome. If not, I will be satisfied in the current implementation in OPM/ewoms#283.

### DIFF
--- a/dune/grid/CpGrid.hpp
+++ b/dune/grid/CpGrid.hpp
@@ -1238,6 +1238,18 @@ namespace Dune
         {
             current_view_data_=distributed_data_.get();
         }
+
+        /// \brief return the curent view
+        cpgrid::CpGridData* const currentView() const
+        {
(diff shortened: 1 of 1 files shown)


GitPaean:
Or we move the functionality `Opm::UgGridHelpers::createEclipseGrid()` as a member function of `CpGrid`? Any other suggestion?

dr-robertk:
@GitPaean : Exposing CpGrid internal data essentially breaks the interface, as for example,
global_cell did. Maybe you could take a look at CartesianIndexMapper and the specializations see if that approach would also work in your case.

GitPaean:
Closing this PR as other approaches suggested in OPM/ewoms#283

PR #312, opened 2018-03-03
partition_iterator_test: use a typedef to define for the iterator variables

andlaus:
ICC does not like the previous code. this is either a (small) bug in ICC or gcc/clang, or a ambiguity in the c++-2011 standard. anyway, the fix is pretty trivial.


atgeirr:
> the fix is pretty trivial.

... agreed. Also, it is a little more readable (would be even more with using instead of typedef though...)

atgeirr:
jenkins build this please

PR #314, opened 2018-03-07
fix the build without opm-common being available

andlaus:
the dune.module file states that opm-common is optional, so opm-grid is supposed to be buildable without it. since opm-grid without the ECL I/O routines has only limited utility, opm-common could also be made mandatory instead.

(note that with the default build system it is not possible to pull the feat of compiling opm-grid without opm-common because the cmake files are located there. the alternative build system works fine, though.)

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -53,12 +53,16 @@ list (APPEND MAIN_SOURCE_FILES
   opm/grid/cpgpreprocess/uniquepoints.c
   opm/grid/UnstructuredGrid.c
   opm/grid/grid_equal.cpp
-  opm/grid/transmissibility/trans_tpfa.c
   opm/grid/utility/compressedToCartesian.cpp
-  opm/grid/utility/VelocityInterpolation.cpp
   opm/grid/utility/StopWatch.cpp
   opm/grid/utility/WachspressCoord.cpp
   )
+
+if (opm-common_FOUND)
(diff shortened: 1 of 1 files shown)


andlaus:
jenkins build this please

andlaus:
jenkins build this please

CMakeLists_files.cmake
```diff
@@ -93,15 +97,16 @@ list (APPEND TEST_SOURCE_FILES
...
-  tests/test_ug.cpp
   tests/test_quadratures.cpp
 	)
 
-if(opm-parser_FOUND)
-  list(APPEND TEST_SOURCE_FILES tests/test_compressedpropertyaccess.cpp)
+if(HAVE_ECL_INPUT)
+  list(APPEND TEST_SOURCE_FILES
+		tests/test_regionmapping.cpp
+    tests/test_ug.cpp
```
akva2: meh.

andlaus: I agree. That said, it seems like these tests use stuff from opm-parser and this is already used in this file..

akva2: i was only referring to the indent.

andlaus: ah, I see. the reason is that the modeline at the beginning of the file instructs emacs to use tabs instead of spaces and the tab with should be assumed to be 2 while github uses a tabwidth of 8. (i.e., in my emacs the indentation is fine).

PR #315, opened 2018-04-03
setup-opm-data -> setup-opm-tests

akva2:


atgeirr:
jenkins build this please

atgeirr:
The test failure was a timeout on an ewoms test, we seem to be cutting it close with some of those. I'll rerun and it will probably succeed with no other Jenkins load.

atgeirr:
jenkins build this please

PR #318, opened 2018-04-26
Construct VariableSizeCommunicator with a sufficient buffer for sending at once

### LINKED ISSUE #316: opm-grid cannot partition the SPE-10 grid
when trying to run the SPE10_MODEL2 deck that is shipped with opm-data on four cores and with ZOLTAN and PTScotch installed, I get this:

```
> mpirun -np 4 ./bin/flow output_dir=. SPE10_MODEL2.DATA

**********************************************************************
*                                                                    *
*                      This is flow 2018.04-pre                      *
*                                                                    *
* Flow is a simulator for fully implicit three-phase black-oil flow, *
*             including solvent and polymer capabilities.            *
*          For more information, see http://opm-project.org          *
*                                                                    *
**********************************************************************

After loadbalancing process 0 has 322630 cells.
```

I've waited for several hours but nothing further happened while the four cores stayed busy. I used the latest dune 2.5 git branch and an optimized build on ubuntu 16.04, i.e. GCC 5.4.0 in conjunction with `-O3 -march=native -DNDEBUG` on a Haswell CPU.

blattms:
Otherwise the message will be chopped into several sends and for DUNE version
less than 2.5.2 the code will deadlock because of a bug. We assume that we will
never send more than 16 values per cell and use it to calculate an upper bound
for the message size. This is send passed to the constructor of
VariableSizeCommunicator to ensure that the message is never chopped into
multiple ones.

Fixes #316. 

The number 16 is up for debate. 4 should probably suffice too (for polymer) or can there be more?

Please test.

### DIFF
--- a/opm/grid/cpgrid/CpGridData.hpp
+++ b/opm/grid/cpgrid/CpGridData.hpp
@@ -580,8 +580,8 @@ void CpGridData::communicateCodim(DataHandle& data, CommunicationDirection dir,
     {
         OPM_THROW(std::logic_error,
                   "You triggered a bug in the communication of DUNE"
-                   << " as yu tried to send more than " << MAX_DATA_PER_CELL
-                   << " values per cell/point. PLease define MAX_DATA_COMMUNICATED_PER_ENTITY"
+                   << " as you tried to send more than " << MAX_DATA_PER_CELL
+                   << " values per cell/point. Please define MAX_DATA_COMMUNICATED_PER_ENTITY"
                    << " to a high enough value when compiling.");
     }
 #endif
--- a/opm/grid/cpgrid/CpGridData.hpp
+++ b/opm/grid/cpgrid/CpGridData.hpp
@@ -517,18 +517,74 @@ void CpGridData::communicateCodim(DataHandle& data, CommunicationDirection dir,
     }
     Entity2IndexDataHandle<DataHandle, codim> data_wrapper(*this, data);
     std::size_t max_interface_entries = 0;
+
+#if DUNE_VERSION_NEWER_REV(DUNE_GRID, 2, 5, 2)
+    VariableSizeCommunicator<> comm(ccobj_, interface);
+#else
     // Work around a bug/deadlock in DUNE <=2.5.1 which happens if the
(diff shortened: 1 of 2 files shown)


atgeirr:
Question: if the max limit chosen here (16 values) *is* insufficient, will the code silently fail or stop?

atgeirr:
jenkins build this please

blattms:
@atgeirr it will silently deadlock, but after loadbalancing. I will try to add an additional check

opm/grid/cpgrid/CpGridData.hpp
```diff
@@ -517,18 +517,74 @@ void CpGridData::communicateCodim(DataHandle& data, CommunicationDirection dir,
...
+
+#ifndef NDEBUG
+    max_items = ccobj_.max(max_items);
+
+    if ( max_items > MAX_DATA_PER_CELL)
+    {
+        OPM_THROW(std::logic_error,
+                  "You triggered a bug in the communication of DUNE"
+                   << " as yu tried to send more than " << MAX_DATA_PER_CELL
+                   << " values per cell/point. PLease define MAX_DATA_COMMUNICATED_PER_ENTITY"
```
akva2: lower case l in please

blattms:
Thanks. Fixed this and another one in cea3c15

akva2:
jenkins build this please

andlaus:
I can confirm that this PR fixes parallel runs of SPE10 on my machine. That said, if this only happens on older dune versions and only for grids with a large number of active cells, it may be better from a maintainance perspective of not merging this work-around and just tell people to use a non-ancient dune version for parallel runs of large decks?

akva2:
i can't imagine why we would ever want these messages chunked so imo it improves the situation everywhere.

andlaus:
agreed. the point is that this PR adds 105 lines of non-trivial code that just works around a bug in old dune versions and which we must maintain. Possibly, that burden is not worth the benefit.

atgeirr:
*Not* including this requires quite new Dune versions, we should be robust with respect to that; the small matrix inversion is another typical example. I'll therefore merge it. Thanks for the patch, and testing!

PR #322, opened 2018-05-09
Stop wasting space when working around VariableSizeCommunicator bug.

blattms:
In OPM we always use one vector block with all unknowns and send this. Ergo we only send 1 datum and not more (previous threshold was 16). The only place where we send more is during grid distribution. But for that case we already compute the correct buffer size and this change has no effect.

This is a follow up to #318. Tested with SPE 10 Model 2. #316 is still fixed with this.

### DIFF
--- a/opm/grid/cpgrid/CpGridData.hpp
+++ b/opm/grid/cpgrid/CpGridData.hpp
@@ -116,8 +116,9 @@ public:
         /// \brief The maximum data items allowed per cell (DUNE < 2.5.2)
         ///
         /// Due to a bug in DUNE < 2.5.2 we need to limit this when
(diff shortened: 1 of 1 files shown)


blattms:
any vetos against merging?

blattms:
jenkins build this with norne please.

atgeirr:
> any vetos against merging?

No. Sorry for ignoring! The syntax for jenkins is not 100% so I'll redo it... See https://wiki.opm-project.org/index.php?title=Jenkins_triggers

atgeirr:
jenkins run norne please

blattms:
Thanks.

PR #332, opened 2018-06-19
Fix parallel unique boundary ids

### LINKED ISSUE #329: computeUniqueBoundaryIds not called in/after loadBalance
This prevents using `boundarySegmentIndex`.

blattms:
these were not correctly scattered when loadbalancing the grid. Actually we did set them up but in a temporary vector that was eventually thrown away. Now we store it as we are supposed to do. Added a test
for the boundary ids, too.

Closes #329

### DIFF
--- a/tests/cpgrid/distribution_test.cpp
+++ b/tests/cpgrid/distribution_test.cpp
@@ -485,6 +485,9 @@ init_unit_test_func()
 int main(int argc, char** argv)
 {
     Dune::MPIHelper::instance(argc, argv);
+    MPI_Errhandler errhandler;
+    MPI_Comm_create_errhandler(MPI_err_handler, &errhandler);
+    MPI_Comm_set_errhandler(MPI_COMM_WORLD, errhandler);
     boost::unit_test::unit_test_main(&init_unit_test_func,
                                      argc, argv);
 }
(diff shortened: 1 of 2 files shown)


blattms:
jenkins build this please

alfbr:
Any chance this will effect convergence/performance on MPI runs?

bska:
Having unique boundary IDs is an essential prerequisite to using periodic boundary conditions for relative permeability upscaling in module opm-upscaling.

alfbr:
Ah, right, thanks for the clarification!

atgeirr:
Of course, the upscaling codes do not use a distributed grid... Which is why this was not discovered before. I'll merge!

tests/cpgrid/distribution_test.cpp
```diff
@@ -485,6 +485,9 @@ init_unit_test_func()
 int main(int argc, char** argv)
 {
     Dune::MPIHelper::instance(argc, argv);
+    MPI_Errhandler errhandler;
```
akva2: missing HAVE_MPI protection is seems, see https://ci.opm-project.org/job/opm-grid/103/

blattms:
Thanks a lot for noticing and sorry for carelessness. Posted #337 to fix this.

PR #333, opened 2018-06-19
Use 2 for MAX_DATA_PER_CELL to run grid checks.

blattms:
While one is enough for OPM these checks need at least 2. Otherwise our tests, compiled with with -DDEBUG, will throw an exception when run.


blattms:
jenkins build this please

PR #336, opened 2018-06-25
Restore build of 'distribution_test' for non-MPI configurations

bska:
See @akva2's [commen](https://github.com/OPM/opm-grid/pull/332#pullrequestreview-131543620) in PR #332.

PR #337, opened 2018-06-25
Add missing HAVE_MPI to ditribution_test.

blattms:
Otherwise this cannot compile without MPI support.

This bug was introduced by #332. Sorry for the inconvenience.


blattms:
To slow. Closing in favor of #336

PR #338, opened 2018-06-27
Refactor/cleanup DefaultGeometryPolicy::geomVector.

blattms:
It used to use boost::mpl::if_c and various helper structs to
get the geometry for different geometries. This seemed very complicated.

This commit removes the dependeny on boost::mpl and uses rather simple
function overloading by transforming the template parameter codim into
a different type using integral_constant.


blattms:
jenkins build this please

atgeirr:
I tried to recall why it was done like this, but failed... I think perhaps you also changed it a bit when implementing the load balancing? In any case, this is a nice simplification, and Jenkins is green. Merging!

blattms:
I am pretty sure that that code originated from me when parallelizing CpGrid.

PR #339, opened 2018-07-12
Prevent unused variable warning for Dune > 2.5.2

blattms:
In that case there is no need to calculate the maximum entries
of an interface. Therefore we move the declaration of variable
max_interface_entries into the compile time if branch where it
is actually used.


blattms:
jenkins build this please

PR #340, opened 2018-07-30
[WIP] Reoder cells by ijk for better linear solver convergence.

blattms:
As a result cells along a column have subsequent local cell indices. Supposably this is better for the convergence of the linear solvers. Indeed we see a 20% decrease in computation time for Norne using the
default solver. For model 2 the decrease is only 5%.

Please note that might want to consider letting users choose the ordering. That is still not possible with this PR.

### DIFF
--- a/opm/grid/CpGrid.hpp
+++ b/opm/grid/CpGrid.hpp
@@ -253,7 +253,8 @@ namespace Dune
         /// \param clip_z if true, the grid will be clipped so that the top and bottom will be planar.
         /// \param poreVolume pore volumes for use in MINPV processing, if asked for in deck
         void processEclipseFormat(const Opm::EclipseGrid& ecl_grid, bool periodic_extension, bool turn_normals = false, bool clip_z = false,
-                                  const std::vector<double>& poreVolume = std::vector<double>());
+                                  const std::vector<double>& poreVolume = std::vector<double>(),
+                                  bool reorder_k_fastest = false);
 #endif
 
         /// Read the Eclipse grid format ('grdecl').
@@ -261,7 +262,8 @@ namespace Dune
         /// \param z_tolerance points along a pillar that are closer together in z
         ///        coordinate than this parameter, will be replaced by a single point.
         /// \param remove_ij_boundary if true, will remove (i, j) boundaries. Used internally.
-        void processEclipseFormat(const grdecl& input_data, double z_tolerance, bool remove_ij_boundary, bool turn_normals = false);
+        void processEclipseFormat(const grdecl& input_data, double z_tolerance, bool remove_ij_boundary, bool turn_normals = false,
+                                  bool reorder_k_fastest = false);
 
         //@}
 
@@ -323,6 +325,16 @@ namespace Dune
         void setUniqueBoundaryIds(bool uids)
         {
             current_view_data_->setUniqueBoundaryIds(uids);
+        }
+
+        /// \brief Get the lookup needed for eclipse output.
+        ///
+        /// This vector is empty if no reordering is needed
+        /// (i is the fastest index for the current ordering,
+        /// then j and then k).
+        const std::vector<int>& eclipseOuputIndexLookup()
+        {
+            return current_view_data_->output_reordering_index_;
         }
 
         // --- Dune interface below ---
--- a/opm/grid/CpGrid.hpp
+++ b/opm/grid/CpGrid.hpp
@@ -332,7 +332,7 @@ namespace Dune
         /// This vector is empty if no reordering is needed
         /// (i is the fastest index for the current ordering,
         /// then j and then k).
-        const std::vector<int>& eclipseOuputIndexLookup()
+        const std::vector<int>& eclipseOuputIndexLookup() const
         {
             return current_view_data_->output_reordering_index_;
         }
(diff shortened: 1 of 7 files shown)


blattms:
Please also note that there could be side effects for modules apart from opm-simulators (e.g. opm-upscaling). I have not checked those modules.

atgeirr:
We should not change the cell order of the simulator, since the current order is the same as used by both Eclipse and MRST, and compatibility of output for comparisons etc. is important to keep. We should exploit equation/unknown ordering to get better performance, this example shows that the potential is quite good. However, since we must not disturb the high-level order of cells, this must be done in a different way. One way could be using a permutation array when writing to/reading from the residual and Jacobian, but that might cost a little performance, reducing the gain. It would be great to have something flexible enough to allow experimentation with data-dependant orderings as well!

andlaus:
do I understand this correctly: each active cell needs to exhibit three different cell indices: internal, ECL-compressed for output and logically Cartesian for input? sounds like a lot of fun...

atgeirr:
> do I understand this correctly: each active cell needs to exhibit three different cell indices: internal, ECL-compressed for output and logically Cartesian for input? sounds like a lot of fun...

Yes. Currently "internal" and "ECL-compressed" is the same though. Also, "ECL-compressed" is the same as the logical Cartesian order with dropped inactive cells, so it is not arbitrary. Perhaps we could to this just by having another array similar to the one we already have mapping active to logical Cartesian indices, and make the output code use that to map data? Would still make debugging a little less intuitive though (different order between in-memory data and output).

blattms:
@atgeirr Can you give more details on how you rely on the cell ordering for comparing? I.e. how do you compare with Eclipse? In a debugger? I would have assumed that this is done with output files. Here I always assumed that they are independent of the local cell ordering and the global index would be used for the order. In that case there is no change at all.

Having a simple command line switch would let compare just like before.

I disagree that this reordering should be done when setting up the jacobian, etc. At this point we cannot even rely on dealing with a CpGrid, it would clutter the interface. Just like you admitted it might also be slower. Doing this when setting up the grid should be fastest and easiest.

Data dependent reorderings are of course something completely different and cannot be done in the grid.

atgeirr:
> @atgeirr Can you give more details on how you rely on the cell ordering for comparing? I.e. how do you compare with Eclipse? In a debugger? I would have assumed that this is done with output files. Here I always assumed that they are independent of the local cell ordering and the global index would be used for the order. In that case there is no change at all.

You are correct, the main way is to use output files to compare. While debugging it is also useful (if not absolutely necessary) to have the same ordering internally as on output. I think that the output from Flow currently is not permuted to be in logical cartesian order, and that this PR would immediately make all output incompatible, but I am not 100% sure of that. I assume it for now.

> I disagree that this reordering should be done when setting up the jacobian, etc. At this point we cannot even rely on dealing with a CpGrid, it would clutter the interface. Just like you admitted it might also be slower. Doing this when setting up the grid should be fastest and easiest.

Since the optimal ordering probably depends on what linear solver or preconditioner you use, I think that conceptually the right thing to do is to permute as we pass data to the linear solver subsystem. That would also support data dependent reordering nicely. (The alternative is permuting as we pass data to output.) In the last release, MRST added this feature, more or less the way I suggest we do it (and yes, the codes have different goals etc...).

atgeirr:
jenkins build this please

andlaus:
@blattms: Just to be clear: are the 20% and 5% performance improvements which you report in the PR description for the overall simulation time or just for the linear solver part? also, are these improvements per-iteration or does the number of required newton/linear iterations improve?

alfbr:
We are in the process of doing full ensemble tests of this PR, so I am not sure how much detail we should put on those two data points.

blattms:
I was talking about total computation time. I doubt that there can be an improvement per linear iteration as we are using ILU0 that uses the same sparsity pattern as the original matrix and the work per iteration is proportional to the nonzeros. Those cannot change due to different ordering.

blattms:
Even C-files in OPM are compiled with the C++ compiler, right? Shall we still stick to C in e.g. opm/grid/cpgpreprocess/preprocess ? And just in case: What C-Standard would that be?

blattms:
Nope the are not. Sorry for the noise,

atgeirr:
> Nope the are not. Sorry for the noise,

We could switch them to C++ if that helps with something you want to do.

blattms:
Not needed. I was just wondering.

blattms:
The changes should be there now. At least at startup the pictures shown by ResInsight seem correct with the
companion ewoms PR OPM/ewoms#376. Not thoroughly checked, though.
As the default has not changed, jenkins will not be able to test this PR other than for pure compilation errors.

blattms:
Double checked. I was not able to compare Norne as I could persuade run-regressionTest.sh to use an existing NORNE_ATW2013.UNRST file. I guess I need to read some bash scripts.

blattms:
Sorry. Please hold of merging. I forgot to query the parameter in ewoms when setting up the grid and it has therefore no effect until that is fixed.

blattms:
i'll do another test round. (I underestimated the effort for this whole task big time.)

blattms:
On second thought, I think we could and should merge this without the PR in ewoms. Without the ewoms PR reordering cannot be activated and results will remain the same.

The fix in the output needs to be done in ewoms and I do not expect additional changes here.

blattms:
New day, new opinion. Reading through #342 I must assume that there are quite some places relying on a specific ordering. Those will definitely break if we would activate reordering.

PR #334, opened 2018-06-19
Let Entity return geometry as a copy.

blattms:
Geometry is lightweight enough and this is what DUNE's grid interface mandates.
Performance for norne stays the same.
Closes PR #328

PR #341, opened 2018-08-03
Revert accidental commits

blattms:
Just in case that these break something. They should have gone into PR #334 and fix the issues that are
still there. But somehow I messed up completely (including the branch of the PR).


bska:
Thanks.  I'll merge this right away to restore the original state.  Then I absolutely support getting to the point where we honour the requirements of the Dune interface.

blattms:
Thanks.

PR #355, opened 2019-01-28
Add nnc connections to grid

atgeirr:
With this, non-neighbouring connections (NNCs) are passed to the CpGrid class and incorporated as (fake) faces in the grid structure. This is done in order to allow downstream code to handle NNCs without any code changes.

Internally, the NNC faces have the lowest indices, the order of faces within the grid should now be NNC, I, J, K. This preserves the internal convention that the K faces are the two last faces of any cell. This is only of interest in the internal implementation, as we do not iterate over faces/intersections directly in any case.

Work remains to use the feature from Flow, in particular to ensure the transmissibilities are correctly ordered.

@totto82: I have removed (commented out) the code adding NNCs from PINCH, replacing it with the more generic code below. It believe it will result in having for such cells both a K face (that is an internal boundary face) and an NNC face, instead of just a K face that connects to the other cell. I do not think it should have any ill effects, but it would be good if you could check this. What test case is the most relevant here?

@blattms: I think you could review this PR now for snags and bugs, in any case it is not really testable until companion PRs are added. I will still fire off Jenkins though.

### DIFF
--- a/opm/grid/cpgrid/Intersection.cpp
+++ b/opm/grid/cpgrid/Intersection.cpp
@@ -58,6 +58,10 @@ int Intersection::boundaryId() const
                             //                   TOP  : BOTTOM
                             ret = normal_is_in ? 5    : 6; // min(K) : max(K)
                             break;
+                        case NNC_FACE:
+                            // This should not be possible, as NNC "faces" always
+                            // have two cell neighbours and thus are not on the boundary.
+                            OPM_THROW(std::logic_error, "NNC face at boundary. This should never happen!");
                         }
                     }
                 }
@@ -127,6 +131,11 @@ int Intersection::indexInInside() const
         return normal_is_in ? 2 : 3; // min(J) : max(J)
     case K_FACE:
         return normal_is_in ? 4 : 5; // min(K) : max(K)
+    case NNC_FACE:
+        // For nnc faces we return the otherwise unused value -1.
+        // The Dune grid interface is essentially meaningless
+        // for non-neighbouring "intersections".
+        return -1;
     default:
         OPM_THROW(std::runtime_error, "Unhandled face tag: " << tag);
     }
--- a/opm/grid/cpgrid/Intersection.cpp
+++ b/opm/grid/cpgrid/Intersection.cpp
@@ -45,15 +45,15 @@ int Intersection::boundaryId() const
                         enum face_tag tag = pgrid_->face_tag_[f];
 
                         switch (tag) {
-                        case LEFT:
+                        case I_FACE:
                             //                   LEFT : RIGHT
                             ret = normal_is_in ? 1    : 2; // min(I) : max(I)
                             break;
-                        case BACK:
+                        case J_FACE:
                             //                   BACK : FRONT
                             ret = normal_is_in ? 3    : 4; // min(J) : max(J)
                             break;
-                        case TOP:
+                        case K_FACE:
                             // Note: TOP at min(K) as 'z' measures *depth*.
                             //                   TOP  : BOTTOM
                             ret = normal_is_in ? 5    : 6; // min(K) : max(K)
@@ -121,11 +121,11 @@ int Intersection::indexInInside() const
     const bool normal_is_in = !f.orientation();
     enum face_tag tag = pgrid_->face_tag_[f];
     switch (tag) {
-    case LEFT:
+    case I_FACE:
         return normal_is_in ? 0 : 1; // min(I) : max(I)
-    case BACK:
+    case J_FACE:
         return normal_is_in ? 2 : 3; // min(J) : max(J)
-    case TOP:
+    case K_FACE:
         return normal_is_in ? 4 : 5; // min(K) : max(K)
     default:
         OPM_THROW(std::runtime_error, "Unhandled face tag: " << tag);
--- a/opm/grid/cpgrid/processEclipseFormat.cpp
+++ b/opm/grid/cpgrid/processEclipseFormat.cpp
@@ -53,10 +53,13 @@
 #include <fstream>
 #include <iostream>
 #include <initializer_list>
+#include <set>
 #include <utility>
 
 namespace Dune
 {
+
+    using NNCMap = std::set<std::pair<int, int>>;
 
 
     // Forward declarations.
@@ -81,7 +84,7 @@ namespace Dune
         void removeOuterCellLayer(processed_grid& grid);
         // void removeUnusedNodes(processed_grid& grid); // NOTE: not deleted, see comment at definition.
         void buildTopo(const processed_grid& output,
-                       const std::multimap<int,int>& nnc,
+                       const NNCMap& nnc,
                        std::vector<int>& global_cell,
                        cpgrid::OrientedEntityTable<0, 1>& c2f,
                        cpgrid::OrientedEntityTable<1, 0>& f2c,
@@ -145,7 +148,7 @@ namespace cpgrid
             }
         }
 
-        std::multimap<int, int> nnc_cells;
+        NNCMap nnc_cells;
         for (const auto& p : nnc_cells_pinch) {
             nnc_cells.insert(p);
         }
@@ -230,7 +233,7 @@ namespace cpgrid
 
 
     /// Read the Eclipse grid format ('.grdecl').
-    void CpGridData::processEclipseFormat(const grdecl& input_data, const std::multimap<int,int>& nnc, double z_tolerance, bool remove_ij_boundary, bool turn_normals)
+    void CpGridData::processEclipseFormat(const grdecl& input_data, const NNCMap& nnc, double z_tolerance, bool remove_ij_boundary, bool turn_normals)
     {
         // Process.
 #ifdef VERBOSE
@@ -682,11 +685,11 @@ namespace cpgrid
 
 
 
-        std::multimap<int, int> filterNNCs(const processed_grid& output,
-                                           const std::multimap<int,int>& nnc,
-                                           const std::vector<int>& global_to_local)
+        NNCMap filterNNCs(const processed_grid& output,
+                          const NNCMap& nnc,
+                          const std::vector<int>& global_to_local)
         {
-            std::multimap<int, int> filtered_nnc;
+            NNCMap filtered_nnc;
             const int num_faces = output.number_of_faces;
             std::vector<std::pair<int, int>> face_cells(num_faces);
             // Sort all face->cell mappings so that lowest cell number comes first.
@@ -734,7 +737,7 @@ namespace cpgrid
 
 
         void buildFaceToCellNNC(const processed_grid& output,
-                                const std::multimap<int,int>& nnc,
+                                const NNCMap& nnc,
                                 const std::vector<int>& global_cell,
                                 cpgrid::OrientedEntityTable<1, 0>& f2c,
                                 std::vector<int>& face_to_output_face)
@@ -748,7 +751,7 @@ namespace cpgrid
             // the geometry-based grid processing. In that case we
             // should ensure we do not add it twice, and therefore we
             // filter them out first.
-            std::multimap<int, int> filtered_nnc = filterNNCs(output, nnc, global_to_local);
+            NNCMap filtered_nnc = filterNNCs(output, nnc, global_to_local);
             cpgrid::EntityRep<0> cells[2];
             for (const auto& nncpair : filtered_nnc) {
                 const int c1 = global_to_local[nncpair.first];
@@ -766,7 +769,7 @@ namespace cpgrid
 
 
         void buildFaceToCell(const processed_grid& output,
-                             const std::multimap<int,int>& nnc,
+                             const NNCMap& nnc,
                              const std::vector<int>& global_cell,
                              cpgrid::OrientedEntityTable<1, 0>& f2c,
                              std::vector<int>& face_to_output_face)
@@ -820,7 +823,7 @@ namespace cpgrid
 
 
         void buildTopo(const processed_grid& output,
-                       const std::multimap<int,int>& nnc,
+                       const NNCMap& nnc,
                        std::vector<int>& global_cell,
                        cpgrid::OrientedEntityTable<0, 1>& c2f,
                        cpgrid::OrientedEntityTable<1, 0>& f2c,
--- a/opm/grid/cpgrid/processEclipseFormat.cpp
+++ b/opm/grid/cpgrid/processEclipseFormat.cpp
@@ -81,7 +81,7 @@ namespace Dune
         void removeOuterCellLayer(processed_grid& grid);
         // void removeUnusedNodes(processed_grid& grid); // NOTE: not deleted, see comment at definition.
         void buildTopo(const processed_grid& output,
-                       const std::map<int,int>& nnc,
+                       const std::multimap<int,int>& nnc,
                        std::vector<int>& global_cell,
                        cpgrid::OrientedEntityTable<0, 1>& c2f,
                        cpgrid::OrientedEntityTable<1, 0>& f2c,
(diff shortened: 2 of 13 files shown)


atgeirr:
jenkins build this please

atgeirr:
jenkins build this please

opm/grid/cpgrid/Intersection.cpp
```diff
@@ -121,12 +125,17 @@ int Intersection::indexInInside() const
...
+    case J_FACE:
         return normal_is_in ? 2 : 3; // min(J) : max(J)
-    case TOP:
+    case K_FACE:
         return normal_is_in ? 4 : 5; // min(K) : max(K)
+    case NNC_FACE:
+        // For nnc faces we return the otherwise unused value -1.
+        // The Dune grid interface is essentially meaningless
+        // for non-neighbouring "intersections".
+        return -1;
```
blattms: I doubt, that this works with the ewoms discretization.

opm/grid/cpgrid/processEclipseFormat.cpp
```diff
@@ -668,9 +680,24 @@ namespace cpgrid
...
+            // Add nnc connections first, to preserve the
+            // property that top and bottom faces come last
+            // (per cell) in the cell-face mapping.
+            for (const auto& nncpair : nnc) {
+                const int c1 = global_to_local[nncpair.first];
+                const int c2 = global_to_local[nncpair.second];
+                if (c1 < 0 || c2 < 0) {
+                    OPM_THROW(std::runtime_error, "NNC connection requested between inactive cells.");
+                }
+                cells[0].setValue(c1, true);
```
blattms: Here you silently assume that an NNC between to cells may only occur once.
This will not work in general, because:

1. It is totally legal to specify an NNC between already connected cells.
2. It is totally legal to specify the same NNC multiple times. In that case transmissibilities will be added.
3. A specified NNC might also occur from PINCH.

opm/grid/cpgrid/processEclipseFormat.cpp
```diff
@@ -682,7 +709,7 @@ namespace cpgrid
                     cells[cellcount].setValue(fnc[1], false);
                     ++cellcount;
                 }
-
+                /*
```
blattms: Even if the current ewoms code would not need `Intersection::indexInSide()`, we are losing information here because NNCs created by pinch are not treated as regular intersections anymore.

opm/grid/cpgrid/processEclipseFormat.cpp
```diff
@@ -714,7 +742,11 @@ namespace cpgrid
             const int* fp = output.face_ptr;
             for (int face = 0; face < num_faces; ++face) {
                 int output_face = face_to_output_face[face];
-                f2p.appendRow(fn + fp[output_face], fn + fp[output_face+1]);
+                if (output_face == cpgrid::NNCFace) {
+                    f2p.appendRow(fn, fn);
```
blattms: This seems highly dangerous! I do not get what you are doing here. Would you mind to explain? Definitely missing at least a comment.

opm/grid/cpgrid/processEclipseFormat.cpp
```diff
@@ -868,14 +900,22 @@ namespace cpgrid
...
-                double area = polygonArea(face_pts, centroid);
-                face_normals.push_back(normal);
-                face_centroids.push_back(centroid);
-                face_areas.push_back(area);
+                if (output_face == cpgrid::NNCFace) {
+                    // NNC faces are purely topological constructs,
+                    // and do not have any embedded geometry.
+                    face_normals.push_back({-1e100, -1e100, -1e100});
+                    face_centroids.push_back({-1e100, -1e100, -1e100});
+                    face_areas.push_back(-1e100);
```
blattms: This will not work with the ewoms code. Why not try to use some reasonable values?

- Normal should use the two two cell centroids as direction and have length 1
- Centroid should be the midpoint between the two cells
- Area should be one to make sure it can be used with the ewoms discretization code.

blattms:
> Internally, the NNC faces have the lowest indices, the order of faces within the grid should now be NNC, I, J, K. This preserves the internal convention that the K faces are the two last faces of any cell. This is only of interest in the internal implementation, as we do not iterate over faces/intersections directly in any case.
> 

I am pretty sure that the discretization code iterates over intersections.

> Work remains to use the feature from Flow, in particular to ensure the transmissibilities are correctly ordered.

What do you mean by this? Transmissibilities are in an unorderd_map in ewoms. The key is the pair of cell indices  with small cell index first. Therefore we only need to add the transmissibilities from real NNCs which are currently lacking.

opm/grid/cpgrid/Intersection.cpp
atgeirr: Indeed, this required a matching change in ecltransmissibilities.hh, as can be seen in the companion PR. It only required that one change though.

opm/grid/cpgrid/processEclipseFormat.cpp
atgeirr: You are correct, I will amend to make sure it works in those three cases.

opm/grid/cpgrid/processEclipseFormat.cpp
atgeirr: I think that is more precise though: the created faces are not normal K-faces. They will still be intersections though.

opm/grid/cpgrid/processEclipseFormat.cpp
atgeirr: I am adding an empty row, thereby associating no points with the face. I'll add a comment.

opm/grid/cpgrid/processEclipseFormat.cpp
atgeirr: I wanted to ensure the values were not actually used, and stand by that: it is better to get crazy results when you try something wrong. However for the area I had to modify it (already in this PR) for the reason you state. I think the normal and centroid should remain invalid. Reasonable values will hide errors.

blattms:
Well, I am 100% sure that area is used in eclfluxmodule.hh and I think it is
highly probable at least some of the others are too when using the two point
flux approximation.

BTW: Looking at eclfluxmodule.hh I would recommend testing without -DNDEBUG as
I have the feeling that there should be some failing asserts with the current
implementation.

opm/grid/cpgrid/processEclipseFormat.cpp
```diff
@@ -139,12 +139,19 @@ namespace cpgrid
...
-            nnc = mp.process(thickness, z_tolerance, poreVolume, ecl_grid.getMinpvVector(), actnumData, opmfil, zcornData.data());
-            if (opmfil || nnc.size() > 0) {
+            nnc_cells = mp.process(thickness, z_tolerance, poreVolume, ecl_grid.getMinpvVector(), actnumData, opmfil, zcornData.data());
+            if (opmfil || nnc_cells.size() > 0) {
                 this->zcorn = zcornData;
             }
         }
 
+        // Add explicit NNCs to the NNC map.
+        if (nncs.hasNNC()) {
```
blattms: if clause seems superfluous as `nncs.nncdata()` will be empty anyway

opm/grid/cpgrid/processEclipseFormat.cpp
```diff
@@ -668,9 +680,24 @@ namespace cpgrid
...
-            int nf = output.number_of_faces;
             cpgrid::EntityRep<0> cells[2];
             face_to_output_face.clear();
+            // Add nnc connections first, to preserve the
+            // property that top and bottom faces come last
+            // (per cell) in the cell-face mapping.
+            for (const auto& nncpair : nnc) {
+                const int c1 = global_to_local[nncpair.first];
+                const int c2 = global_to_local[nncpair.second];
+                if (c1 < 0 || c2 < 0) {
```
blattms: I do not think that specifiying NNCs with inactive cells should be fatal. But I could be wrong

opm/grid/cpgrid/processEclipseFormat.cpp
blattms: I tried to understand why the test of the companion PR in ewoms fails.

Could it be that we now end up with two intersections in case an NNC was created via PINCH? From the commented code below it seems that there is an intersection between the old cell pairs (before pinching) that is never removed. Instead we now add a new intersection. Hence we should have one intersection with the old (now inactive cell, that hopefully gets zero transmissibility later) and another one with the correct cell.

atgeirr:
A brief update: I have added tests and fixed some bugs, the issues reported in the review should be addressed (except for those noted in earlier comments). I still want to add a test for the PINCH behaviour and refactor a function that became unreadably long before I consider myself done.

atgeirr:
Another thing to note: checking that repeated NNCs behave properly is "too easy" for the grid (the repeats should have no effect). The responsibility for handling the transmissibility properly (adding) in that case is on ecltransmissibility.hh, which cannot be unit tested (only integration tested) for the usual type tag reasons. I'll add an integration test for Flow verifying that the transmissibilities are ok, that will be a separate PR.

atgeirr:
I now consider this PR ready for final review. The question of how to handle PINCH nncs should be decided though (add nnc or replace regular connection), it is not obvious to me that doing it one way or the other is better (and I do not know what other simulators do).

I'll run this through jenkins both by itself, and together with the ewoms PR.

atgeirr:
jenkins build this please

totto82:
I have tested this on a homemade test case and the results change significantly. I will send the case via email.

opm/grid/cpgrid/processEclipseFormat.cpp
```diff
@@ -868,14 +1007,27 @@ namespace cpgrid
...
+                    // NNC faces are purely topological constructs,
+                    // and do not have any embedded geometry.
+                    // However, since the ewoms code will multiply and
+                    // divide by the face area even if not necessary
+                    // for the cell-centered FV discretization (because
+                    // it wants to deal with velocities rather than fluxes),
+                    // we have to set the areas to 1 to avoid trouble.
+                    face_normals.push_back({-1e100, -1e100, -1e100});
+                    face_centroids.push_back({-1e100, -1e100, -1e100});
+                    face_areas.push_back(1.0);
```
blattms: I have to reiterate: Are you sure that these values will work? 

I have a feeling that they will break code at [ecltransmissibility.hh#L212-312](/OPM/ewoms/blob/2e1a6f34ca7a58a1aa1b17ec2a5cde36c2c1f38a/ebos/ecltransmissibility.hh#L212-L312). I would guess that this breaks PINCH for cases without NNC.

atgeirr:
I have now added proper treatment of PINCH connections. Similar to @totto82's original solution, we again replace a K+ face with one that connects the two cells. In addition, the lower cell's K- face is deleted (as it is replaced with the connecting face). I believe that latter is a small improvement relative to the current treatment of these cells.

opm/grid/cpgrid/processEclipseFormat.cpp
```diff
@@ -139,12 +145,32 @@ namespace cpgrid
...
+        }
+        // Add explicit NNCs.
+        for (const auto single_nnc : nncs.nncdata()) {
+            // Repeated NNCs will only exist in the map once
+            // (repeated insertions have no effect), and we make
+            // sure NNCs specified using either order of cells
+            // end up with the same {low, high} pair.
+            // The code that computes the transmissibilities is responsible
+            // for ensuring repeated NNC transmissibilities are added.
+            auto low = std::min(single_nnc.cell1, single_nnc.cell2);
```
blattms: Thinking about why I would prefer to use
```
auto low  = single_nnc.cell1;
auto high = single_nnc.cell2;
if ( low > high ){
  std:swap(low, high)
}
```
makes me wonder how flow should behave if `low == high` holds. The parser just eats it. I guess that is an error in the deck and we should abort?

opm/grid/cpgrid/processEclipseFormat.cpp
```diff
@@ -638,39 +670,126 @@ namespace cpgrid
...
+                    continue;
+                }
+                const int c1 = global_to_local[nncpair.first];
+                const int c2 = global_to_local[nncpair.second];
+                if (c1 < 0 || c2 < 0) {
+                    Opm::OpmLog::warning("nnc_inactive", "NNC connection requested between inactive cells.");
+                    continue;
+                }
+                auto beg = std::lower_bound(face_cells.begin(), face_cells.end(), std::make_pair(c1, -1));
+                auto end = std::upper_bound(face_cells.begin(), face_cells.end(), std::make_pair(c1 + 1, -1));
```
blattms: Do we really want to binary searches over all face_cells?

There should not be two many faces for a cell and therefore the second one could be linear:
```
auto end=std::find(begin, face_cells.end(), [](const auto& pair){return pair->first==c1+1;})
```
As `nnc` is sorted the first one could start with the last end.

opm/grid/cpgrid/processEclipseFormat.cpp
```diff
@@ -638,39 +670,126 @@ namespace cpgrid
...
+                bool found_same_connection = false;
+                for (auto it = beg; it < end; ++it) {
```
blattms: And another find algorithm. Maybe we should merge this with the search for end?

atgeirr:
jenkins build this please

opm/grid/cpgrid/processEclipseFormat.cpp
atgeirr: I think explicit NNCs between a cell and itself should be flagged as an error in the parser. @joakim-hove do you agree?

atgeirr:
jenkins build this please

alfbr:
Build has broken for this pull request. @blattms maybe you can help restore building?

alfbr:
It turns out that our Boost version was too old for the new test here. This PR requires Boost 1.60 or newer. We have now successfully tested the PR. It gives correct results for NNC keyword. @akva2 , we still have a fall-out on the upscaling tests.

bska:
> This PR requires Boost 1.60 or newer.

In that case something's got to give.  We're currently looking for Boost 1.44 during [system probe](https://github.com/OPM/opm-grid/blob/0b9f55571e33a6a7b9d28c180a8df83fe9a21a5b/opm-grid-prereqs.cmake#L26-L27), so we've either got to upgrade OPM's minimum Boost requirement or we need to rewrite the implementation of the new test to support the older versions of Boost.Test.

alfbr:
Actually, it is not important to us. Thanks to the absolutely fantastic work done on OPM's build system we have not had build issues for years. Even on archaic Red Hat 6, non-privileged user, deploying out-of-tree and linking in the latest Boost 1.69 out-of-tree just worked today. Just as we are now on Dune 2.5.2 and GCC 7.3, Swapping out CMAKE variables to the correct version number is all it takes. It is probably the latest Ubuntu LTS or Debian stable that dictates minimum versions these days. Building with default GCC on Red Hat 7 (which is version 4.8) broke some time ago without anybody noticing.

alfbr:
Just checked. Ubuntu 16.04 is on version 1.58, same as we were yesterday. Latest Debian stable, i.e., Jenkins and benchmarking is on version 1.62. The latest Ubuntu LTS is 18.04 which is on 1.65, but I believe 18.04 has issues with the MPI libraries.

blattms:
Turns out supporting older boost is pretty easy. See #356

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1 | 1.035    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8 | 0.958    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1 | 0.982    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8 | 0.917    |



View result details @ https://www.ytelses.com/opm/?page=result&id=222

alfbr:
Hm, performance should not be impacted by this PR. None of the tested models use the NNC keyword. The trigger was unintended though, this PR should be considered with https://github.com/OPM/ewoms/pull/466

PR #356, opened 2019-02-12
Add nnc connections old boost

blattms:
Substitute of #355 with support for older boost versions.


bska:
jenkins build this please

PR #358, opened 2019-02-15
fix build of nnc test with dune 2.4

akva2:
The new handy iterators are not available in dune 2.4.


akva2:
jenkins build this dune2.4 please

Issue #366, opened 2019-05-09
zoltan_test 'MPI_Errhandler_create’ was not declared in this scope


fgfuchs:
When I am compiling, I get the following error:

```
opm-grid/tests/cpgrid/zoltan_test.cpp:100:9: error: ‘MPI_Errhandler_create’ was not declared in this scope
         MPI_Errhandler_create(MPI_err_handler, &handler);
         ^~~~~~~~~~~~~~~~~~~~~
opm-grid/tests/cpgrid/zoltan_test.cpp:100:9: note: suggested alternative: ‘MPI_Errhandler_free’
         MPI_Errhandler_create(MPI_err_handler, &handler);
```
My versions are:
openmpi 4.0.1-1
zoltan 3.83-5

Can anybody help me with this?

akva2:
MPI_Errhandler_create was deprecate in MPI 2 and removed, see https://www.open-mpi.org/faq/?category=mpi-removed#mpi-1-mpi-errhandler-create

akva2:
```diff
diff --git a/tests/cpgrid/zoltan_test.cpp b/tests/cpgrid/zoltan_test.cpp
index ceadb0c..950c9b0 100644
--- a/tests/cpgrid/zoltan_test.cpp
+++ b/tests/cpgrid/zoltan_test.cpp
@@ -97,7 +97,11 @@ BOOST_AUTO_TEST_CASE(zoltan)
         Zoltan_Set_Param(zz, "PHG_EDGE_SIZE_THRESHOLD", ".35");  /* 0-remove all, 1-remove none */
 
         MPI_Errhandler handler;
+#ifdef MPI_2
+        MPI_Comm_create_errhandler(MPI_err_handler, &handler);
+#else
         MPI_Errhandler_create(MPI_err_handler, &handler);
+#endif
         MPI_Errhandler_set(MPI_COMM_WORLD, handler);
         MPI_Comm_size(MPI_COMM_WORLD, &procs);
         MPI_Comm_rank(MPI_COMM_WORLD, &myRank);
```

fgfuchs:
OK, so this means I need to downgrade my openmpi version?

fgfuchs:
I will create a pull request based on your diff @akva2 . Thanks

blattms:
Fixed with merge of #367

PR #367, opened 2019-05-09
deprecated version only if MPI_2 is not defined

fgfuchs:
use deprecated MPI calls only if MPI_2 is not defined

### DIFF
--- a/tests/cpgrid/zoltan_test.cpp
+++ b/tests/cpgrid/zoltan_test.cpp
@@ -97,8 +97,13 @@ BOOST_AUTO_TEST_CASE(zoltan)
         Zoltan_Set_Param(zz, "PHG_EDGE_SIZE_THRESHOLD", ".35");  /* 0-remove all, 1-remove none */
 
         MPI_Errhandler handler;
+#ifdef MPI_2
+        MPI_Comm_create_errhandler(MPI_err_handler, &handler);
+        MPI_Comm_set_errhandler(MPI_COMM_WORLD, handler);
+#else
         MPI_Errhandler_create(MPI_err_handler, &handler);
         MPI_Errhandler_set(MPI_COMM_WORLD, handler);
+#endif
         MPI_Comm_size(MPI_COMM_WORLD, &procs);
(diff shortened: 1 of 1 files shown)


fgfuchs:
jenkins build this please

akva2:
s/defined/not defined

fgfuchs:
I guess the branch should be called update_to_mpi4...

atgeirr:
If `MPI_Errhandler_create` was deprecated in favour of `MPI_Comm_create_errhandler ` in MPI2, why don't we just use `MPI_Comm_create_errhandler` unconditionally? The only risk of failure should be if anyone require MPI1 (the standard that was superseded in 1996...) and that should be ok?

akva2:
i don't think carrying this one ifdef is a high maintenance cost but..

the old symbol was deprecated in MPI 2 and removed in MPI 3. so branch name is correct as such, there is no MPI 4 standard (openmpi 4 implements the MPI 3 standard).

fgfuchs:
closes issue #366

fgfuchs:
jenkins build this please

atgeirr:
I think the logic is backwards now, as the new call is inside the #if and the deprecated is outside?

fgfuchs:
> I think the logic is backwards now, as the new call is inside the #if and the deprecated is outside?

As I understand, version 2 is the new one, version 1 is the deprecated one.

akva2:
It is correct as is. The new symbol is MPI_Comm_create_errhandler. We want to use this for mpi 2 or newer, and the old MPI_Errhandler_create only with v1.

atgeirr:
> We want to use this for mpi 2 or newer, and the old MPI_Errhandler_create only with v1.

Then we should #if test for MPI 1 (as that is the special case) rather than MPI 2, because if we use MPI 3 (as I think I have on my system, not sure though) it will then use the deprecated call and fail.

akva2:
wrong. MPI_3 has the MPI_2 symbols and thus the MPI_2 macro is set.

atgeirr:
Thanks for the explanation, then I am happy with this! Merging.

PR #371, opened 2019-06-27
Send CMake error if we are configuring with MPI but without ZOLTAN.

blattms:
At least for more complex models our loadbalancing without ZOLTAN
is not good enough and may lead to cryptic error message and abortion
of the problem. Therefore this commit resorts to give a good error
message during the CMake run that makes clear that the user needs ZOLTAN.

See OPM/opm-simulators#1910 for support requests that this change will prevent.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -87,6 +87,10 @@ macro (config_hook)
 	list (APPEND ${project}_CONFIG_IMPL_VARS
 		HAVE_DUNE_GRID_CHECKS
 		)
+	if(NOT ZOLTAN_FOUND AND MPI_C_FOUND)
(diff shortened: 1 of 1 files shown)


blattms:
If anybody is using OPM in parallel without ZOLTAN deliberately, please yell. Otherwise I will merge this tomorrow

atgeirr:
Only one afterthought: In order to ensure there is no bug in the rest of the code, we might want to reproduce the error reported in OPM/opm-simulators#1910 to ensure whatever causes the segfault only can manifest with the fall-back partitioner.

blattms:
I agree, just did not find the time yet. Mind reopen the bug or create a new one?

PR #372, opened 2019-07-01
quell unused variable warnings without MPI

akva2:
Since we do not have the warning suppression header available here, do it manually.

if you prefer i can use attribute suppression instead.


atgeirr:
I think it is ok as it is, the attribute suppression would not eliminate the preprocessor directives. Will run jenkins and merge.

atgeirr:

PR #376, opened 2019-08-11
Correct order of function parameters when commputing overlap.

blattms:
Fortunately the wrong order only resulted in always using
overlap 1 due to the way the code is written.


blattms:
jenkins build this please

atgeirr:
Thanks for noticing and fixing! Merging.

PR #378, opened 2019-08-13
quell unused parameter warning when building without zoltan

akva2:
not very pretty, but..


akva2:

akva2:
missed that void casting was already used for the purpose. let's do that here as well.

akva2:

blattms:
Thanks

PR #380, opened 2019-08-14
Use mpi to scatter cell data

blattms:
This is in preparation for a larger change. It should not affect current flow at all since we never scatter data. Will selfmerge once tests are green.

### DIFF
--- a/opm/grid/cpgrid/CpGridData.hpp
+++ b/opm/grid/cpgrid/CpGridData.hpp
@@ -448,10 +448,6 @@ private:
     /// \brief Interface from interior and border to interior and border for the faces.
     std::tuple<InterfaceMap,InterfaceMap,InterfaceMap,InterfaceMap,InterfaceMap>
     point_interfaces_;
-    /// \brief Interface for gathering and scattering cell data.
-    InterfaceMap cell_gather_scatter_interface;
-    /// \brief Interface for gathering and scattering point data.
-    InterfaceMap point_gather_scatter_interface;
 
 #endif
 
--- a/opm/grid/cpgrid/CpGridData.hpp
+++ b/opm/grid/cpgrid/CpGridData.hpp
@@ -301,6 +301,9 @@ private:
     void gatherCodimData(DataHandle& data, CpGridData* global_data,
                          CpGridData* distributed_data);
 
+    /// \brief The type of the interface map for communication.
+    typedef VariableSizeCommunicator<>::InterfaceMap InterfaceMap;
+
     /// \brief Scatter data from a global grid representation
     /// to a distributed representation of the same grid.
     /// \param data A data handle for getting or setting the data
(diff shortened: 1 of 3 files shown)


blattms:
jenkins build this please

blattms:
Thanks jenkins. Should have checked with older DUNE versions. Amend last commit and trying again.

blattms:
jenkins build this please.

blattms:
jenkins build this please

blattms:
jenkins build this please

atgeirr:
> It should not affect current flow at all since we never scatter data.

Just to see I understand this: we never scatter data currently because all processes start with the full grid, and then just extract their own partition of it without needing to ask any other process (except for whatever the zoltan partitioning does). Correct?

blattms:
Kind of. We never call scatterData because we either compute the values or read them from the global deck in ebos/ewoms. 

Before ewoms we used scatterData to scatter properties computed by the master process. Hence even though every process had a complete global deck we simply ignored this fact (as we were on the route to going fully parallel with the grid only on one process, the state was supposed to be intermediate). Unfortunately, going down the ewoms route we did go half step back that we now have before us again...

blattms:
Finally compiles with DUNE 2.5.1 on jenkins

blattms:
BTW There also is  a test for the scatterData in the module that still succeeds.

blattms:
@atgeirr approved?

blattms:
I will selfmerge this this afternoon since the unit tests still run through.
At the same time I will reget becoming a permanent member of the self-merge club as I will have lost the right to complain about self-merging.

atgeirr:
@blattms I'll look at this before lunch, hopefully I can save you from the dreaded self-merger club...

opm/grid/cpgrid/CpGridData.hpp
```diff
@@ -746,13 +747,18 @@ struct Mover<DataHandle,3> : public BaseMover<DataHandle>
...
 void CpGridData::scatterData(DataHandle& data, CpGridData* global_data,
-                          CpGridData* distributed_data)
+                             CpGridData* distributed_data, const InterfaceMap& inf)
 {
 #if HAVE_MPI
     if(data.contains(3,0))
-       scatterCodimData<0>(data, global_data, distributed_data);
+    {
+        Entity2IndexDataHandle<DataHandle, 0> data_wrapper(*global_data, *distributed_data, data);
+        communicateCodim<0>(data_wrapper, ForwardCommunication, inf);
```
atgeirr: This seems to be the focus of the change, what is the advantage to using communicateCodim() rather than scatterCodimData()? Both are using MPI at a lower level I assume? Also, for vertex data scatterCodimData() is still used, do you intend to replace that also with communicateCodim()?

blattms:
Nope. Only the former uses MPI. This is just the first step. For the second more work would be needed
as we need to compute from which indices to send and for which to receive.

blattms:
scatterData assumes that all processes know the global grid as uses local copying. The other one uses comunication and
assumes the global grid only on one process.

codim-3 might follow somewhen if I find the time. AFAIK OPM does not make use of it, does it?

atgeirr:
> codim-3 might follow somewhen if I find the time. AFAIK OPM does not make use of it, does it?

No, not my knowledge.

PR #384, opened 2019-08-16
added: run distribution_test in parallel as well

akva2:
I would guess a serial execution of this test does not really test all parts of the code?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -110,3 +110,11 @@ endmacro (tests_hook)
 
 # all setup common to the OPM library modules is done here
 include (OpmLibMain)
+
+# Additional parallel tests
+if(MPI_FOUND)
+  if(NOT MPIEXEC_EXECUTABLE)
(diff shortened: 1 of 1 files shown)


akva2:
jenkins build this please

blattms:
Just a question. A parallel test is always run in parallel with a fixed certain
number of processes?

akva2:
yes, if we just run it normally that fixed number is 1, i now added a fixed number of 4. i can change to anything you'd like of course.

akva2:
grml, stupid cmake version differences.

akva2:
jenkins build this please

akva2:
okay, now it passed. @blattms merge if you think this has additional value, i did not really look into what the test actually does.

blattms:
Thanks. This certainly has value.

Is there a maximum runtime for parallel tests? Just in case we produce a deadlock...

akva2:
ctest has a default of 1500s, but there's nothing apart from that.

PR #390, opened 2019-09-28
Avoid use of property.iget()

joakim-hove:
Followup to: https://github.com/OPM/opm-common/pull/1051


joakim-hove:
I have just started a new jenkins run on this from opm-common - will merge if it is green.

PR #391, opened 2019-09-30
Remove GridManagere::createGrdecl() which accesses Deck directly

joakim-hove:
It is a goal to reduce/avoid the use of the low level datastructure outside opm-common. This is a small step along the way.


joakim-hove:
jenkins build this please

bska:
Thanks.  This should probably have been a part of commit d17bd4a.

PR #393, opened 2019-10-03
WIP: The eclipse gride preprocessing is not allowed to deactivate cells

joakim-hove:
Related to: https://github.com/OPM/opm-common/pull/1059

Observe that the uglyness in the second commit is to support opm-upscaling: https://github.com/OPM/opm-upscaling/pull/279

### DIFF
--- a/opm/grid/cpgrid/processEclipseFormat.cpp
+++ b/opm/grid/cpgrid/processEclipseFormat.cpp
@@ -112,6 +112,21 @@ namespace Dune
 namespace cpgrid
 {
 
+namespace {
+
+void assert_active_cells(const grdecl& grdecl_grid, const Opm::EclipseGrid& ecl_grid) {
+// #ifndef NDEBUG
+    const auto& ecl_actnum = ecl_grid.getACTNUM();
+    for (std::size_t g = 0; g < ecl_grid.getCartesianSize(); g++) {
+        if (ecl_actnum[g] != grdecl_grid.actnum[g])
+            throw std::logic_error("Cells have been deactivated in the grid processing - should not happen");
+    }
+// #endif
+}
+
+}
+
+
 #if HAVE_ECL_INPUT
     void CpGridData::processEclipseFormat(const Opm::EclipseGrid& ecl_grid, bool periodic_extension, bool turn_normals, bool clip_z,
                                           const std::vector<double>& poreVolume, const Opm::NNC& nncs)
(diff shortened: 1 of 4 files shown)


opm/grid/cpgrid/processEclipseFormat.cpp
```diff
@@ -231,6 +231,14 @@ namespace cpgrid
             // Make the grid.
             processEclipseFormat(g, nnc_cells, z_tolerance, false, turn_normals);
         }
+
+        {
+            const auto& ecl_actnum = ecl_grid.getACTNUM();
+            for (std::size_t g = 0; g < ecl_grid.getCartesianSize(); g++) {
```
atgeirr: Using `g` for cell index is a little weird...? Other than that trivial observation, I'll try to run this through some cases to see if it is triggered.

opm/grid/cpgrid/processEclipseFormat.cpp
```diff
@@ -231,6 +231,14 @@ namespace cpgrid
             // Make the grid.
             processEclipseFormat(g, nnc_cells, z_tolerance, false, turn_normals);
         }
+
```
blattms: Should we maybe deactivate this test in production by `#ifndef NDEDUG`

opm/grid/cpgrid/processEclipseFormat.cpp
```diff
@@ -231,6 +231,14 @@ namespace cpgrid
...
+        {
+            const auto& ecl_actnum = ecl_grid.getACTNUM();
+            for (std::size_t g = 0; g < ecl_grid.getCartesianSize(); g++) {
+                if (ecl_actnum[g] != g.actnum[g])
```
blattms: Are you sure this compiles? Looks like you are calling a member function on a `g` which is of type `size_t` here.

joakim-hove: Hehe - no this probably does not compile ... :-(

opm/grid/cpgrid/processEclipseFormat.cpp
joakim-hove: I like using `g` for `global_index` - happy to change it thougth.

opm/grid/cpgrid/processEclipseFormat.cpp
joakim-hove: Better safe than sorry - I would prefer to leave that test in - even in release mode; can add a commented of `#ifndef NDEBUG` - which can be activated at some point in the future.

blattms:
Closing this one as the companion PR is closed (and pinch processing might also deactivate cells.

There is hope that that processing will move to an earlier stage and the changes in opm-common will really make sense.

PR #394, opened 2019-10-11
Make gatherData/scatterData const functions.

blattms:
These functions should have been const for a long time, but somehow I missed this. Currently this is need for working restart with the global grid only on one process (one can only get a const grid from
the vanguard).

Should also be uncontroversial. Will merge after a build test.


blattms:
jenkins build this please

blattms:
Tests should be fixed with the merge of OPM/opm-simulators#555. Trying jenkins again.

blattms:
jenkins build this please

blattms:
Backported to release via #395

Issue #353, opened 2018-12-18
opm-grid seems to depend on opm-common but is only listed under "suggests"


timokoch:
The file `opm-grid-prereqs.cmake` uses the CMake macro `find_package_deps` which is defined in `opm-common`. So opm-grid depends on opm-common but in `dune.module` it is only listed under `suggests`.

An error occurred in the following setting: Module A suggests (and finds) `opm-grid`, Module B depends on module A (but not on OPM at all); configuring Module B leads to the following CMake error:

```
CMake Error at /home/timok/dumux-course/opm-grid/opm-grid-prereqs.cmake:37 (find_package_deps):
  Unknown CMake command "find_package_deps".
```

I tested version release 2018.10 and used `dunecontrol`. Shouldn't `opm-grid` __depend__ on `opm-common` in that case? At least the call would need an if guard if `opm-common` is really optional.

blattms:
Fixed via #396

PR #396, opened 2019-10-15
Moved opm-common to Depends section of dune-module.

blattms:
According to our build system it is a required dependency and dune.module should be consistent with this.

### DIFF
--- a/dune.module
+++ b/dune.module
@@ -10,5 +10,5 @@ Label: 2019.10-pre
 Maintainer: atgeirr@sintef.no
 MaintainerName: Atgeirr F. Rasmussen
 Url: http://opm-project.org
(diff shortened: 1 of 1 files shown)


blattms:
Closes #353

blattms:
No need to run jenkins as it does not test this. Will query dunecontrol and self-merge afterwards.

blattms:
dunecontrol seems happy on my system:
```
$ dunecontrol --module=opm-grid print
dune-common dune-geometry dune-uggrid dune-grid opm-common dune-istl opm-grid
$
```

blattms:
@akva2 Thanks for the approval. That was really quick.

akva2:
np. it was like that as a compromise for the alternative buildsystem that was floating around..

blattms:
backported to release via PR #398

blattms:
Totally forgot about that alternate thingy.

PR #399, opened 2019-10-16
Renamed functions in opm/grid/polyhedralgrid/grid.hh after changes in previous commit in opm-common

rbe051:
I'm a new Post-doc at Norce.

### DIFF
--- a/opm/grid/polyhedralgrid/grid.hh
+++ b/opm/grid/polyhedralgrid/grid.hh
@@ -797,19 +797,15 @@ namespace Dune
         const auto eclipseGrid = std::make_shared<Opm::EclipseGrid>(deck, rawactnum);
 
         struct grdecl g;
-        std::vector<int> actnum;
-        std::vector<double> coord;
-        std::vector<double> zcorn;
-        std::vector<double> mapaxes;
 
         g.dims[0] = eclipseGrid->getNX();
         g.dims[1] = eclipseGrid->getNY();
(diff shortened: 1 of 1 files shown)


atgeirr:
> I'm a new Post-doc at Norce.

Welcome to OPM!

I'll run Jenkins on this although I think it will not pick up any mistakes, as there are no tests for this grid (which is why I think this change was not caught earlier).

atgeirr:
jenkins build this please

opm/grid/polyhedralgrid/grid.hh
```diff
@@ -797,19 +797,15 @@ namespace Dune
...
         g.dims[0] = eclipseGrid->getNX();
         g.dims[1] = eclipseGrid->getNY();
         g.dims[2] = eclipseGrid->getNZ();
-
-        eclipseGrid->exportMAPAXES( mapaxes );
-        eclipseGrid->exportCOORD( coord );
-        eclipseGrid->exportZCORN( zcorn );
-        eclipseGrid->exportACTNUM( actnum );
+        
+        std::vector<double> mapaxes = eclipseGrid->getMAPAXES( );
```
akva2: these can be const refs afaict.

blattms:
release or not release, that is the question.

atgeirr:
> release or not release, that is the question.

This is a compile failure for the code in question, so I added it to the release milestone.

PR #389, opened 2019-09-20
Hold global grid only on process

blattms:
This is a refactoring of the grid distribution code. We now do not assume that the grid is stored on every processed but is only available on the master process. Therefore we now use communication to set up the distributed grid.

The WIP as there might stil be some issues. Currently this is only tested stand-alone.

PR #404, opened 2019-10-21
Revert to parallel working state where all processes know global grid.

blattms:
Revert "Merge pull request #389 from blattms/hold-global-grid-only-on-process"

This reverts commit 079d46b5318d19fe351afae25685f29e162ddc0f, reversing
changes made to 97d29109284035b280435a3e071f706a943fe7bd.

This is done as the underlying error introduced in PR #389 might need
some inverstigation


atgeirr:
Too bad, but it seems prudent. I'll merge this.

blattms:
Just a short heads up. I might have found the bug (and the whole town know sbecause of my partying about it). Expect a revert of the revert coming up today.

PR #406, opened 2019-10-23
Fixed overlap computation

blattms:
PR #376 introduced a subtle bug with great percussions for parallel
runs. As said in that PR overlap_layers was at the wrong position but
the overlap was still computed correctly. The important thing was that
the last parameter evaluated to true to make the process compute the
overlap for all processes. This was needed but changed in the PR (and
I imagine that the overlap was at least not correct, maybe even not there at
all).

With this PR we set the last parameter of addOverlapLayers to true and
get rid of the excessive time step chopping experienced in parallel
runs.

No more warnings about "finding the bubble point pressure failed  for cells..."

### DIFF
--- a/opm/grid/cpgrid/CpGridData.cpp
+++ b/opm/grid/cpgrid/CpGridData.cpp
@@ -534,7 +534,7 @@ void CpGridData::distributeGlobalGrid(const CpGrid& grid,
     std::vector<std::set<int> > overlap;
 
     overlap.resize(cell_part.size());
-    addOverlapLayer(grid, cell_part, overlap, my_rank, overlap_layers, false);
+    addOverlapLayer(grid, cell_part, overlap, my_rank, overlap_layers, true);
     // count number of cells
     struct CellCounter
     {


blattms:
PR was tested on top of OPM as of August 12, 2019 with model2 and fixed the timestep chopping.

That leaves the deadlock to be fixed.

blattms:
jenkins build this please

bska:
> PR was tested on top of OPM as of August 12, 2019 with model2 and fixed the timestep chopping.

Excellent find!  Do you know of a process&mdash;preferably automated or at least semi-automated&mdash;that would have discovered this earlier?  Developers being super-extra-especially careful doesn't really scale to unfamiliar territories.

atgeirr:
In retrospect, this can be seen at http://www.linuxbenchmarking.com/?open-porous-media-git to some extent. For the "MPI Extra 8 threads" case there is a gap in time (not obvious unless you look at the dates) from July 25 to August 30. Across that gap, runtime increased significantly. Another such increase happened around September 21, we have tried to identify its cause, but not quite succeeded (but we have identified some suspicious behaviour in the process, such as the z-bounds of the ODE solved for EQUIL initialization depending on the partition's max and min z-bounds).

I hope that with this, things will improve! Merging.

blattms:
backported to release via #408

PR #413, opened 2019-10-29
Added entries to changelog concerning 2019.10-2018.04 releases

blattms:


blattms:
backported to release via #414

PR #416, opened 2019-11-11
[feature][PolyhedralGrid] Make PolyhedralGrid fully coherent with the DUNE grid interface.

dr-robertk:
This PR make the implementation of PolyhedralGrid fully compliant with DUNE grid interface and enables DUNE grid checks. Also, some minor compilation errors are fixed for CpGrid and current DUNE master.

### DIFF
--- a/opm/grid/polyhedralgrid/geometry.hh
+++ b/opm/grid/polyhedralgrid/geometry.hh
@@ -3,12 +3,16 @@
 #ifndef DUNE_POLYHEDRALGRID_GEOMETRY_HH
 #define DUNE_POLYHEDRALGRID_GEOMETRY_HH
 
+#include <memory>
+
 #include <dune/common/version.hh>
 #include <dune/common/fmatrix.hh>
 #include <dune/grid/common/geometry.hh>
 
 #if DUNE_VERSION_NEWER(DUNE_GEOMETRY, 2, 5 )
+#include <dune/geometry/referenceelements.hh>
 #include <dune/geometry/type.hh>
+#include <dune/geometry/multilineargeometry.hh>
 #else
 #include <dune/geometry/genericgeometry/geometrytraits.hh>
 #include <dune/geometry/genericgeometry/matrixhelper.hh>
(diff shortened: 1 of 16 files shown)


dr-robertk:
jenkins build this please

atgeirr:
Do you want any reviewing or checking of the PolyhedralGrid parts? The CpGrid-related parts look fine.

dr-robertk:
If you want to spend time that would be good, but I understand it's a big PR and I just squashed things together. This PR basically added a lot of things I had implemented over the last years in various different branches. The fact that test_polyhedralgrid passed the DUNE grid check gives enough confidence and that the implementation is correct. What may be interesting is the Geometry implementation where I use the DUNE geometries to get the local-global mappings. That may be something for CpGrid also.

dr-robertk:
jenkins build this please

atgeirr:
> The fact that test_polyhedralgrid passed the DUNE grid check gives enough confidence and that the implementation is correct.

I agree that gives some confidence.

> What may be interesting is the Geometry implementation where I use the DUNE geometries to get the local-global mappings. That may be something for CpGrid also.

Yes, that is interesting. I am curious: how did you deal with the generic case when a cell has no reasonable reference element? How do you distinguish between those cases? (I'd be happy to read the code, given a pointer.)

opm/grid/polyhedralgrid/geometry.hh
```diff
@@ -41,6 +45,70 @@ namespace Dune
     typedef Dune::FieldVector< ctype, coorddimension > GlobalCoordinate;
     typedef Dune::FieldVector< ctype, mydimension >    LocalCoordinate;
 
+    typedef typename Grid::Traits::ExtraData  ExtraData;
+    typedef typename Grid::Traits::template Codim<codimension>::EntitySeed EntitySeed;
+
```
dr-robertk: @atgeirr : You can start reading from here.

atgeirr: Thanks, I looked at it and saw some things that were new to me. I did notice a minor possible bug, will make a note in the diff.

opm/grid/polyhedralgrid/geometry.hh
```diff
@@ -41,6 +45,70 @@ namespace Dune
...
+        Storage( ExtraData data, EntitySeed seed )
+          : data_( data ), seed_( seed )
+        {}
+
+        Storage( ExtraData data )
+          : data_( data ), seed_()
+        {}
+
+        ExtraData data() const { return data_; }
```
atgeirr: Should this rather return a const reference? I imagine ExtraData is not something you want to copy for every call to corner() or center()?

dr-robertk: ExtraData is of type Grid* so this is no problem (see grid.hh). EntitySeed is essentially the element index or similar so copying this is also fine.

atgeirr:
jenkins build this please

opm/grid/polyhedralgrid/geometry.hh
atgeirr: Ok, if you want this merged now that is fine with me.

dr-robertk:
It's ok from my side and would nice if it could be merged.

akva2:
this broke build on ubuntu xenial (dune 2.4), ref https://ci.opm-project.org/job/opm-nightly-xenial/164/console

PR #405, opened 2019-10-22
Hold global grid only on process (2nd try)

blattms:
This is my second try of #389. This time I did test with parallel Norne. This PR also includes @akva2's #403 to make it compile without MPI.

The fix is in commit 690e08c. The communication interfaces where not consistent if the cells had vertices besides their corners. Small bug with huge impact.

I would still like to run a test with another model, but I need to adjust it a bit since the parser is more strict
about whitespace now.

PR #418, opened 2019-11-14
Hold global grid only on process (2nd try, rebased on current master)

blattms:
This is a corrected and rebased version of #405. Due to the non-linear history there there have been
some glitches during rebasing it which could only be removed get rid of some merges there and cherry-picking them back.

This is now a working version of have the global grid only on one process in parallel. Tested on SPE, and norne. Tests on model2 still running.

Needs OPM/opm-simulators#2161 as otherwise wells will be ignored on non-root processes.

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -32,6 +32,7 @@ list (APPEND MAIN_SOURCE_FILES
   opm/grid/cpgrid/Intersection.cpp
   opm/grid/cpgrid/CpGridData.cpp
   opm/grid/cpgrid/CpGrid.cpp
+  opm/grid/cpgrid/DataHandleWrappers.cpp
   opm/grid/cpgrid/GridHelpers.cpp
   opm/grid/cpgrid/PartitionTypeIndicator.cpp
   opm/grid/cpgrid/processEclipseFormat.cpp
@@ -152,6 +153,7 @@ list (APPEND PUBLIC_HEADER_FILES
   opm/grid/common/p2pcommunicator_impl.hh
   opm/grid/cpgrid/CartesianIndexMapper.hpp
   opm/grid/cpgrid/CpGridData.hpp
+  opm/grid/cpgrid/DataHandleWrappers.hpp
   opm/grid/cpgrid/DefaultGeometryPolicy.hpp
   opm/grid/cpgrid/dgfparser.hh
   opm/grid/cpgrid/Entity2IndexDataHandle.hpp
@@ -216,6 +218,7 @@ list (APPEND PUBLIC_HEADER_FILES
   opm/grid/utility/RegionMapping.hpp
   opm/grid/utility/SparseTable.hpp
   opm/grid/utility/StopWatch.hpp
+  opm/grid/utility/VariableSizeCommunicator.hpp
   opm/grid/utility/VelocityInterpolation.hpp
   opm/grid/utility/WachspressCoord.hpp
   opm/grid/utility/ErrorMacros.hpp
--- a/opm/grid/CpGrid.hpp
+++ b/opm/grid/CpGrid.hpp
@@ -50,7 +50,11 @@
 #include <dune/common/version.hh>
 
 #if HAVE_MPI
+#if DUNE_VERSION_NEWER(DUNE_GRID, 2, 7)
 #include <dune/common/parallel/variablesizecommunicator.hh>
+#else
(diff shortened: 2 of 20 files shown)


blattms:
jenkins build this opm-simulators=2161 please.

blattms:
jenkins build this opm-simulators=2161 please

joakim-hove:
> Could be due to different timestepping, but to test this hypothesis I would need to know how often and when entries are written to them.

The summary data is written by a call to `writeOutput()` in `opm-simulators/ebos/eclwriter.hh` - which will write summarydata to disk *and* update the mutable `SummaryState` object which is owned by the `eclvanguard`. The latter is subsequently broadcasted to all nodes iin `evalSummaryState()`

bska:
> Seems like these vectors (WBHP,WGPR, WOPR, WWPR) are too short for all wells

As @joakim-hove says, the size of those vectors is determined exclusively by the number of times that the output process calls [`EclipseIO::writeTimeStep()`](https://github.com/OPM/opm-common/blob/6388fbe5849ca7ca14abfdcea8404602020897a9/src/opm/output/eclipse/EclipseIO.cpp#L193-L201) in opm-common.  That output, in turn, is initiated in opm-simulator's [`EclWriter<>::writeOutput()`](https://github.com/OPM/opm-simulators/blob/2a5c93356ce4f27afea689c170fa22e5e4803e17/ebos/eclwriter.hh#L384-L391) function.  If the timestepping is different then different sizes of summary vectors written at each timestep is a plausible outcome.

joakim-hove:
> Seems like these vectors (WBHP,WGPR, WOPR, WWPR) are to short for all wells (difference is 3 entries)

If it is due to timestepping differences I would assume that *all* vectors had different length? I don't actually understand how it is possible for some vectors to have different length?

blattms:
I do not think that vectors unrelated to wells are checked per default. Hence these are all that are checked.

bska:
> I do not think that vectors unrelated to wells are checked per default. Hence these are all that are checked.

It is true that
```
compareECL -i
```
will run an integration test that only looks at WOPR, WGPR, WWPR and WBHP.  I can't quite tell whether the regression tests use `-i` or not.  Maybe @akva2 would be able to answer.

In any case an easy way to check the vector sizes in the two result sets would be to convert the `.UNSMRY` file of both datasets to formatted (`convertECL CASE.UNSMRY`) and then `grep -c PARAMS` on the resulting `.FUNSMRY` files.  Each summary vector has a size that is equal to the number of `PARAMS` keywords in the summary files.

akva2:
the -i option is not currently in use.

blattms:
Finally, this is ready to be merged. With the last commit we are playing it safe and resembling the same
iteration counts and well curves as for master (tested with 16 MPI processes on model2).

I am a bit surprised that add cells to the overlap that only share corners interior cells and not faces has any effect on convergence. That might be a hint that there is still a glitch in some parallelisation as this
should not have any effect due to the concept (unless I missed something as I do far too often).

Anyway this PR is playing it as safe as possible now. Just one more jenkins run and then it should go in.

blattms:
Will do another local serial build before firing up jenkins.

blattms:
jenkins build this opm-simulators=2161 serial parallel please

blattms:
jenkins run norne parallel please

akva2:
It is norne_parallel

blattms:
I don't get jenkins... Why does it build again (and probably without the right downstream)? I just wanted to run the test with the previously compiled version. Probably its my fault again.

blattms:
jenkins run norne_parallel opm-simulators=2161 please

atgeirr:
Thanks a lot for your effort on this Markus, I'll merge now.

blattms:
Thanks for the fast merge.

akva2:
Maybe intentional, I don't know, but this broke build on ubuntu xenial (dune 2.4).
```
opm/grid/common/GridPartitioning.cpp:40 No such file dune/istl/owneroverlapcopy.hh
```

blattms:
Not intentional at all. Will fix.

blattms:
Or better: Would fix if I had a clue what is wrong. Since this file is there in dune-istl 2.4 I am bit stunned. Missing include directive? Will need to start Ubuntu for this and see for myself.

akva2:
don't. i'm sorry for the false alarm, i had a missing install in my vm. however, there's still an issue cause dune-istl is marked as an optional dependency. should it be a hard dep?

PR #419, opened 2019-11-19
fix build with dune 2.4 after PR #416

akva2:


akva2:
jenkins build this dune2.4 please

akva2:
@dr-robertk while this does the job, i'm afraid it's not a proper fix since the HAVE_DUNE_GRID_CHECK is also false on my system, and it compiles without them. please advice.

dr-robertk:
@akva2: Thanks for looking into this. The HAVE_DUNE_GRID_CHECK is a good idea. I think you could just use #ifdef DUNE_GRID_TEST_GRIDCHECK_HH which is the header guard for this. It's also fine if this gridcheck is just disabled for dune2.4 because these newer features should only be used with a fairly new DUNE version.

akva2:
okay, i made it dependent on dune >= 2.5 like the header include.

akva2:
jenkins build this dune2.4 please

dr-robertk:
@akva2: Thanks for fixing this. Just merge it please when the tests go through.

PR #422, opened 2019-11-21
Some janitoring

akva2:
Quell warnings introduced in PR #416

### DIFF
--- a/opm/grid/polyhedralgrid/gridview.hh
+++ b/opm/grid/polyhedralgrid/gridview.hh
@@ -57,6 +57,7 @@ namespace Dune
     PolyhedralGridView ( const Grid &grid, const int level = 0 )
     : grid_( &grid )
     {
+      (void)level;
     }
 
     const Grid &grid () const
--- a/opm/grid/cpgrid/Indexsets.hpp
+++ b/opm/grid/cpgrid/Indexsets.hpp
@@ -336,10 +336,12 @@ namespace Dune
             {
                 return(*mapping_)[i];
             }
-            else
+            else if (grid_)
             {
                 return i - grid_->size(0) - grid_->size(1) - grid_->size(2);
             }
+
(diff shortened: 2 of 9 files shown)


akva2:
jenkins build this please

akva2:
jenkins build this please

akva2:
jenkins build this please

akva2:
jenkins build this please

akva2:
@atgeirr @blattms my OCD is still twiching. this kills approxmately 50 warnings.

opm/grid/polyhedralgrid/gridview.hh
```diff
@@ -54,7 +54,7 @@ namespace Dune
     static const bool conforming = Traits :: conforming;
     static const PartitionIteratorType pitype = Traits :: pitype;
 
-    PolyhedralGridView ( const Grid &grid, const int level = 0 )
+    PolyhedralGridView ( const Grid &grid, const int = 0 )
```
atgeirr: This is the only thing I request a change for: without the name it is impossible to see what the parameter is usually used for (i.e. what is not used for PolyhedralGrid), so I would suggest commenting it instead (with C comments), using `OPM_UNUSED` or a `(void)level;` inside the method.

In other cases it is clear what the unused argument is, so I'm fine with just deleting the name.

akva2:
jenkins build this please

opm/grid/polyhedralgrid/gridview.hh
akva2: did the (void) thing as the macro is not available in this repo.

blattms: I think that is a fair thing to do hear. Basically the constructors here are part of the interface to make them callable the same as if it was a CpGrid even so the parameters make no sense.

blattms:
48 fixed warnings. In it goes.

PR #423, opened 2019-11-21
Reduce the value of MPI tags to avoid crash

andrthu:
Parallel flow crashes on my computer due to too large MPI tag values in the MPI_send calls in the files changed in this PR.

According to https://github.com/lanl/SNAP/issues/6, the lowest allowed upper bound for MPI tags is  32767. Changed the tags to reflect this.


atgeirr:
jenkins build this please

atgeirr:
Thanks for the fix, I'll merge. I do suggest that MPI tags in the future are named using a shared enum that distinguishes the messages by name, rather than using arbitrary numbers. That would communicate better what the send/receive in question is doing.

PR #429, opened 2019-12-09
Add ENABLE_3DPROPS_TESTING switch in opm-grid

joakim-hove:
Add #ifdef around the only 3D properties getter in opm-grid


joakim-hove:
jenkins build this please

PR #430, opened 2019-12-11
[For testing] Enable "owners first" ordering of cells in parallel grid.

andrthu:
Enable an "owners first" ordering by adding an argument in the loadBalance function. 

In addition, a new addOverlapLayerNoZeroTrans method is added. In this method ghost cells will not be added if the transmissibility on the interface between the interior and potential overlap cell is 0. 

Since the transmissibility is zero, the neighbour cell is not needed, and by not adding this cell, we can avoid some parallel overhead.

This PR require changes in opm-simulators.

### DIFF
--- a/opm/grid/common/GridPartitioning.cpp
+++ b/opm/grid/common/GridPartitioning.cpp
@@ -278,6 +278,16 @@ void addOverlapCornerCell(const CpGrid& grid, int owner,
                           const std::vector<int>& cell_part,
                           std::vector<std::tuple<int,int,char>>& exportList)
 {
+    // Add corner cells to the overlap layer. Example of a subdomain of a 4x4 grid
+    // with and without corner cells in the overlap is given below. Note that the
+    // corner cell is not needed for cell centered finite volume schemes. 
+    // I = interior cells, O = overlap cells and E = exterior cells.
+    //
+    //  With corner     Without corner
+    //  I I O E         I I O E 
+    //  I I O E         I I O E 
+    //  O O O E         O O E E
+    //  E E E E         E E E E
     using AttributeSet = Dune::cpgrid::CpGridData::AttributeSet;
     const CpGrid::LeafIndexSet& ix = grid.leafIndexSet();
     int my_index = ix.index(from);
@@ -365,7 +375,7 @@ void addOverlapLayer(const CpGrid& grid, int index, const CpGrid::Codim<0>::Enti
     void addOverlapLayer(const CpGrid& grid, int index, const CpGrid::Codim<0>::Entity& e,
                          const int owner, const std::vector<int>& cell_part,
                          std::vector<std::tuple<int,int,char>>& exportList,
-                         int recursion_deps)
+                         bool addCornerCells, int recursion_deps)
     {
         using AttributeSet = Dune::cpgrid::CpGridData::AttributeSet;
         const CpGrid::LeafIndexSet& ix = grid.leafIndexSet();
@@ -381,9 +391,9 @@ void addOverlapLayer(const CpGrid& grid, int index, const CpGrid::Codim<0>::Enti
                     {
                         // Add another layer
                         addOverlapLayer(grid, nb_index, *(iit->outside()), owner,
-                                        cell_part, exportList, recursion_deps-1);
+                                        cell_part, exportList, addCornerCells, recursion_deps-1);
                     }
-                    else
+                    else if (addCornerCells)
                     {
                         // Add cells to the overlap that just share a corner with e.
                         for (CpGrid::LeafIntersectionIterator iit2 = iit->outside()->ileafbegin();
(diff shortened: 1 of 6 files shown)


blattms:
Would it make sense to make this PR mergeable without the other one, by making the additional arguments to loadbalance optional?

blattms:
> Would it make sense to make this PR mergeable without the other one, by making the additional arguments to loadbalance optional?

@andrthu is this now mergable without downstream?

opm/grid/common/GridPartitioning.cpp
```diff
@@ -419,7 +452,8 @@ void addOverlapLayer(const CpGrid& grid, int index, const CpGrid::Codim<0>::Enti
             int index = ix.index(*it);
             auto owner = cell_part[index];
             exportProcs.insert(std::make_pair(owner, 0));
-            addOverlapLayer(grid, index, *it, owner, cell_part, exportList, layers-1);
+            //addOverlapLayer(grid, index, *it, owner, cell_part, exportList, layers-1);
+            addOverlapLayerNoZeroTrans(grid, index, owner, cell_part, exportList, layers-1, trans);
```
blattms: What will happen if loadbalance is called without transmissibilities? Please add a test for this situation. Also: Please remove the old code rather than commenting it out.

opm/grid/common/ZoltanGraphFunctions.hpp
```diff
@@ -167,7 +167,7 @@ class CombinedGridWellGraph
     double edgeWeight(int face_index) const
     {
         if (edgeWeightsMethod_ == uniformEdgeWgt)
-            return 1.0;
+            return transmissibilities_[face_index] == 0.0 ? 0.0 : 1.0;
```
blattms: Here we seem to rely on transmissibilties being there, too. That seems a bit odd for uniform weights. Isn't transmissibilities even a nullptr if loadbalance is called without it.

opm/grid/common/GridPartitioning.cpp
```diff
@@ -402,11 +402,44 @@ void addOverlapLayer(const CpGrid& grid, int index, const CpGrid::Codim<0>::Enti
...
+    void addOverlapLayerNoZeroTrans(const CpGrid& grid, int index,
+                                    const int owner, const std::vector<int>& cell_part,
+                                    std::vector<std::tuple<int,int,char>>& exportList,
+                                    int recursion_deps, const double* trans)
+    {
+        using AttributeSet = Dune::OwnerOverlapCopyAttributeSet::AttributeSet;
+        for (int loc_face = 0; loc_face < grid.numCellFaces(index); ++loc_face) {
+            int face = grid.cellFace(index, loc_face);
+            int faceCell0 = grid.faceCell(face, 0);
+            int faceCell1 = grid.faceCell(face, 1);
```
blattms: Are we sure that there always is a neighbor?

andrthu:
Regarding loadBalance and transmissibilities=NULL:

The distribution_test now passes (mpirun -np 4 distribution_test). In distribution_test the methods grid.loadBalance() and grid.loadBalance(data) are called. When these versions of loadBalance are called the default edgeWgt strategy is used, which is transmissibility, but when transmissibility=NULL, this results in uniform(=1) edge weights in the Zoltan partitioner.

What would probably have failed before, and still fails without this PR or not, is: grid.loadBalance(logWgt,wells,trans=NULL). I now believe I have fixed this.

opm/grid/common/GridPartitioning.cpp
andrthu: If face is on the boundary otherCell == -1.

blattms:
Please update the documentation of loadBalance and scatterGrid. Please also add some sentences about how the transmissibilities are now used and what neighbor cells will not become part of the overlap.

blattms:
Other than that this looks fine.

blattms:
I guess dropping the neighbours connected with transmissibility zero is important? Do you have some numbers comparing dropping and adding them?

blattms:
thanks for the rebase.

opm/grid/common/ZoltanGraphFunctions.hpp
```diff
@@ -135,7 +135,7 @@ class CombinedGridWellGraph
                           const std::vector<OpmWellType> * wells,
                           const double* transmissibilities,
                           bool pretendEmptyGrid,
-			  EdgeWeightMethod edgeWeightsMethod);
+                          EdgeWeightMethod edgeWeightsMethod);
```
blattms: tabs should be removed

opm/grid/common/ZoltanPartition.cpp
```diff
@@ -77,7 +77,7 @@ zoltanGraphPartitionGridOnRoot(const CpGrid& cpgrid,
                                                        wells,
                                                        transmissibilities,
                                                        partitionIsEmpty,
-						       edgeWeightsMethod));
+                                                       edgeWeightsMethod));
```
blattms: no tabs please.

opm/grid/common/ZoltanGraphFunctions.hpp
andrthu: That is a removal of tabs. With this PR there are no tabs in ZoltanGraphFunctions.hpp. If you do: git grep -P "\t", you see where the tabs are.

opm/grid/common/ZoltanPartition.cpp
andrthu: Same as above. No tabs in ZoltanPartition.cpp.

blattms:
Great thanks. Then I misinterpreted the diff. Sorry for the noise.

blattms:
jenkins build this opm-simulators=2220 please

PR #428, opened 2019-12-09
Make utility function UgGridHelpers::createACTNUM()

joakim-hove:
The functionality to map between a list of global cells (opm-grid style) and an actnum array (opm-common/Eclipses style) is required for the `FieldPropsManager`. I suggest this implementation - but it could also be located in opm-common if you feel it does not fit in opm-grid.

Also: observe that this is an away match for me - please advise if I have done things against convention.

PR #432, opened 2019-12-12
Implement GridHelper createACTNUM() based on manual overload

joakim-hove:
Hmm - struggled getting the template based solution in #428 to compile from opm-simulators.

### DIFF
--- a/opm/grid/GridHelpers.cpp
+++ b/opm/grid/GridHelpers.cpp
@@ -57,6 +57,11 @@ const int* cartDims(const UnstructuredGrid& grid)
     return grid.cartdims;
 }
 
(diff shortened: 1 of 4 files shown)


joakim-hove:
jenkins build this please

blattms:
What was the struggle?

Using overloading is fine of course.

joakim-hove:
> What was the struggle?

To be 100% honest - I don't really know - the number of moving parts is quite high on my machine; but the end result was at least that `flow` would not link. The simple overload based method works - and "works" is hard to beat when Christmas is approaching.

blattms:
Here you are: :hammer: :nails:

PR #435, opened 2020-01-13
NNC Test: Follow-Up to Removal of ENABLE_3DPROPS_TESTING

bska:
Otherwise, we'll have a build error if ENABLE_3DPROPS_TESTING is not defined.

### DIFF
--- a/tests/cpgrid/grid_nnc.cpp
+++ b/tests/cpgrid/grid_nnc.cpp
@@ -69,15 +69,11 @@ struct Fixture
     {
         Opm::EclipseState es(parser.parseFile(filename));
         std::vector<double> porv;
-#ifdef ENABLE_3DPROPS_TESTING
+
         if (use_deck_porv) {
             porv = es.fieldProps().porv(true);
         }
-#else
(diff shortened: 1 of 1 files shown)


bska:
jenkins build this please

bska:
The [nightly xenial (Ubuntu 16.04 LTS)](https://ci.opm-project.org/job/opm-nightly-xenial/220/console) build fails in the same way that lead me to create this PR:
```
${ROOT}/opm-grid/tests/cpgrid/grid_nnc.cpp: In member function 'void Fixture::testCase(const string&, const Opm::NNC&, int, int, int, const std::vector<std::pair<int, int> >&, bool)':
${ROOT}/opm-grid/tests/cpgrid/grid_nnc.cpp:78:23: error: 'class Opm::EclipseState' has no member named 'get3DProperties'
             porv = es.get3DProperties().getDoubleGridProperty("PORV").getData();
```

If there are no objections, I'd like to merge this now to restore the default build (no additional preprocessor flags)

joakim-hove:
Go I would say!

PR #441, opened 2020-02-10
Make change processEclipseFormat backwards compatible.

blattms:
It used get a reference to EclipseGrid and since PR #433 it only takes a pointer to it. This is unfortunate since this code is also used by other projects such as dumux and breaks compilation of them when using master. This commit adds a processEclipseFormat using a reference to fix this.


blattms:
jenkins build this please

blattms:
Fine with me.

There is also a code snippet that caused problems: [project-config.cmake.in#L73-L76](/OPM/opm-common/blob/01b0dcaedfa390e38ada62d88671754a9815d872/cmake/Templates/opm-project-config.cmake.in#L73-L76)

PR #443, opened 2020-02-13
Do not use bind2nd, which is removed in C++17.

atgeirr:
I am curious why this passed on Jenkins?


atgeirr:
jenkins build this please

atgeirr:
This is simple, but not so trivial that I'll self-merge it. It is required for me to compile though, so I hope someone can look at it and merge it rather quickly. @blattms, @bska or @akva2 ?

akva2:
less bind, more correct. gcc is more relaxed on removed stuff, hence it passed jenkins.

PR #446, opened 2020-02-19
Lessen boost usage

akva2:
- remove dead includes
- avoid boost to get value of pi

### DIFF
--- a/opm/grid/GridUtilities.cpp
+++ b/opm/grid/GridUtilities.cpp
@@ -21,13 +21,18 @@
 #include <opm/grid/GridHelpers.hpp>
 
 #include <opm/grid/utility/platform_dependent/disable_warnings.h>
-#include <boost/math/constants/constants.hpp>
 #include <opm/grid/utility/platform_dependent/reenable_warnings.h>
 
 #include <set>
 #include <vector>
 #include <cmath>
 #include <algorithm>
+
+namespace {
+    double pi() {
+        return 3.14159265358979323846264338327950288;
+    }
+}
 
 namespace Opm
 {
@@ -119,7 +124,7 @@ namespace Opm
                 const double v[2] = { grid.cell_centroids[2*cell2] - grid.cell_centroids[2*cell],
                                       grid.cell_centroids[2*cell2 + 1] - grid.cell_centroids[2*cell + 1] };
                 // The formula below gives an angle in [0, 2*pi] with the positive x axis.
-                const double angle = boost::math::constants::pi<double>() - std::atan2(v[1], -v[0]);
+                const double angle = pi() - std::atan2(v[1], -v[0]);
                 angle_and_pos[ii] = std::make_pair(angle, ii);
             }
             original.assign(nb[cell].begin(), nb[cell].end());
--- a/opm/grid/cpgrid/EntityRep.hpp
+++ b/opm/grid/cpgrid/EntityRep.hpp
@@ -71,7 +71,6 @@ namespace Dune
 //#include <opm/core/utility/SparseTable.hpp>
 #include <opm/grid/utility/ErrorMacros.hpp>
 #include <climits>
-//#include <boost/algorithm/minmax_element.hpp>
 #include <vector>
 
 /// The namespace Dune is the main namespace for all Dune code.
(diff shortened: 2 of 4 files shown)


akva2:
jenkins build this please

opm/grid/GridUtilities.cpp
```diff
@@ -119,7 +118,8 @@ namespace Opm
                 const double v[2] = { grid.cell_centroids[2*cell2] - grid.cell_centroids[2*cell],
                                       grid.cell_centroids[2*cell2 + 1] - grid.cell_centroids[2*cell + 1] };
                 // The formula below gives an angle in [0, 2*pi] with the positive x axis.
-                const double angle = boost::math::constants::pi<double>() - std::atan2(v[1], -v[0]);
+                constexpr double pi = std::atan(1)*4.0;
```
blattms: Are your that std::atan is a constexpr according to the standard? Might be a gcc extension.
Why not use the one from cmath?
```
#include<cmath>
#include<iostream>
#include<limits>
#include<iomanip>

int main()
{
    constexpr double pi =M_PI;
    std::cout<<std::setprecision(58)
             <<pi<<std::endl;
}
```
works with g++-8 at least or putting the numbers in

bska: > Why not use the one from cmath?
> ```C++
> constexpr double pi =M_PI;
> ```

That's a POSIX extension.  There's no such thing as `M_PI` in standard C++.  There is hope though.  In C++20 we get
```C++
#include <numbers>
std::numbers::pi;
```

atgeirr: I agree that until std::atan is constexpr (my google-fu did not turn up anything solid on that) we should avoid it. The following program fails:
```
#include <cmath>
#include <iostream>

constexpr double pi = std::atan(1)*4.0;

int main()
{
    std::cout << pi << std::endl;
}
```
with the error:
```
pi.cpp:4:18: error: constexpr variable 'pi' must be initialized by a constant expression
constexpr double pi = std::atan(1)*4.0;
                 ^    ~~~~~~~~~~~~~~~~
pi.cpp:4:23: note: non-constexpr function 'atan<int>' cannot be used in a constant expression
constexpr double pi = std::atan(1)*4.0;
                      ^
```
The error message also points out something that worries me: `atan<int>`? I always use 1.0 rather that 1 to avoid nasty surprises of that sort, but that caught me off guard.

atgeirr: > That's a POSIX extension. There's no such thing as `M_PI` in standard C++.

I say we use it anyway. We should be able to assume POSIX.

But perhaps a better solution is to keep using the boost constant until C++20. I assume it is a header-only thing.

bska: > I agree that until `std::atan` is constexpr […] we should avoid it.

It's very unlikely that the any of the trigonometric functions will ever be `constexpr`.  They definitely are not today&mdash;mostly due to associate requirements on side effects such as setting `errno` in failure conditions&mdash;nor are there any proposals to make them `constexpr`, at least that I'm aware of.  As @blattms correctly [remarks](https://github.com/OPM/opm-grid/pull/446#discussion_r382582769), the `constexpr` `std::atan` is a non-conforming GCC extension.

> But perhaps a better solution is to keep using the boost constant until C++20. I assume it is a header-only thing.

This becomes a cost/benefit value judgement.  The cost of using Boost is non-zero, but so is the benefit.  Another option might be to just have a simple function that calculates the value once and stores it in a function-level static varible
```C++
namespace {
    double pi()
    {
        static const auto value = 4.0 * std::atan2(1.0, 1.0); // or = std::acos(-1.0) or ...
        return value;
    }
}
```
or even just
```C++
namespace {
    double pi()
    {
        return 3.14159265358979323846264338327950288; // or some other approximation
    }
}
```

blattms:
Yep, I like the latter. I guess boost is a bit fancier but boils down to the same for double.

opm/grid/GridUtilities.cpp
akva2: done as the latter suggestion.

akva2:
jenkins build this please

PR #448, opened 2020-02-27
log loadbalancing info on rank 0

fgfuchs:
load balancing info is collected on rank 0 and logged with OpmLog

also removed unused variable my_num

### DIFF
--- a/opm/grid/cpgrid/CpGrid.cpp
+++ b/opm/grid/cpgrid/CpGrid.cpp
@@ -152,7 +152,6 @@ CpGrid::scatterGrid(EdgeWeightMethod method, const std::vector<cpgrid::OpmWellTy
 
     if (cc.size() > 1)
     {
-        int my_num=cc.rank();
 #ifdef HAVE_ZOLTAN
         auto part_and_wells =
             cpgrid::zoltanGraphPartitionGridOnRoot(*this, wells, transmissibilities, cc, method, 0);
@@ -247,12 +246,23 @@ CpGrid::scatterGrid(EdgeWeightMethod method, const std::vector<cpgrid::OpmWellTy
 
         distributed_data_->distributeGlobalGrid(*this,*this->current_view_data_, cell_part);
         int num_cells = distributed_data_->cell_to_face_.size();
-        std::ostringstream message;
-        message << "After loadbalancing process " << my_num << " has " << num_cells << " cells.";
-        if (num_cells == 0) {
-            throw std::runtime_error(message.str() + " Aborting.");
-        } else {
-            std::cout << message.str() << "\n";
+
+        std::vector<int> cc_num_cells(cc.size(), 0);
+        cc_num_cells[cc.rank()] = num_cells;
+        cc.sum(cc_num_cells.data(),cc.size());
+
+        if (cc.rank() == 0) {
+            std::string str = "\nAfter loadbalancing on " + std::to_string(cc.size()) + " processes we have\nrank: num cells\n----------------\n";
+            for (int i = 0; i < cc.size(); ++i) {
+                str += std::to_string(i) + ": " + std::to_string(cc_num_cells[i]) + "\n";
+            }
+            str += "----------------\n";
+            Opm::OpmLog::info(str);
+        }
+        for (const auto& nc : cc_num_cells) {
+            if (nc == 0) {
+                throw std::runtime_error("At least one process has zero cells. Aborting.");
+            }
         }
 
         current_view_data_ = distributed_data_.get();


akva2:
jenkins build this please

blattms:
Interesting approach. Do we need the sum?
I would have used `cc.gather(&num_cells, num_cells.data(), 1, 0);`

BTW: Is it me or is github not correctly working, currently?
I had problems logging in (HTTP 500) and cannot comment on the diff online.

bska:
> BTW: Is it me or is github not correctly working, currently?

There does seem to be something that's not quite right.  The GitHub site [status page](
https://www.githubstatus.com/) also suggests similar experience.

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -248,7 +249,18 @@ CpGrid::scatterGrid(EdgeWeightMethod method, const std::vector<cpgrid::OpmWellTy
         distributed_data_->distributeGlobalGrid(*this,*this->current_view_data_, cell_part);
         int num_cells = distributed_data_->cell_to_face_.size();
         std::ostringstream message;
-        message << "After loadbalancing process " << my_num << " has " << num_cells << " cells.";
+
+        cc_num_cells[cc.rank()] = num_cells;
+        cc.sum(&cc_num_cells[0],cc.size());
+
+        if (cc.rank()==0) {
+            std::string str = "\nAfter loadbalancing on " + std::to_string(cc.size()) + " proccesses we have\nrank: num cells\n----------------\n";
```
bska: Tiny nit: You should remove one of the 'c's in "proccesses".

bska:
I'm just going to rerun the build test.  The earlier test, though successful, did not propagate the successful status back to the PR.

bska:
jenkins build this please

fgfuchs:
thanks for the code review. I have fixed the typo. I will leave it up to Atgeirr to decide if I should change more.

atgeirr:
> Interesting approach. Do we need the sum? I would have used `cc.gather(&num_cells, num_cells.data(), 1, 0);`

I think you are right, in that gather() is perhaps more like how it is usually done. It is interesting though that sum() is simpler to understand (do not need the 1, 0 arguments). Perhaps "single item from each -> array on root/all" overloads of gather()/allgather() could be added to the collective communication class for cases like this.

As is I think that using sum() is fine here.

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -248,7 +249,18 @@ CpGrid::scatterGrid(EdgeWeightMethod method, const std::vector<cpgrid::OpmWellTy
         distributed_data_->distributeGlobalGrid(*this,*this->current_view_data_, cell_part);
         int num_cells = distributed_data_->cell_to_face_.size();
         std::ostringstream message;
-        message << "After loadbalancing process " << my_num << " has " << num_cells << " cells.";
+
+        cc_num_cells[cc.rank()] = num_cells;
+        cc.sum(&cc_num_cells[0],cc.size());
```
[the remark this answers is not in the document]
atgeirr: I am fine with sum(), but please use cc_num_cells.data().

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -248,7 +249,18 @@ CpGrid::scatterGrid(EdgeWeightMethod method, const std::vector<cpgrid::OpmWellTy
...
+
+        if (cc.rank()==0) {
```
atgeirr: Nitpick: space around binary operator. Also on line 261.

atgeirr:
Ok, squash the commits and we are done!

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -248,7 +249,18 @@ CpGrid::scatterGrid(EdgeWeightMethod method, const std::vector<cpgrid::OpmWellTy
...
+        if (cc.rank() == 0) {
+            std::string str = "\nAfter loadbalancing on " + std::to_string(cc.size()) + " processes we have\nrank: num cells\n----------------\n";
+            for (int i = 0; i < cc.size(); ++i) {
+                str += std::to_string(i) + ": " + std::to_string(cc_num_cells[i]) + "\n";
+            }
+            str += "----------------\n";
+            Opm::OpmLog::info(str);
+        }
         if (num_cells == 0) {
             throw std::runtime_error(message.str() + " Aborting.");
```
akva2: this + else branch is now pretty useless ?

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -248,7 +249,18 @@ CpGrid::scatterGrid(EdgeWeightMethod method, const std::vector<cpgrid::OpmWellTy
         distributed_data_->distributeGlobalGrid(*this,*this->current_view_data_, cell_part);
         int num_cells = distributed_data_->cell_to_face_.size();
         std::ostringstream message;
```
akva2: now unused.

fgfuchs: your are right, I will write a fix

fgfuchs:
jenkins build this please

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -247,12 +248,22 @@ CpGrid::scatterGrid(EdgeWeightMethod method, const std::vector<cpgrid::OpmWellTy
...
+
+        if (cc.rank() == 0) {
+            std::string str = "\nAfter loadbalancing on " + std::to_string(cc.size()) + " processes we have\nrank: num cells\n----------------\n";
+            for (int i = 0; i < cc.size(); ++i) {
+                str += std::to_string(i) + ": " + std::to_string(cc_num_cells[i]) + "\n";
+            }
+            str += "----------------\n";
+            Opm::OpmLog::info(str);
+        }
+        for(const auto& cc_num_cells: nc) {
```
atgeirr: You must put nc before cc_num_cells.

Also some nitpicks: space after for, space around colon, and move the definition of cc_num_cells to line 251 instead of 100 lines above.

fgfuchs:
I think that formatting should be handled automatically via some hooks.

fgfuchs:
jenkins build this please

opm/grid/cpgrid/CpGrid.cpp
fgfuchs: done

bska:
> jenkins build this please

It appears you (currently?) don't have the ability to launch Jenkins build tests.  Don't worry.  I'll do it for you.

bska:
jenkins build this please

fgfuchs:
Thank you

atgeirr:
> It is an all-to-all communication versus an all-to-one communication. That means that it sends twice as many message as gather (that might depend on the implementation of course). In addition, each message has size `P`, the number of processes, versus size 1 for the gather.

Point taken, I am convinced. Let's use gather().

fgfuchs:
> I don't.
> It is an all-to-all communication versus an all-to-one communication. That means that it sends twice as many message as gather (that might depend on the implementation of course). In addition, each message has size `P`, the number of processes, versus size 1 for the gather.

At the moment, I am using sum to throw an exception on all processes, if one process has zero cells.

blattms:
Was that throw there before and needed?

Somehow I thought the bug with one process having no cells has been fixed somewhen. Am I mistaking?

atgeirr:
Yes, the check and throw was already there. I think adding it was the fix, it used to just crash.

blattms:
ok. then let us merge like it is, but please open issue, that we should retest
whether there is still a problem with zero elements on one process and if not
remove the sum

atgeirr:
> we should retest whether there is still a problem with zero elements on one process

I think this was observed with MPI but not Zoltan, because then cartesian division of the domain is used. Of course, running a 3-cell case on 4 processes should also trigger it, Zoltan or not...

atgeirr:
Issue made as requested, I'll then merge.

PR #452, opened 2020-03-03
Add missing loadBalance method taking data handle and egde weights.

blattms:
The corresponding version without the data handle was there but somehow the one with the data handle was missing. That one is supposed to also distribute data attached to the grid during load balancing and will be needed and used in the future. This PR adds it.

### DIFF
--- a/opm/grid/CpGrid.hpp
+++ b/opm/grid/CpGrid.hpp
@@ -686,6 +686,33 @@ namespace Dune
             return ret;
         }
 
+        /// \brief Distributes this grid over the available nodes in a distributed machine
+        ///
+        /// This will construct the corresponding graph to the grid and use the transmissibilities
+        /// specified to calculate the  weights associated with its edges. The graph will be passed
+        ///  to the load balancer.
+        /// \param data A data handle describing how to distribute attached data.
+        /// \param method The edge-weighting method to be used on the Zoltan partitioner.
+        /// \param wells The information about all possible wells. If null then
+        ///            the wells will be neglected. Otherwise the wells will be
+        ///            used to make sure that all the possible completion cells
+        ///            of each well are stored on one process. This is done by
+        ///            adding an edge with a very high edge weight for all
+        ///            possible pairs of cells in the completion set of a well.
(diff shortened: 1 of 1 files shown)


blattms:
jenkins build this please

opm/grid/CpGrid.hpp
```diff
@@ -686,6 +686,33 @@ namespace Dune
...
+        template<class DataHandle>
+        std::pair<bool, std::unordered_set<std::string> >
+        loadBalance(DataHandle& data, EdgeWeightMethod method,
+                    const std::vector<cpgrid::OpmWellType> * wells,
+                    const double* transmissibilities = nullptr,
+                    int overlapLayers=1)
+        {
+            const auto& ret = scatterGrid(method, wells, transmissibilities, overlapLayers);
+            scatterData(data);
+            return ret;
```
bska: Just out of curiosity: Does `scatterData()` depend on `scatterGrid()` being called first?  If not, I think this could be shortened to
```C++
scatterData(data);
return scatterGrid(/* … */);
```
On the other hand, if `scatterData()` **does** depend on `scatterGrid()` being called first, I think you should write the body slightly differently:
```C++
auto ret = scatterGrid(/* … */);
scatterData(data);
return ret;
```
We're targeting C++17 meaning we have mandatory copy elision of return values.  There is no additional cost to having `ret` as an actual object rather than a "reference-to-const" here&mdash;at least if `scatterGrid()` has the same return type as `loadBalance()`.

blattms: It does depend on it.

I tried to read this up (RVO and copy elision often confuse me). To me it seems that named return value optimization (NRVO) is still not guaranteed in C++17. There is a [proposal for C++21](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/p2025r0.html). But maybe it is wise to bet on compilers being smart. Will change.

blattms: Did that anywhere else, anyway. Hence more consistent now. Thanks.

blattms:
jenkins build this please

blattms:
Shameless self-merge since this cannot break anything as it not yet used by master.

PR #456, opened 2020-03-09
fix build without dune-istl

### LINKED ISSUE #455: dune-istl is required rather than a suggestion
In the dune module file states that this module may optionally use `dune-istl`, however, when it comes to compilation the project fails because there is a bunch of places where dune istl objects are used regartheless if the `dune-istl` was found or not.

akva2:
Closes #455

### DIFF
--- a/opm/grid/common/GridPartitioning.cpp
+++ b/opm/grid/common/GridPartitioning.cpp
@@ -37,7 +37,9 @@
 #if HAVE_CONFIG_H
 #include "config.h"
 #endif
+#if HAVE_DUNE_ISTL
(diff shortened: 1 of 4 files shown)


akva2:
jenkins build this please

opm/grid/common/GridPartitioning.cpp
```diff
@@ -242,6 +244,7 @@ namespace Dune
         }
     }
 
+#if HAVE_DUNE_IST
```
blattms: typo

akva2:
can you just do it like you want it then?

blattms:
Sure, will do.

PR #459, opened 2020-03-10
Use GlobalIdSet also as local id set to make it persistent

### LINKED ISSUE #454: local ids not persistent across loadbalancing
The way the locel id set is currently implement, the ids get computed by based on the number of entities there are for each codim. Since this number changes when loadbalancing, the ids change too.
This is a violation of the DUNE grid interface that requires ids to be consistent across grid changes. It will prevent people from using their usual approach of distributing data.

This was also the reason for the differing well curves in OPM/opm-simulators#2406 where we used one unordered map with local ids as keys to store the values to send and the same map for receiving them.

blattms:
during grid modifications. This required for being a dune grid. The additional trait of being globally unique does not hurt us and this is the least invasive change to accomplish persistence.

Closes #454

depends on OPM/opm-simulators#2458

### DIFF
--- a/opm/grid/CpGrid.hpp
+++ b/opm/grid/CpGrid.hpp
@@ -172,7 +172,7 @@ namespace Dune
         /// \brief The type of the global id set.
         typedef cpgrid::GlobalIdSet GlobalIdSet;
         /// \brief The type of the local id set.
-        typedef cpgrid::GlobalIdSet LocalIdSet;
+        typedef GlobalIdSet LocalIdSet;
 
         /// \brief The type of the collective communication.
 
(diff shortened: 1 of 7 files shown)


blattms:
jenkins build this please

blattms:
jenkins build this opm-simulators=2458 please

blattms:
Somehow jenkins run did not pickup the force-pushed changes of 2458. Will rerun once that is merged.

blattms:
jenkins build this please

blattms:
jenkins run norne norne_parallel please

blattms:
norne looks good. Merging.

PR #462, opened 2020-03-19
quell unused parameter warnings without MPI

akva2:
Have to copy the macros as the OPM ones are defined in opm-material.


akva2:
jenkins build this serial no downstreams please

bska:
Is there a reason you decided not to use the C++17 `[[maybe_unused]]` attribute here?

akva2:
not really, i just did like we do in the rest of opm. i can certainly do that instead.

akva2:
jenkins build this serial no downstreams please

bska:
> i just did like we do in the rest of opm. i can certainly do that instead.

Thanks for the updates.  I think we should gradually move away from the macro-based system that only works on a subset of compilers to using standard facilities where possible.  Certainly for new code the attributes are preferable in my opinion.

Note: I'm not advocating that we change all existing macros in one sweeping commit across the code base, but that we move to using the applicable attributes when making other changes in relevant locations.

bska:
> standardized attribute seem cool.

They absolutely are.  Many are C++14 and later, some require C++17 and we'll get a few new ones in C++20.  [Cppreference](https://en.cppreference.com/w/cpp/language/attributes) has a decent overview.

PR #471, opened 2020-06-04
Removes unsued file PinchProcesser.hpp

blattms:
as it is not used anywhere. It seems like pinch-outs are handled by `MinpvProcessor` these days.


blattms:
jenkins build this please

atgeirr:
I think this was abandoned, but for some time it was unclear if we wanted to pick up the work again. Thanks for cleaning up.

Issue #472, opened 2020-06-05
Question about conditionals in MinpvProcessor::process


blattms:
I am working on incorporating `PINCH NOGAP` functionality and am currently looking at [MinpvProcessor::process](/OPM/opm-grid/blob/f7df1bd22eabdf7fa6696113e4c5a73965747143/opm/grid/MinpvProcessor.hpp#L76). If we look at [MinpvProcessor.hpp:138](/OPM/opm-grid/blob/f7df1bd22eabdf7fa6696113e4c5a73965747143/opm/grid/MinpvProcessor.hpp#L138), we read:
```C++
// bypass inactive cells with thickness less then the tolerance
while ( ((actnum.empty() || !actnum[c_below]) && (thickness[c_below] <= z_tolerance))  ){
....
```
If actnum is nonempty, then this code does what the comment says. But if it is empty, then my impression always was that all cells are active. If my impression is correct, then the code actually bypasses active cells with thickness below the threshold for empty `actnum`. The correct code would be:
```C+
while ( !actnum.empty() && !actnum[c_below] && (thickness[c_below] <= z_tolerance)  ){
```

Hence the question is: Is the comment correct and we should bypass only inactive cells with thickness below threshold? 

Or is it that `Opm::EclGrid::getACTNUM()` never returns an empty vector?

The same holds for the first part of the conditions in [MinpvProcessor.hpp:162](/OPM/opm-grid/blob/f7df1bd22eabdf7fa6696113e4c5a73965747143/opm/grid/MinpvProcessor.hpp#L162) and [MinpvProcessor.hpp:171](/OPM/opm-grid/blob/f7df1bd22eabdf7fa6696113e4c5a73965747143/opm/grid/MinpvProcessor.hpp#L171).

blattms:
closing as it is stale

PR #473, opened 2020-06-05
[WIP] Added a variant for MINPV testcase 2 with all cells active.

blattms:
This test should fail currently but should show my point outlined in #472 .

Like I said there actnum being empty should never happen in production.


blattms:
jenkins build this please

blattms:
This is a bit academic. Closing as the problem is only there for the test code.

PR #474, opened 2020-06-08
[bugfix] Do not MPI in processEclipseFormat.cpp

blattms:
Scatter is of course wrong and might result in segmentation faults. We are already broadcasting the stuff in the outer methods and scatter will never be called on all processes (this might have very strange side effects). Now we consistently only process eclipse on one rank and broadcast afterwards. For `CpGrid::createCartesian`, we did not do that before this PR. This PR removes MPI from processEclipseFormat.cpp

Thanks a lot to @akva2 for finding this.

### DIFF
--- a/opm/grid/cpgrid/processEclipseFormat.cpp
+++ b/opm/grid/cpgrid/processEclipseFormat.cpp
@@ -250,9 +250,7 @@ namespace cpgrid
     {
         if( ccobj_.rank() != 0 )
         {
-            // We do not store any grid information
-            ccobj_.scatter(logical_cartesian_size_.data(), logical_cartesian_size_.data(), 3, 0);
-            return;
+            OPM_THROW(std::logic_error, "Processing  eclipse file only allowed on rank 0");
         }
         // Process.
 #ifdef VERBOSE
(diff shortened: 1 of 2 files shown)


blattms:
jenkins build this please

blattms:
jenkins build this please

opm/grid/cpgrid/processEclipseFormat.cpp
```diff
@@ -251,7 +251,7 @@ namespace cpgrid
         if( ccobj_.rank() != 0 )
         {
             // We do not store any grid information
-            ccobj_.scatter(logical_cartesian_size_.data(), logical_cartesian_size_.data(), 3, 0);
+            ccobj_.broadcast(logical_cartesian_size_.data(), 3, 0);
```
akva2, line 254: this should go in the return block in the processEclipseFormat overload that calls this one.

blattms: Turns out that this was indeed quite a mess. Thanks for the hint

akva2:
jenkins build this please

blattms:
I did address the comments and force pushed. The MPI bug should be fixed with this.

But I guess we need to fix the SintefLegacyFormat that still uses the old approach (reading on all processes)

PR #475, opened 2020-06-08
Read and write SINTEF legacy format only on rank zero.

blattms:
This needed for consistency. Otherwise a lot of space is wasted in parallel and the expectations made in the rest of the code (full grid only on rank zero) is violated.

Somebody who knows that format needs to take a look and maybe add a test.


akva2:
i'll take care of adding a test later, when i get hold of a grid. in the mean time, let's not let this keep the  broadcast fix back. this looks to be following the pattern for the deck based io.

akva2:
jenkins build this please

PR #477, opened 2020-06-15
Added PINCH NOGAP functionality with small test.

blattms:


blattms:
I am a bit uncomfortable with this PR and would assume that it needs further testing. If I assume that that the difference in transmissibilities for `PINCH GAP` compared to the commercial simulator are expected, then this PR makes sense as it has the same transmissibilities as before and results in the same neighboring cells as the commercial simulator.

See also my questions in #472.

blattms:
There tests in OPM/opm-tests#336 that illustrate the differences in transmissibility.

blattms:
jenkins build this please

blattms:
jenkins build this please

blattms:
E.g. the differences for [FIVE_PINCH_GAP.DATA](/OPM/opm-tests/blob/6f86aaa706f83975b029fc29affbbab077d75f02/pinch/FIVE_PINCH_GAP.DATA) are:
- flow
```
 'TRANZ   '           4 'REAL'
   0.32727274E+03   0.36000000E+03   0.36000000E+03   0.00000000E+00
```
- commercial
```
 'TRANZ   '           4 'REAL'
   0.36000000E+03   0.36000000E+03   0.36000000E+03   0.00000000E+00
```
I guess this might be due to changing the volume of the upper cell for neglecting its neighbor. But somebody more knowledgeable should take another look. Anyway, removing the WIP now.

Would be cool if somebody would use the added tests in opm-tests for regression tests. I will not find time soon.

alfbr:
> this PR makes sense as it has the same transmissibilities as before

Then I believe the NOGAP option is implemented correctly in this PR. Reproducing transmissibilities of another simulator may or may not make sense, but that is a discusssion for another PR. From my end, it would be great to get the NOGAP functionality in now. Please consider removing WIP.

blattms:
Will merge this on Wednesday if there are no durther comments.

opm/grid/MinpvProcessor.hpp
```diff
@@ -159,28 +164,30 @@ namespace Opm
                             int c_above = ii + dims_[0] * (jj + dims_[1] * (kk - 1));
 
                             // Bypass inactive cells with thickness below tolerance and active cells with volume below minpv
-                            if (((actnum.empty() || !actnum[c_above]) && thickness[c_above] < z_tolerance) || ((actnum.empty() || actnum[c_above]) && pv[c_above] < minpvv[c_above]) ) {
+                            if (((actnum.empty() || !actnum[c_above]) && thickness[c_above] < z_tolerance) || ((actnum.empty() || actnum[c_above]) && pv[c_above] < minpvv[c_above]
```
atgeirr, line 167: I think the logic is correct here, and can be merged. However, it was already hard to read, and became even harder now. I suggest making a few bool variables (active_above, thin_above, smallvolume_above for example) to improve readability and reduce the possibility of mistakes. I think the way things are written right now is a bit more complicated than necessary, it should not be necesssary to evaluate the minpvv condition when pinchNOGAP is true.

There is one minor thing that puzzles me: we seem to be inconsistent about the tests, using `<` and then later `>` rather than `>=` for the converse. That would automatically be addressed by using bool variables, and possibly cause a change in behaviour, if we were unlucky enough to depend on this (unlikely though).

Nitpick: line 190 indentation change?

blattms: Yeah, that brings me straight to #472. Either the conditionals are wrong or the comment.
`(actnum.empty() || !actnum[c_above])` means to me "(all cells are active due to no ACTNUM) or cell above is inactive)".
I will do your suggestion as this makes the problems clearer.

atgeirr: I understand #472 better now. I do not know the correct behaviour, but agree that it looks wrong. (On the other hand, I suspect that this code is never called with an empty actnum.)

blattms: I did break down the logical statements.

Concerning not evaluating minpvv for pinch NOGAP. Currently, this reproduces the test and changing it would make the logic even more involved. If you case then please provide a test showing that this is wrong (I cannot really judge, here).
In that case i am happy to change.

bska:
jenkins build this please

bska:
I'm going to merge this now.  Thanks a lot for your persistence.

PR #484, opened 2020-09-04
Abort for zero cells before scattering grid and allow sparse tables with zero rows

### LINKED ISSUE #482: Parallel running gets stuck with the following error message
When one process reports the following message, the whole parallel simulation got stuck. 
```
OPM/opm-grid/opm/grid/utility/SparseTable.hpp:266] Must have at least one row. Got 0 rows.
```
Not sure what is the most proper way in handling this. Either continuing the simulation with some special treatment, or we should just exit the running. 

Either way, it is not good for the running just getting stuck (silently).

blattms:
We move the printing of the loadbalancing (and the abort for zero cells) before actually distributing the grid. Hence the deadlock described in #482 should not occur any more but all processes should abort.
In addition we change SparseTable to allow for zero rows and rows with zero entries. This is the first step to simulate with partitions containing zero cells. Unfortunately there are more bugs to be resolved before this can work.

Closes #482

### DIFF
--- a/opm/grid/common/ZoltanPartition.hpp
+++ b/opm/grid/common/ZoltanPartition.hpp
@@ -46,10 +46,13 @@ namespace cpgrid
 /// @param edgeWeightMethod The method used to calculate the weights associated
 ///             with the edges of the graph (uniform, transmissibilities, log thereof)
 /// @param root The process number that holds the global grid.
-/// @return A pair consisting of a vector that contains for each local cell of the grid the
+/// @return A tuple consisting of a vector that contains for each local cell of the original grid the
 ///         the number of the process that owns it after repartitioning,
-///         and a set of names of wells that should be defunct in a parallel
-///         simulation.
+///         a set of names of wells that should be defunct in a parallel
+///         simulation, vector containing information for each exported cell (global id
+///         of cell, process id to send to, attribute there), and a vector containing
+///         information for each imported cell (global index, process id that sends, attribute here, local index
+///         here)
 std::tuple<std::vector<int>,std::unordered_set<std::string>,
            std::vector<std::tuple<int,int,char> >,
            std::vector<std::tuple<int,int,char,int> >  >
--- a/opm/grid/cpgrid/CpGrid.cpp
+++ b/opm/grid/cpgrid/CpGrid.cpp
@@ -230,6 +230,63 @@ CpGrid::scatterGrid(EdgeWeightMethod method,
                                importList.end(), compareImport);
         }
 
+        int procsWithZeroCells{};
+
+        if (cc.rank()==0)
+        {
+            // Print some statistics without communication
+            std::vector<int> ownedCells(cc.size(), 0);
+            std::vector<int> overlapCells(cc.size(), 0);
+            for (const auto& entry: exportList)
+            {
+                if(std::get<2>(entry) == AttributeSet::owner)
+                {
+                    ++ownedCells[std::get<1>(entry)];
+                }
+                else
+                {
+                    ++overlapCells[std::get<1>(entry)];
+                }
+            }
+
+            for(const auto& cellsOnProc: ownedCells)
+            {
+                procsWithZeroCells += (cellsOnProc == 0);
+            }
+            std::ostringstream ostr;
+            ostr << "\nLoad balancing distributes " << data_->size(0)
+                 << " active cells on " << cc.size() << " processes as follows:\n";
+            ostr << "  rank   owned cells   overlap cells   total cells\n";
+            ostr << "--------------------------------------------------\n";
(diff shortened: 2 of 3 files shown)


blattms:
jenkins build this please

blattms:
jenkins build this please

blattms:
@GitPaean Please merge if you have tested this.

blattms:
jenkins run norne please

blattms:
jenkins build this please

blattms:
@GitPaean Did find time to test this? I think this should be merged once jenkins is green and you approve

GitPaean:
Sorry. I have been focusing looking at the failure during the Zoltan calls, have not found time to test this. 

I will test this soon, by Monday. 

The other question, have you managed to use PT-Scotch through Zoltan to run OPM simulation. I built Zoltan with PTScotch, and built OPM with Zoltan and PTScotch. And specifying to use `Scotch` in function `zoltanGraphPartitionGridOnRoot`, but I could not manage to make any parallel run, because the rank other than rank 0 claims no cells allocated (SparseTable failure).

blattms:
That this PTScotch was just a guess as it seems like the loadbalancing behaves differently on
different processors. I would need to look into what Zoltan uses as default.

The SparseTable failure means that the are ranks without any cells. With this PR you will get that
as an error message.

GitPaean:
> That this PTScotch was just a guess as it seems like the loadbalancing behaves differently on
different processors. I would need to look into what Zoltan uses as default.

My point is that it looks like the PTScotch did not do its job. All the cells are only on rank 0. It did not distribute cells to other process at all.

GitPaean:
Hi, with this PR, it prints out the load balance information. But it still hangs there when some processes do not get any cells. 

When I forced to terminate the job, it prints the backtrace information. Both features above is very helpful. 

But it did not abort the running.

GitPaean:
Wait a little bit.  It looks like the code is a little different from the one in this PR. 

```
277         cc.sum(procsWithZeroCells);
278         if (procsWithZeroCells == 0 && false) {
279             if (cc.rank()==0)
280             {
281                 OPM_THROW(std::runtime_error, "At least one process has zero cells. Aborting.");
282             }
283             else
284             {
285                 OPM_THROW_NOLOG(std::runtime_error, "At least one process has zero cells. Aborting.");
286             }
287         }
288 

```

Let me check a little bit what is going on. The cluster can not download things directly, so testing PR sometimes is a little bit troublesome.

blattms:
That looks like the other PR.

GitPaean:
> That looks like the other PR.

Yes. There are too many files copied around. I messed up.  Re-building and testing.

GitPaean:
Hi, it still hangs and gets stuck. 

I added a little bit code and it looks like `cc.sum` not updating  `procsWithZeroCells` in all the processes, which is a surprise to me. I thought it should be the behavior of `AllReduce`. 

```c++
277         cc.sum(procsWithZeroCells);
278 
279         if (cc.rank() == 2) {
280             std::cout << " procsWithZeroCells in rank " << cc.rank() << " == " << procsWithZeroCells << std::endl;
281         }
282 
283         if (cc.rank() == 9) {
284             std::cout << " procsWithZeroCells in rank " << cc.rank() << " == " << procsWithZeroCells << std::endl;
285         }
286 
287         if (cc.rank() == 0) {
288             std::cout << " procsWithZeroCells in rank " << cc.rank() << " == " << procsWithZeroCells << std::endl;
289         }
```
The output is 
```
 1821  procsWithZeroCells in rank 2 == 0
1822  procsWithZeroCells in rank 9 == 0
1958  procsWithZeroCells in rank 0 == 3
```

blattms:
Thanks for testing. there is indeed a bug.

blattms:
:flushed: Embarrassing. One should assign the result, of course.  Should work now

blattms:
jenkins build this please

bska (approved):
I think this looks good.  Do you want to run another test, @GitPaean, before we merge this?

GitPaean:
It works fine for its purpose.  I think it is okay to merge it. 

I only have one question here,  should we see the following message somewhere? 
```
                 OPM_THROW_NOLOG(std::runtime_error, "At least one process has zero cells. Aborting."); 
```

blattms:
Probably not, as we hopefully make sure that only rank 0 prints messages when catching exceptions (i it is not catched you will see messages on all processes about termination). I just throw on every process to be sure to prevent any deadlocks. OPM_THROW_LOG is used to make sure ranks other than 0 do not print any messages to the logs or console

blattms:
@GitPaean Thanks a lot for testing. Highly appreciated.

GitPaean:
Thanks for the efforts to make the parallel running easier and more informative.

PR #486, opened 2020-09-17
Fixing zoltan test

GitPaean:
```
BOOST_REQUIRE(grid.comm()==MPI_COMM_SELF);
```
This line never works but the test passes anyway. reported by @kjetilly . 

```
11: Test command: /usr/bin/mpiexec "-n" "1" "/home/kaib/OPM-devel/debug/opm-grid-build/bin/zoltan_test"
11: Test timeout computed to be: 1500
11: Invalid MIT-MAGIC-COOKIE-1 keyRunning 1 test case...
11: /home/kaib/OPM-devel/debug/opm-grid/tests/cpgrid/zoltan_test.cpp(120): fatal error: in "zoltan": critical check grid.comm()==(static_cast<MPI_Comm> (static_cast<void *> (&(ompi_mpi_comm_self)))) has failed
11: 
11: *** 1 failure is detected in the test module "ZoltanTests"
11: 
1/1 Test #11: zoltan ...........................   Passed    0.42 sec
```

@akva2 , not sure if this line related to what you mention that creating Zoltan with `MPI_COMM_SELF`, but in the code `zz = Zoltan_Create(MPI_COMM_WORLD);	` anyway. 

There might be something messy in this example. Maybe during discussion this PR, we can clean up the setup of the Zoltan a little bit.  

`MPI_Barrier` is because when I run this test in some cluster with two processes, one process is still creating the grid, and the other one is already beginning to partition. So `segmentation fault` and the running gets stuck.

### DIFF
--- a/tests/cpgrid/zoltan_test.cpp
+++ b/tests/cpgrid/zoltan_test.cpp
@@ -117,7 +117,7 @@ BOOST_AUTO_TEST_CASE(zoltan)
 
         Dune::cpgrid::setCpGridZoltanGraphFunctions(zz, grid);
 
-        BOOST_REQUIRE(grid.comm()==MPI_COMM_SELF);
+        BOOST_REQUIRE(grid.comm()==MPI_COMM_WORLD);
 
         //ZOLTAN_TRACE_ENTER(zz, yo);
         rc = Zoltan_LB_Partition(zz, /* input (all remaining fields are output) */
--- a/tests/cpgrid/zoltan_test.cpp
+++ b/tests/cpgrid/zoltan_test.cpp
@@ -115,6 +115,8 @@ BOOST_AUTO_TEST_CASE(zoltan)
 #endif
             grid.createCartesian(dims, size);
 
+        MPI_Barrier(MPI_COMM_WORLD);
+
         Dune::cpgrid::setCpGridZoltanGraphFunctions(zz, grid);
 
         BOOST_REQUIRE(grid.comm()==MPI_COMM_WORLD);


GitPaean:
jenkins build this please

GitPaean:
The unanswered question for me here is that even there is a failure, the test still passes. 

I am not that familiar with boost_test.

bska:
> The unanswered question for me here is that even there is a failure, the test still passes.

That's because the [`main`](https://github.com/OPM/opm-grid/blob/1608ab073b3a7aee60d10f78b25d6a99679999d3/tests/cpgrid/zoltan_test.cpp#L177) function unconditionally returns success.

GitPaean:
> That's because the main function unconditionally returns success.

I am guessing it is using `BOOST_TEST_NO_MAIN` because the parallel or Dune.  Any suggestion to make this test a proper test?

Thanks.

blattms:
Will provide a patch.
Is this ever run with more than 1 process by ctest?

akva2:
not currently. it's easily added though if of interest (see bottom of CMakeLists.txt where we do the distribution test).

blattms:
that explains why the error was never seen.

@GitPaean I posted a PR to your branch that uses fixtures to initialize MPI. Hopefully that fixes issues. When I wrote this test I somehow could not make that work, but now it seems to do. Maybe that was due to an issue with former boost versions.

tests/cpgrid/zoltan_test.cpp
```diff
@@ -115,9 +115,11 @@ BOOST_AUTO_TEST_CASE(zoltan)
 #endif
             grid.createCartesian(dims, size);
 
+        MPI_Barrier(MPI_COMM_WORLD);
```
blattms, line 118: Doesn't hurt, but it is needed?

blattms (approved):
Looks good.

tests/cpgrid/zoltan_test.cpp
GitPaean: It is needed. As mentioned above. At the same time, it was in the middle of us trying to figure out the Zoltan partition problem, it killed a little while for me to figure out. 

```MPI_Barrier is because when I run this test in some cluster with two processes, one process is still creating the grid, and the other one is already beginning to partition. So segmentation fault and the running gets stuck.```

**Comment by GitPaean** [comment]
jenkins build this please

File: tests/cpgrid/zoltan_test.cpp [hunk 2]
```diff
@@ -53,6 +52,35 @@ void MPI_err_handler(MPI_Comm *, int *err_code, ...){
...
+#endif
+    }
+    ~MPIFixture()
+    {
+#if HAVE_MPI
+        MPI_Finalize();
+#endif
+    }
+    Dune::MPIHelper* helper;
+    MPI_Errhandler handler;
```
bska, line 79: @blattms: This data member (`handler`) should **probably** be guarded by `#if HAVE_MPI`.  The post-merge sequential build [failed](https://ci.opm-project.org/job/opm-grid/81/console) with the diagnostic message
```
/var/lib/jenkins/post-builder/workspace/opm-grid/tests/cpgrid/zoltan_test.cpp:79:5: error: 'MPI_Errhandler' does not name a type; did you mean 'MPIError'?
     MPI_Errhandler handler;
     ^~~~~~~~~~~~~~
     MPIError
```

blattms: Thanks, I could figure jenkins description of the problem

PR #487, opened 2020-09-22
Refactored how loadbalance exports information about parallel wells.

blattms:
Previously, we exported an unordered map containing all names of wells that are not present in the local part of the grid.

As we envision to have wells that are distributed across multiple processors, this information does not seem to be enough. We need
to be able to set up communication for each well. To do this we need to find out who handles perforations of each well.

We now export a full list of well name together with a boolean indicating whether it perforates local cells (vector of pair of string and bool).

Needs a downstream PR: OPM/opm-simulators#2802

### DIFF
--- a/opm/grid/common/WellConnections.cpp
+++ b/opm/grid/common/WellConnections.cpp
@@ -310,14 +310,13 @@ postProcessPartitioningForWells(std::vector<int>& parts,
 
 }
 
-std::unordered_set<std::string>
-computeDefunctWellNames(const std::vector<std::vector<int> >& wells_on_proc,
-                        const std::vector<OpmWellType>& wells,
-                        const CollectiveCommunication<MPI_Comm>& cc,
-                        int root)
+std::vector<std::pair<std::string,bool>>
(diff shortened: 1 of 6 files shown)


blattms:
jenkins build this opm-simulators=2802 please

blattms:
BTW: Not sure whether we should merge this as there might be changes in the further work towards distributed wells. But I need to test a system where failing tests are real failures (jenkins currently for me).

opm/grid/common/WellConnections.cpp
```diff
@@ -398,24 +397,33 @@ computeDefunctWellNames(const std::vector<std::vector<int> >& wells_on_proc,
         }
     }
 
-    // Compute defunct wells in parallel run.
-    std::vector<int> defunct_wells(wells.size(), true);
+    // Compute wells active/inactive in parallel run.
+    // boolean indicates whether the well perforates local cells
+    std::vector<std::pair<std::string,bool>> parallel_wells;
```
atgeirr, line 402: This type is used in many places, perhaps a type alias (`WellsPresent` for example) would be good?

PR #488, opened 2020-09-22
Implementation of running Zoltan in Serial

kjetilly:
On several clusters, we have experienced issues with running Zoltan for partitioning graphs of 3D Cartesian grids. This does not seem to be an isssue of OPM, but rather an issue of Zoltan, since one can reproduce the issue with a standalone example.

 A simple workaround is to  run Zoltan in serial. This pull request enables running Zoltan in serial, and fixes several observed crashes (Segmentation faults) on HPC clusters.

### DIFF
--- a/opm/grid/common/ZoltanPartition.cpp
+++ b/opm/grid/common/ZoltanPartition.cpp
@@ -24,6 +24,7 @@
 #include <opm/grid/utility/OpmParserIncludes.hpp>
 #include <opm/grid/cpgrid/CpGridData.hpp>
 #include <opm/grid/cpgrid/Entity.hpp>
+#include <algorithm>
 
 #if defined(HAVE_ZOLTAN) && defined(HAVE_MPI)
 namespace Dune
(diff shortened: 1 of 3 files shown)


kjetilly:
> I guess you have reported the problems at [Trilinos's bugtracker](https://github.com/trilinos/Trilinos/issues?q=is%3Aopen+is%3Aissue+label%3A%22pkg%3A+Zoltan2%22). If not, please do so and open a bug report here (with information about which cluster and which number of processes) also to help tracking, if there is any progress.

Thank you for your reply. 

[I have submitted an issue to the Zoltan repository now, with the issue id 8087](https://github.com/trilinos/Trilinos/issues/8087)

kddevin:
Even with the nice test code and graph from @kjetilly, I am unable to reproduce the Zoltan issue above.  I am hopeful that @kjetilly can provide more details.
See details https://github.com/trilinos/Trilinos/issues/8087

GitPaean:
@kddevin , can you tell us a little more about the what types of cluster system you are testing with.

We did not see this when we use it on workstations. It also works fine with a 4-CPU (72 core) cluster.  On a supercomputer, with the same executable and same number of the processes, it works fine when we just run it directly with `mpirun -np`  on the login/compiling node, but when we submit it as a job (bsub) to the computing node, the segmentation fault shows up.

blattms:
> On a supercomputer, with the same executable and same number of the processes,
> it works fine when we just run it directly with `mpirun -np`  on the 
> login/compiling node, but when we submit it as a job (bsub) to the computing
>  node, the segmentation fault shows up. 

Now that is very useful information and I think you should ask the sysadmins
what the difference could be. Maybe you are using different MPI versions in
the two cases?

On some of these supercomputers the systems of the login node is different
from the compute nodes and you need to cross-compile correctly. (The executable
cannot be run on both, but just one). Please consult the system documentation
or post a link. I have done that many years ago and could be of help.

Information about the system is always useful.

kjetilly:
> 
>On a supercomputer, with the same executable and same number of the processes, it works fine when we just run it directly with `mpirun -np` on the login/compiling node, but when we submit it as a job (bsub) to the computing node, the segmentation fault shows up.
> Now that is very useful information and I think you should ask the sysadmins what the difference could be. Maybe you are using different MPI versions in the two cases? 

When you run MPI on the login-node you will only see one node, and never be subject to the interconnect.  This seems to fit well with the observation that it runs fine on single computers or home-made clusters without the same setup as these larger clusters.

>On some of these supercomputers the systems of the login node is different
>from the compute nodes and you need to cross-compile correctly. (The executable
>cannot be run on both, but just one). Please consult the system documentation
>or post a link. I have done that many years ago and could be of help.
>
>Information about the system is always useful.

This does not seem to be the case here. We have tested this on three different, independent clusters. I doubt this is a cluster problem. 

I have submitted Valgrind outputs and lots of other stuff to the Zoltan developers, they will hopefully get to the bottom of this, but in the mean-time, this serial option is probably a good temporary solution.

blattms:
>I doubt this is a cluster problem

 At least @totto82 has run OPM successfully on different clusters. I would also assume that Zoltan has been used on some clusters. E.g. it is the loadbalancer in dune-alugrid, and I am sure that one was used on clusters, too. But they might use the hypergraph for loadbalancing directly. Maybe @dr-robertk could comment on this.

The invalid reads in the valgrind logs look hideous. Deep down inside the mellanox stack.

It seems like at least on Euler a different MPI implementation (MVAPICH2) is available. Switching might be worth a try.

Concerning this PR could you make an option for using the sequential loadbalance and leave the parallel as the default?

GitPaean:
The symptom is similar with the follows, 
```
It has been observed that when Xyce and Trilinos are built with
OpenMPI 2.1.x or 3.0.0 on certain unsupported operating
systems, a small number of test cases in the regression suite
crash with a segmentation fault inside the Zoltan library.
The Xyce team has determined that this is not a bug in either
Xyce or Zoltan, but is instead due to some pre-packaged
OpenMPI binaries on some operating systems having been
built with an inappropriate option. This option,
"–enable-heterogeneous" is explicitly documented in
OpenMPI documentation as broken and unusable since 2013,
but some package managers have OpenMPI binaries built with
this option explicitly enabled. Turning on this option causes the
resulting OpenMPI build to perform certain communication
operations in a way that does not adhere to the MPI standard.
There is nothing that can be done in Xyce or Zoltan to fix this
issue — it is entirely a bug in the OpenMPI library as built on
that system.
A new test case has been added to the Xyce test suite in order
to detect this problem. The test is "MPI Test/bug 967", and it
will be run whenever the test suite is invoked with the
"+parallel" tag as described in the documentation for the test
suite at https://xyce.sandia.gov/documentation/
RunningTheTests.html. If this test fails, your system has a
broken OpenMPI build that cannot be used with Xyce.
At the time of this writing, this issue is present in Ubuntu
Linux versions 17.10 and later, and there is an open bug report
for it at https://bugs.launchpad.net/ubuntu/+source/
openmpi/+bug/1731938.
The issue may be present in other distros of Linux that are
derived from Debian (as is Ubuntu), but we cannot confirm
this.
Workaround: The only workaround for this problem is to build
OpenMPI from source yourself, and not to include
"–enable-heterogeneous" in its configure options.
You should also post a bug report in your operating system's
issue tracker requesting that they rebuild their OpenMPI
binaries without the "–enable-heterogeneous" option. If you
are using Ubuntu, you should register with that issue tracking
system and add yourself to the list of people it affects in the
existing bug report (doing so increases the "heat" of the bug,
which may increase the likelihood of it being fixed).
```

which is from https://xyce.sandia.gov/downloads/_assets/documents/Release_Notes_6.11.pdf

Although the description is to claim that it is related to broken MPI in Ubuntu 18.04 (I never seen segmentation fault with this system though. I only saw I can not run OPM in parallel with it )

I am not totally sure they fixed/investigated the above issue in depth. Not necessarily our problem is really related to the reason they claim, but it shows under some circumstances, it might cause segmentation fault within Zoltan with some processes.

blattms:
thanks. So on Euler this should be gone with switching to MVAPICh2. Maybe there is another MPI implementation on SAGA to be tested, too. Anyway, if not done, this should be reported to the cluster admins also.

GitPaean:
> thanks. So on Euler this should be gone with switching to MVAPICh2. Maybe there is another MPI implementation on SAGA to be tested, too. Anyway, if not done, this should be reported to the cluster admins also.

I sent message about the problem we are facing to the system administration, but we can not expect we will receive any response.  In term of MPI options, basically, we tried all the MPI options we can find, other MPI options are facing other difficulties, basically unusable to us. This PR is very important for us being able to continue our development of OPM on the various platforms we are using.

blattms:
So will you guys make the serial loadbalancing optional and keep the current default?

blattms:
Just in case somebody with similar problems finds this via a websearch, it would be very useful for them to read what failed with MVAPICH2 and what the difficulties were. Would be useful for me, too.

blattms:
I have another question: 
Since information about the systems shared here is rather scarse, I scrolled through the documentation of [modules available on Saga](https://documentation.sigma2.no/software/installed_software/saga_modules.html). Trilinos seems to be listed there. Did you also try that one or is it missing something?

atgeirr:
Closing in favour of #491

PR #489, opened 2020-09-23
Fixes wrong communicator in CpGrid and memory overwrite in Zoltan callback

blattms:
The discussion with @GitPaean in #487 sparked this investigation. I thought that there must be a synchronisation in CpGrid::createCartesian` (rank 0 does it in line [391](https://github.com/OPM/opm-grid/blob/bea8a13fb7210b3c718b50fb508a6da7018ba477/opm/grid/cpgrid/CpGrid.cpp#L391), the others in line [336](https://github.com/OPM/opm-grid/blob/bea8a13fb7210b3c718b50fb508a6da7018ba477/opm/grid/cpgrid/CpGrid.cpp#L336). Still he was right as the communciator used was still MPI_COMM_SELF. :flushed:

Then I discovered that on nonzero ranks we might overwrite Zoltan pointers with zero (second commit fixes that). That could of course produce rather arbitrary segmentation faults and we might just be lucky on some machines/models.

### DIFF
--- a/opm/grid/common/ZoltanGraphFunctions.cpp
+++ b/opm/grid/common/ZoltanGraphFunctions.cpp
@@ -82,7 +82,7 @@ void getNullNumEdgesList(void *cpGridPointer, int sizeGID, int sizeLID,
     (void) sizeGID; (void) sizeLID; (void) numCells; (void) globalID;
     (void) localID; (void) numEdges; (void) cpGridPointer;
     // Pretend that there are no edges
-    numEdges = 0;
+    assert(numCells=0);
     *err = ZOLTAN_OK;
 }
 
--- a/opm/grid/common/ZoltanGraphFunctions.cpp
+++ b/opm/grid/common/ZoltanGraphFunctions.cpp
@@ -82,8 +82,10 @@ void getNullNumEdgesList(void *cpGridPointer, int sizeGID, int sizeLID,
     (void) sizeGID; (void) sizeLID; (void) numCells; (void) globalID;
     (void) localID; (void) numEdges; (void) cpGridPointer;
     // Pretend that there are no edges
-    assert(numCells=0);
-    *err = ZOLTAN_OK;
+    if (numCells!=0)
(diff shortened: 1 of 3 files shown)


blattms:
@GitPaean Could you give this a spin on the problematic cluster?

blattms:
jenkins build this please

opm/grid/common/ZoltanGraphFunctions.cpp
```diff
@@ -82,7 +82,7 @@ void getNullNumEdgesList(void *cpGridPointer, int sizeGID, int sizeLID,
     (void) sizeGID; (void) sizeLID; (void) numCells; (void) globalID;
     (void) localID; (void) numEdges; (void) cpGridPointer;
     // Pretend that there are no edges
-    numEdges = 0;
+    assert(numCells=0);
```
akva2, line 85: == ?

blattms: thanks. we now return ZOLTAN_FATAL if `numCells!=0`

atgeirr:
> Then I discovered that on nonzero ranks we might overwrite Zoltan pointers with zero (second commit fixes that). That could of course produce rather arbitrary segmentation faults and we might just be lucky on some machines/models.

I think this is wrong. The variable `numEdges` is a straight `int*` and not a reference, so code as it was would set the pointer to null, but not do anything about any pointed-to memory, or the pointer's value at the call site inside Zoltan.

Still, the changes here are fine and improve safety a little bit, so it can be merged when green, unless you want to refine or add more to the PR.

atgeirr:
jenkins build this please

blattms:
Yes you are right. Seems like the strange problems you guys are having really
make me nervous.

Anyway it seems good to not overwrite things unneeded even if they don't have
effects on the outside. The additional check shouldn't hurt either.

atgeirr:
I can merge this now, if you want to?

blattms:
Cool.

PR #490, opened 2020-09-23
Fix zoltan_test without MPI

blattms:


blattms:

PR #491, opened 2020-09-29
Serial partitioning

atgeirr:
This replaces #488.

This has been rebased on master, and squashed into the first commit. The second commits adds the option to choose partitioning at runtime.

A note on argument order: I put the new bool argument between the wells and trans arguments to avoid possible argument mistakes, especially since the myriad loadBalance() variants have lots of default arguments. (Why are there so many? Six main variants, or 20(!) if you count various defaulting possibilities. Happily they all call the same scatterGrid() method.)

Since this changes the API for one of the loadBalance() methods (the one used from Flow) a downstream PR is required.

### DIFF
--- a/opm/grid/common/ZoltanPartition.cpp
+++ b/opm/grid/common/ZoltanPartition.cpp
@@ -24,12 +24,131 @@
 #include <opm/grid/utility/OpmParserIncludes.hpp>
 #include <opm/grid/cpgrid/CpGridData.hpp>
 #include <opm/grid/cpgrid/Entity.hpp>
+#include <algorithm>
 
 #if defined(HAVE_ZOLTAN) && defined(HAVE_MPI)
 namespace Dune
 {
 namespace cpgrid
 {
+
+namespace {
+void setDefaultZoltanParameters(Zoltan_Struct* zz) {
+    Zoltan_Set_Param(zz, "IMBALANCE_TOL", "1.1");
+    Zoltan_Set_Param(zz, "DEBUG_LEVEL", "0");
+    Zoltan_Set_Param(zz, "LB_METHOD", "GRAPH");
+    Zoltan_Set_Param(zz, "LB_APPROACH", "PARTITION");
+    Zoltan_Set_Param(zz, "NUM_GID_ENTRIES", "1");
+    Zoltan_Set_Param(zz, "NUM_LID_ENTRIES", "1");
+    Zoltan_Set_Param(zz, "RETURN_LISTS", "ALL");
+    Zoltan_Set_Param(zz, "CHECK_GRAPH", "2");
+    Zoltan_Set_Param(zz,"EDGE_WEIGHT_DIM","0");
+    Zoltan_Set_Param(zz, "OBJ_WEIGHT_DIM", "0");
+    Zoltan_Set_Param(zz, "PHG_EDGE_SIZE_THRESHOLD", ".35");  /* 0-remove all, 1-remove none */
+}
+
+
+std::tuple<std::vector<int>, std::vector<std::pair<std::string,bool>>,
+           std::vector<std::tuple<int,int,char> >,
+           std::vector<std::tuple<int,int,char,int> > >
+makeImportAndExportLists(const CpGrid& cpgrid,
+                         const CollectiveCommunication<MPI_Comm>& cc,
+                         const std::vector<OpmWellType> * wells,
+                         const CombinedGridWellGraph& gridAndWells,
+                         int root,
+                         int numExport,
+                         int numImport,
+                         const ZOLTAN_ID_PTR exportLocalGids,
+                         const ZOLTAN_ID_PTR exportGlobalGids,
+                         const int* exportToPart,
+                         const ZOLTAN_ID_PTR importGlobalGids) {
+    int                         size = cpgrid.numCells();
+    int                         rank  = cc.rank();
+    std::vector<int>            parts(size, rank);
+    std::vector<std::vector<int> > wellsOnProc;
+
+    // List entry: process to export to, (global) index, process rank, attribute there (not needed?)
+    std::vector<std::tuple<int,int,char>> myExportList(numExport);
+    // List entry: process to import from, global index, process rank, attribute here, local index
+    // (determined later)
+    std::vector<std::tuple<int,int,char,int>> myImportList(numImport);
+    myExportList.reserve(1.2*myExportList.size());
+    myImportList.reserve(1.2*myImportList.size());
+    using AttributeSet = CpGridData::AttributeSet;
+
+    for ( int i=0; i < numExport; ++i )
+    {
+        parts[exportLocalGids[i]] = exportToPart[i];
+        myExportList[i] = std::make_tuple(exportGlobalGids[i], exportToPart[i], static_cast<char>(AttributeSet::owner));
+    }
+
+    for ( int i=0; i < numImport; ++i )
+    {
+        myImportList[i] = std::make_tuple(importGlobalGids[i], root, static_cast<char>(AttributeSet::owner),-1);
+    }
+
+
+    // Add cells that stay here to the lists. Somehow I could not persuade Zoltan to do this.
+    for ( std::size_t i = 0; i < parts.size(); ++i)
+    {
+        if ( parts[i] == rank )
+        {
+            myExportList.emplace_back(i, rank, static_cast<char>(AttributeSet::owner) );
+            myImportList.emplace_back(i, rank, static_cast<char>(AttributeSet::owner), -1 );
+        }
+    }
+    std::inplace_merge(myImportList.begin(), myImportList.begin() + numImport, myImportList.end());
+    std::inplace_merge(myExportList.begin(), myExportList.begin() + numExport, myExportList.end());
+
+
+
+
+    if( wells )
+    {
+        auto gidGetter = [&cpgrid](int i) { return cpgrid.globalIdSet().id(createEntity<0>(cpgrid, i, true));};
+        wellsOnProc =
+            postProcessPartitioningForWells(parts,
+                                            gidGetter,
+                                            *wells,
+                                            gridAndWells.getWellConnections(),
+                                            myExportList, myImportList,
+                                            cc);
+
+
+#ifndef NDEBUG
+        int index = 0;
+        for( auto well : gridAndWells.getWellsGraph() )
+        {
+            int part=parts[index];
+            std::set<std::pair<int,int> > cells_on_other;
+            for( auto vertex : well )
+            {
+                if( part != parts[vertex] )
+                {
+                    cells_on_other.insert(std::make_pair(vertex, parts[vertex]));
+                }
+            }
+            if ( cells_on_other.size() )
+            {
+                OPM_THROW(std::domain_error, "Well is distributed between processes, which should not be the case!");
+            }
+            ++index;
+        }
+#endif
+    }
+
+    std::vector<std::pair<std::string,bool>> parallel_wells;
+    if( wells )
+    {
+        parallel_wells = computeParallelWells(wellsOnProc,
+                                              *wells,
+                                              cc,
+                                              root);
+    }
+    return std::make_tuple(parts, parallel_wells, myExportList, myImportList);
+}
+} // anon namespace
+
 std::tuple<std::vector<int>, std::vector<std::pair<std::string,bool>>,
            std::vector<std::tuple<int,int,char> >,
            std::vector<std::tuple<int,int,char,int> > >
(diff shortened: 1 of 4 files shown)


atgeirr:
I consider this merge-ready, I have started a Jenkins job in the downstream PR. @blattms This should satisfy your request for keeping the current (parallel partitioning) default. If you have any other changes, please notify. Otherwise I consider this to be my "review" of the original PR.

opm/grid/common/ZoltanPartition.cpp
```diff
@@ -176,16 +137,239 @@ zoltanGraphPartitionGridOnRoot(const CpGrid& cpgrid,
...
+        for (int i = 0; i < numExport; ++i) {
+            if (exportToPart[i] >= currentIndex.size()) {
+                OPM_THROW(std::runtime_error, "Something wrong with Zoltan decomposition. "
+                          << "Debug information: exportToPart[i] = "
+                          << exportToPart[i] << ", " << "currentIndex.size() = " << currentIndex.size());
+            }
+            const auto index = currentIndex[exportToPart[i]]++ + offsets[exportToPart[i]];
+
+            if (index >= globalIndicesToSend.size()) {
+                OPM_THROW(std::runtime_error, "Something wrong with Zoltan decomposition. "
```
blattms, line 331: This will create a deadlock

opm/grid/common/ZoltanPartition.cpp
```diff
@@ -176,16 +137,239 @@ zoltanGraphPartitionGridOnRoot(const CpGrid& cpgrid,
...
+    if (cc.rank() == root) {
+        int argc=0;
+        char** argv = 0;
+        float ver = 0;
+
+        rc = Zoltan_Initialize(argc, argv, &ver);
+        zz = Zoltan_Create(MPI_COMM_SELF);
+        if ( rc != ZOLTAN_OK )
+        {
+            OPM_THROW(std::runtime_error, "Could not initialize Zoltan!");
```
blattms, line 257: This will create a deadlock.

opm/grid/common/ZoltanPartition.cpp
```diff
@@ -176,16 +137,239 @@ zoltanGraphPartitionGridOnRoot(const CpGrid& cpgrid,
...
+
+        // This is very important: by default, zoltan sets numImport to -1,
+        // as we are only running Zoltan in serial.
+        // In order to make sense of the rest of  the code, this must be set
+        // to 0.
+        numImport = 0;
+
+
+    }
+    MPI_Barrier(cc);
```
blattms, line 306: In this version this is not need as the latter scatter is a global synchronization.

opm/grid/common/ZoltanPartition.cpp
```diff
@@ -176,16 +137,239 @@ zoltanGraphPartitionGridOnRoot(const CpGrid& cpgrid,
...
+                          << "index " << index << ", " << "globalIndicesToSend.size() = "
+                          << globalIndicesToSend.size()
+                          << "\n\noffsets[exportToPart[i]] = " << offsets[exportToPart[i]]
+                          << "\n\ncurrentIndex[exportToPart[i]] = "<< currentIndex[exportToPart[i]]
+                          <<"\n\nexportToPart[i] = " << exportToPart[i]);
+            }
+
+            globalIndicesToSend[index] = exportGlobalGids[i];
+        }
+        std::vector<unsigned int> dummyIndicesForRoot(numExport, 0);
```
blattms, line 341: Isn't this a bit big, if root is receiving 0 data items?

opm/grid/common/ZoltanPartition.cpp
```diff
@@ -176,16 +137,239 @@ zoltanGraphPartitionGridOnRoot(const CpGrid& cpgrid,
...
+                                 &importProcs,    /* Process rank for source of each incoming vertex */
+                                 &importToPart,   /* New partition for each incoming vertex */
+                                 &numExport,      /* Number of vertices I must send to other processes*/
+                                 &exportGlobalGids,  /* Global IDs of the vertices I must send */
+                                 &exportLocalGids,   /* Local IDs of the vertices I must send */
+                                 &exportProcs,    /* Process to which I send each of the vertices */
+                                 &exportToPart);  /* Partition to which each vertex will belong */
+
+        // This is very important: by default, zoltan sets numImport to -1,
+        // as we are only running Zoltan in serial.
```
blattms, line 299: Are you sure that the value is different from the parallel version? For both all vertices are on rank 0 before loadbalancing and zoltan only counts objects that move to other ranks.

I would assume you need to set it because you are using it below to create vectors.

blattms (changes requested):
There are some potentials for deadlocks and wasted space in this PR.

Geat that you agree that space between if and ( are not that important.
```
if( wells )
```

opm/grid/common/ZoltanPartition.cpp
kjetil-lye: True, but what is the better solution: Doing ```MPI_Abort``` here in the innermost region of the code, or having an exception catch higher up in the hiearchy that then calls ```MPI_Abort``` on an exception?

opm/grid/common/ZoltanPartition.cpp
kjetil-lye: See comment above.

opm/grid/common/ZoltanPartition.cpp
kjetil-lye: This can be removed, it was only put in for debugging and then I forgot to remove it.

opm/grid/common/ZoltanPartition.cpp
kjetil-lye: True, this can probably be made of size 1.

opm/grid/common/ZoltanPartition.cpp
kjetil-lye: I tested this. In the beginning, ```numImport``` is set to ```0```, but then after calling zoltan, it gets set to ```-1``` when run in serial (which makes sense, when running in serial, this value is unused). 

Yes, this is set for consistency with the rest of the code.

opm/grid/common/ZoltanPartition.cpp
blattms: Catching higher does not change the problem. Cleanest way is probably
```
int rc;
if(rank==0){
  rc = ...;
}
cc.scatter(rc); //pseudo code
if (rc != ZOLTAN_OK)
  thow ...
```
Would probably need more ifs.

opm/grid/common/ZoltanPartition.cpp
blattms: Thanks for the answer

opm/grid/common/ZoltanPartition.cpp
kjetil-lye: This would only postpone the deadlock until the next global MPI operation in the code, no?

blattms: In my example all processes throw. But yes there need to multiple of these.

kjetil-lye: How would all throw? Zoltan is only initialized on rank zero, all the others have not initialized Zoltan, and therefore will not get a non-zero value of  ```rc``` giving the throw.

kjetil-lye: Nevermind, didn't see the scatter.

But why is this preferable to having

```C++
try {
   ...
} catch (...) { // or whatever exception type
   MPI_Abort(cc, -1);
}
```

this will also abort on all procs, but will have the added benifit that also non-MPI parts of the code can throw exceptions without raising a deadlock.

blattms: You are asking the wrong person. I did that but had to change to prevent user confusion.

See  bed4093b1568068faae2ffa3151148819896ede4 for what I meant.

blattms: Beware I did not even try to compile that.

blattms: use 288d8a08, which compiles

opm/grid/common/ZoltanPartition.cpp
```diff
@@ -176,16 +137,239 @@ zoltanGraphPartitionGridOnRoot(const CpGrid& cpgrid,
...
+        cc.scatter<int>(numberOfExportedVerticesPerProcess.data(), &dummyForRoot, 1, root);
+
+
+        std::vector<int> offsets(cc.size() + 1, 0);
+        std::partial_sum(numberOfExportedVerticesPerProcess.begin(),
+                         numberOfExportedVerticesPerProcess.end(), offsets.begin() + 1);
+        std::vector<unsigned int> globalIndicesToSend(numExport, 0);
+        std::vector<int> currentIndex(cc.size(), 0);
+        for (int i = 0; i < numExport; ++i) {
+            if (exportToPart[i] >= currentIndex.size()) {
```
blattms, line 323: triggers a warning because we are comparing signed and unsigned integers.

opm/grid/common/ZoltanPartition.cpp
```diff
@@ -176,16 +137,239 @@ zoltanGraphPartitionGridOnRoot(const CpGrid& cpgrid,
...
+                OPM_THROW(std::runtime_error, "Something wrong with Zoltan decomposition. "
+                          << "Debug information: exportToPart[i] = "
+                          << exportToPart[i] << ", " << "currentIndex.size() = " << currentIndex.size());
+            }
+            const auto index = currentIndex[exportToPart[i]]++ + offsets[exportToPart[i]];
+
+            if (index >= globalIndicesToSend.size()) {
```
blattms, line 330: triggers a warning because we are comparing signed and unsigned integers.

blattms:
There is one thing if you use MPI_Abort: Cleanup will be missing and you will find output files from each process on your disk.

atgeirr:
> There is one thing if you use MPI_Abort: Cleanup will be missing and you will find output files from each process on your disk.

I agree, we should communicate and terminate via exceptions (as we usually do for fatal stuff) on each process.

atgeirr:
Thanks for reviews and discussions, I'll revise the PR.

atgeirr:
> Great that you agree that space between if and ( are not that important.

It's more like I have given up... until we can get back on the clang-format train.
;-)

atgeirr:
With the current state of the PR I believe the deadlocks mentioned should be handled. Though the code is not exception-safe as such: allocation failures in standard vector could still give deadlocks for example.

I did a somewhat invasive refactoring, to wrap my head around the code. Therefore I think that the parts touched (the ZoltanSerialPartitioner class) would benefit from a new review.

blattms (approved):
I reviewed this and could not see any apparent flaws. Just some nitpicks.

I have not tested this, please run Norne Serial and Parallel for this one, ytelses and please make sure it works on the problematic clusters before merging.

blattms:
default stay the same, hence only testing on the problematic cluster with non-default is needed.

opm/grid/common/ZoltanPartition.cpp
```diff
@@ -229,147 +229,248 @@ zoltanGraphPartitionGridOnRoot(const CpGrid& cpgrid,
...
+            }
+        } else {
+            importGlobalGidsVector.resize(numImport, 0);
+        }
+        // Check for errors
+        int ok = error.empty();
+        if (cc.rank() == root) {
+            cc.broadcast<int>(&ok, 1, root);
+        } else {
+            cc.broadcast<int>(&ok, 1, root);
```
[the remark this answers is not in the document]
atgeirr: It started out as being different, then I rewrote it to be more uniform, did not notice it had become completely uniform! Will fix.

opm/grid/common/ZoltanPartition.hpp
```diff
@@ -62,6 +62,40 @@ zoltanGraphPartitionGridOnRoot(const CpGrid& grid,
...
+/// @param root The process number that holds the global grid.
+/// @return A tuple consisting of a vector that contains for each local cell of the original grid the
+///         the number of the process that owns it after repartitioning,
+///         a set of names of wells that should be defunct in a parallel
+///         simulation, vector containing information for each exported cell (global id
+///         of cell, process id to send to, attribute there), and a vector containing
+///         information for each imported cell (global index, process id that sends, attribute here, local index
+///         here)
+///
+/// @note This function will only do *serial* partioning.
```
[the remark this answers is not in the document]
atgeirr: Will fix.

atgeirr:
Issues addressed. Will run Jenkins and wait for confirmation that it works before merging.

atgeirr:
jenkins build this opm-simulators=2820

akva2:
go wash that filthy mouth!

atgeirr:
> go wash that filthy mouth!

It's a good thing I've worked with you for long enough to understand what you mean :-)

atgeirr:
jenkins build this opm-simulators=2820 please

PR #492, opened 2020-10-14
Cleanup serial zoltan

blattms:
This removes the if-else branches around the collective communications and skips qualifying the communicated type with the template parameter.

### DIFF
--- a/opm/grid/common/ZoltanPartition.cpp
+++ b/opm/grid/common/ZoltanPartition.cpp
@@ -25,6 +25,7 @@
 #include <opm/grid/cpgrid/CpGridData.hpp>
 #include <opm/grid/cpgrid/Entity.hpp>
 #include <algorithm>
+#include <type_traits>
 
 #if defined(HAVE_ZOLTAN) && defined(HAVE_MPI)
 namespace Dune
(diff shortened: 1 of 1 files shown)


blattms:
jenkins build this please

blattms:
Did missed one template parameter.

blattms:
jenkins build this please

opm/grid/common/ZoltanPartition.cpp
```diff
@@ -279,12 +280,11 @@ class ZoltanSerialPartitioner
             for (int i = 0; i < numExport; ++i) {
                 ++numberOfExportedVerticesPerProcess[exportToPart[i]];
             }
-            int dummyForRoot = 0;
-            cc.scatter<int>(numberOfExportedVerticesPerProcess.data(), &dummyForRoot, 1, root);
-        } else {
-            cc.scatter<int>(nullptr, &numImport, 1, root);
         }
 
+        cc.scatter(numberOfExportedVerticesPerProcess.data(), &numImport, 1, root);
```
atgeirr, line 285: For an empty vector, data() may or may not return a null pointer. Will scatter() work correctly here even if a non-null pointer is passed? Also for scatterv() on line 332-333.

atgeirr (commented):
Looks ok, just one question about scatter()-ing with empty arrays.

blattms:
I can run it under valgrind, but as we are scattering from root, no other rank should need to read from that array.

This view seems to be supported by the [mpi documentation](https://www.mpich.org/static/docs/latest/www3/MPI_Scatter.html), which says:

"int MPI_Scatter(const void *sendbuf, int sendcount, MPI_Datatype sendtype,
                void *recvbuf, int recvcount, MPI_Datatype recvtype, int root, MPI_Comm comm)

Input Parameters

sendbuf
    address of send buffer (choice, significant only at root) ... "

blattms:
jenkins run norne norne_parallel

atgeirr:
jenkins run norne norne_parallel please

PR #494, opened 2020-10-29
Allow partitioning without zoltan by cmake option

blattms:
Default is still to require Zoltan. But this will give me the option to test without it (to force wells to be distributed). Even if that is not possible, most of the code will be needed to allow a force loadbalance (by providing the partition number for each cell from outside.

This should not change anything for flow.

### DIFF
--- a/opm/grid/common/WellConnections.hpp
+++ b/opm/grid/common/WellConnections.hpp
@@ -40,6 +40,7 @@
 
 
 #include <opm/grid/utility/OpmParserIncludes.hpp>
+#include <opm/grid/CpGrid.hpp>
 
 namespace Dune
 {
@@ -65,7 +66,7 @@ public:
     WellConnections() = default;
 
     /// \brief Constructor
-    /// \param schedule The eclipse information
+    /// \param wells The eclipse information about the wells
     /// \param cartesianSize The logical cartesian size of the grid.
     /// \param cartesian_to_compressed Mapping of cartesian index
     ///        compressed cell index. The compressed index is used
@@ -73,6 +74,15 @@ public:
     WellConnections(const std::vector<OpmWellType>& wells,
                     const std::array<int, 3>& cartesianSize,
                     const std::vector<int>& cartesian_to_compressed);
+
+    /// \brief Constructor
+    /// \param wells The eclipse information about the wells
+    /// \param cpGrid The corner point grid
+    /// \param cartesian_to_compressed Mapping of cartesian index
+    ///        compressed cell index. The compressed index is used
+    ///        to represent the well conditions.
+    WellConnections(const std::vector<OpmWellType>& wells,
+                    const Dune::CpGrid& cpGrid);
 
     /// \brief Initialze the data of the container
     /// \param schedule The eclipse information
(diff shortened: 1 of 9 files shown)


blattms:
I think there still something in here that should not be there.

blattms:
Indeed, but it is fixed now.

blattms:
jenkins build this please

blattms:
jenkins build this please

blattms:
I am considering myself a patient man. But there is a limit. Will merge this after a jenkins run.

blattms:
jenkins build this please

blattms:
Seems like nobody cares. Might interpret that as 30 days of silent consent.

atgeirr:
I asked some time ago if you wanted me to prioritize this, and you said no, I assume the answer now would be yes?

atgeirr:
I will make a review of this on Monday morning.

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -159,49 +162,87 @@ CpGrid::scatterGrid(EdgeWeightMethod method,
...
+        int root = 0;
+
+        if (cc.rank() == root)
+        {
+            std::vector<int> parts(current_view_data_->global_cell_.size());
+            int  numParts=-1;
+            std::array<int, 3> initialSplit;
+            initialSplit[1]=initialSplit[2]=std::pow(cc.size(), 1.0/3.0);
+            initialSplit[0]=cc.size()/(initialSplit[1]*initialSplit[2]);
+            partition(*this, initialSplit, numParts, parts, false, false);
```
atgeirr, line 197: In general, numParts will not be equal to cc.size(). First, the product of initialSplit[k], k = 0, 1, 2, may not be cc.size(). Second, numParts may become smaller than that product because of inactive cells (if the final argument was true, splitting to have connected partitions could also make it larger). It seems to me that the code subsequently assumes numParts == cc.size()?

atgeirr (commented):
This looks mostly fine, but I think there is an assumption about the number of partitions that may not hold. Please check.

opm/grid/cpgrid/CpGrid.cpp
blattms: Actually `numParts`is not used below and hence disregarded. The only assumption is that `numParts <= cc.size()`. I guess that holds?
After the loadbalance all ranks will compute (possibly with an empty partition, which will result in an exception later on).

opm/grid/common/WellConnections.hpp
```diff
@@ -74,6 +75,15 @@ class WellConnections
                     const std::array<int, 3>& cartesianSize,
                     const std::vector<int>& cartesian_to_compressed);
 
+    /// \brief Constructor
+    /// \param wells The eclipse information about the wells
+    /// \param cpGrid The corner point grid
+    /// \param cartesian_to_compressed Mapping of cartesian index
```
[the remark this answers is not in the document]
blattms: fixed

blattms:
jenkins build this please

opm/grid/cpgrid/CpGrid.cpp
atgeirr: > The only assumption is that `numParts <= cc.size()`. I guess that holds?

I think you are right, since we may only round down, not up, in lines 195 and 196. It may very easily result in empty partitions and termination, for example cc.size() == 9 will give just 8 partitions. However, since this is intended as a testing device, I think it is acceptable (if we want to offer this partitioning as a backup option for users we must revisit this).

You may merge when green.

atgeirr:
Green, merging!

PR #499, opened 2020-11-06
Improve warning message for invalid NNC

joakim-hove:
Downstream of: https://github.com/OPM/opm-common/pull/2095

OK: It started out as work to reduce/improve warnings about invalid NNCs, but in the process the following things happened:

1. The initial processing of NNCs in opm-common was refactored, thereby filtering out a large fraction of the warnings due to coupling to inactive cells.

2. It turned out that the remaining warnings about NNC problems came from handling the `EDITNNC` keyword in the transmissibility calculations in opm-simulators.

So - all in all this PR is now only a tiny adaption to upstream changes in opm-common

### DIFF
--- a/opm/grid/cpgrid/processEclipseFormat.cpp
+++ b/opm/grid/cpgrid/processEclipseFormat.cpp
@@ -167,16 +167,12 @@ namespace cpgrid
             nnc_cells[PinchNNC].insert({low, high});
         }
         // Add explicit NNCs.
-        for (const auto single_nnc : nncs.data()) {
-            // Repeated NNCs will only exist in the map once
-            // (repeated insertions have no effect), and we make
-            // sure NNCs specified using either order of cells
-            // end up with the same {low, high} pair.
-            // The code that computes the transmissibilities is responsible
-            // for ensuring repeated NNC transmissibilities are added.
-            auto low = std::min(single_nnc.cell1, single_nnc.cell2);
-            auto high = std::max(single_nnc.cell1, single_nnc.cell2);
-            nnc_cells[ExplicitNNC].insert({low, high});
+        for (const auto single_nnc : nncs.input()) {
+            // Repeated NNCs will only exist in the map once (repeated
+            // insertions have no effect). The code that computes the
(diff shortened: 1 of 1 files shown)


opm/grid/cpgrid/processEclipseFormat.cpp
```diff
@@ -239,6 +239,36 @@ namespace cpgrid
             // Make the grid.
             processEclipseFormat(g, nnc_cells, z_tolerance, false, turn_normals);
         }
+
+        {
+            auto format_ijk = [this](int cell) -> std::string {
+                std::array<int,3> ijk;
+                this->getIJK(cell, ijk);
+                return std::string{"("} + std::to_string(ijk[0]) + std::string{","} + std::to_string(ijk[1]) + std::string{","} + std::to_string(ijk[2]) + std::string{")"};
```
bska, line 247: Just a tangential note:  `CpGridData::getIJK()` returns zero-based IJK tuples.  Do we want to report them to the user as one-based indices (i.e., `std::to_string(ijk[0] + 1)` and so on)

opm/grid/cpgrid/processEclipseFormat.cpp
```diff
@@ -239,6 +230,36 @@ namespace cpgrid
...
+        {
+            auto format_ijk = [this](int cell) -> std::string {
+                std::array<int,3> ijk;
+                this->getIJK(cell, ijk);
+                return std::string{"("} + std::to_string(ijk[0] + 1) + std::string{","} + std::to_string(ijk[1] + 1) + std::string{","} + std::to_string(ijk[2] + 1) + std::string{")"};
+            };
+
+
+            auto make_warning = [this, &format_ijk](const std::string& reason, const Opm::KeywordLocation& location, const Opm::NNCdata& nnc) -> std::string {
+                return std::string{"Problem with NNC connection \n"} +
```
bska, line 243: As the `C` in `NNC` means "connection", this is slightly redundant (e.g., ATM machine).  We should also remove the trailing blank before the newline.

PR #500, opened 2020-11-10
Some fixes to the Polyhedrealgrid

totto82:
This adds come changes to the polyhedrealgrid to make it compile and run with flow on SPE1
Review @dr-robertk?

### DIFF
--- a/opm/grid/polyhedralgrid/grid.hh
+++ b/opm/grid/polyhedralgrid/grid.hh
@@ -11,6 +11,7 @@
 
 //- dune-common includes
 #include <dune/common/version.hh>
+#include <dune/common/parallel/mpihelper.hh>
 
 //- dune-grid includes
 #include <dune/grid/common/grid.hh>
@@ -120,7 +121,8 @@ namespace Dune
       typedef PolyhedralGridIdSet< dim, dimworld, ctype > GlobalIdSet;
       typedef GlobalIdSet  LocalIdSet;
 
-      typedef Dune::CollectiveCommunication< Grid > CollectiveCommunication;
+      typedef Dune::MPIHelper::MPICommunicator MPICommunicator;
+      typedef Dune::CollectiveCommunication<MPICommunicator> CollectiveCommunication;
 
       template< PartitionIteratorType pitype >
       struct Partition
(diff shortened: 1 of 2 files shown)


opm/grid/polyhedralgrid/grid.hh
```diff
@@ -848,61 +848,84 @@ namespace Dune
...
+    /// \param handle The data handle describing the data and responsible for
+    ///         gathering and scattering the data.
+    template<class DataHandle>
+    void scatterData(DataHandle& handle) const
+    {
+#if HAVE_MPI
+        //if(!distributed_data_)
+        //    OPM_THROW(std::runtime_error, "Moving Data only allowed with a load balanced grid!");
+        //distributed_data_->scatterData(handle, data_.get(), distributed_data_.get(), cellScatterGatherInterface(),
+                                       pointScatterGatherInterface());
```
blattms, line 872: Should this line be commented out?

opm/grid/polyhedralgrid/grid.hh
```diff
@@ -848,61 +848,84 @@ namespace Dune
...
+    /// \brief Moves data from the global (all data on process) view to the distributed view.
+    ///
+    /// This method does not do communication but assumes that the global grid
+    /// is present on every process and simply copies data to the distributed view.
+    /// \tparam DataHandle The type of the data handle describing the data and responsible for
+    ///         gathering and scattering the data.
+    /// \param handle The data handle describing the data and responsible for
+    ///         gathering and scattering the data.
+    template<class DataHandle>
+    void scatterData(DataHandle& handle) const
```
blattms, line 866: I am a bit torn here.
Providing a function that does not do what it says it would might be a bit surprising. If we keep it, then we should at least unconditionally throw an exception, if there are more than 1 MPI processes. Or work around it at the call site.

blattms: e.g. by specializing for CpGrid or using the vanguards. Where is it called BTW?

blattms (changes requested):
Apart form the method pointed out in the comments this seems sound

opm/grid/polyhedralgrid/grid.hh
totto82: The scatterData and the communicate method are just copied from the CPGRID. MPI does not work with polyhedrealgrid so I will make sure the methods throw if they are called and cleanup the comments.

totto82:
I have added throws and tested it with MPI.

PR #507, opened 2021-01-06
make it possible to set the imbalance tolerance dynamicaly

totto82:
Too large imbalance tolerance may lead to empty partitions. This gives the user a way to set this from the application.


bska:
jenkins build this opm-simulators=3003 please

atgeirr:
Looks fine, merging.

PR #510, opened 2021-01-20
Support partition with external loadbalancing information

blattms:
Instead of using Zoltan or our Vanilla loadbalancing the information about the loadbalancing (partition index per cell) is provided from outside. That allows to use any loadbalancer or even guess a load balancing.


blattms:
jenkins build this please

blattms:
Closing in favor of  #511

PR #511, opened 2021-01-25
Activate distributed wells

blattms:
Based on #510 which I will close in favor of this one.

Besides loadbalancing using external information, this adds a parameter to allow loadbalancing with Zoltan without taking the wells into account and ending with wells that perforate interior cells on multiple processes. Checking whether this makes sense (i.e. that we only have standard wells) is left to the simulator.


blattms:
Needs OPM/opm-common#2241

blattms:
The PR in opm-common is now included here (hopefully with all comments addressed).

blattms:
jenkins build this serial opm-simulators=3029 please

tests/test_communication_utils.cpp
```diff
@@ -0,0 +1,196 @@
...
+  MPIError(std::string s, int e) : errorstring(s), errorcode(e){}
+  /** @brief The error string. */
+  std::string errorstring;
+  /** @brief The mpi error code. */
+  int errorcode;
+};
+
+#ifdef HAVE_MPI
+void MPI_err_handler(MPI_Comm *, int *err_code, ...){
+  char *err_string=new char[MPI_MAX_ERROR_STRING];
```
bska, line 43: Do you need to use the free store here?  What's a typical value of `MPI_MAX_ERROR_STRING`?  If it's just a couple of hundred bytes would you be able to use a `std::array<char,N>` here?  By the way, does `MPI_MAX_ERROR_STRING` include the terminating `\0` character?  If you **do** have to use the free store, could you use `unique_ptr<char[]>` instead?

tests/test_communication_utils.cpp
```diff
@@ -0,0 +1,196 @@
...
+  int err_length;
+  MPI_Error_string(*err_code, err_string, &err_length);
+  std::string s(err_string, err_length);
+  std::cerr << "An MPI Error ocurred:"<<std::endl<<s<<std::endl;
```
bska, line 47: Using `std::cerr` and `std::string` here means you need to
```C++
#include <iostream>
#include <string>
```
in this translation unit.

tests/test_communication_utils.cpp
```diff
@@ -0,0 +1,196 @@
...
+        std::mt19937 mersenneEngine {rndDevice()};  // Generates random integers
+        std::uniform_int_distribution<int> dist {1, 10};
+        
+        auto gen = [&dist, &mersenneEngine](){
+                       return dist(mersenneEngine);
+                   };
+        auto begin = std::begin(sizes);
+        auto end = std::end(sizes);
+        std::generate(begin, end, gen);
+        std::partial_sum(begin, end, std::begin(displ)+1);
```
bska, line 105: This needs
```C++
#include <numeric>
```

tests/test_communication_utils.cpp
```diff
@@ -0,0 +1,196 @@
...
+    //std::vector<int> sizes(comm.size());
+    std::vector<int> expectedDispl(comm.size()+1);
+    std::vector<int> displ;
+    std::vector<double> expectedAllVals;
+    std::vector<double> allVals;
+    std::vector<double> myVals;
+    double d=0;
+    
+    std::tie(expectedAllVals, myVals, std::ignore, expectedDispl) = createParallelData(comm, d);
+    std::tie(allVals, displ) = Opm::allGatherv(myVals, comm);
```
bska, line 147: Using `std::tie` means you need to
```C++
#include <tuple>
```
in this translation unit.

bska (commented):
I just looked at the functions for gathering values onto a root process.  Please take some care in making sure that the translation units are self contained by including the requisite headers.

opm/grid/common/CommunicationUtils.hpp
```diff
@@ -0,0 +1,92 @@
...
+
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+*/
+#ifndef OPM_COMMUNICATION_UTILS_HPP
+#define OPM_COMMUNICATION_UTILS_HPP
+
+#include<vector>
+#include<numeric>
+#include<tuple>
```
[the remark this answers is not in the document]
blattms: Which compiler? I think this is self contained as tuple includes utility. Using tuple helps the user here in case he wants to use `std::tie`.

tests/test_communication_utils.cpp
blattms: Shouldn't the include of tuple in opm/grid/common/CommunicationUtils.hpp be enough?

blattms:
Is there a new rule that we include in each header and source file all the headers it uses no matter whether it is already included transitively from somewhere else?
In that case somebody needs to come up with an utility to test this (no idea how that should be done), because my brain is not capable to know each header that is needed and I need to rely on the compiler or some other tests.

tests/test_communication_utils.cpp
blattms: Shouldn't the include of that header in opm/grid/common/CommunicationUtils.hpp be enough?

tests/test_communication_utils.cpp
blattms: > Do you need to use the free store here?

Nope and I will change that (use a vector) just to make you happy.

> What's a typical value of `MPI_MAX_ERROR_STRING`? If it's just a couple of hundred bytes would you be able to use a `std::array<char,N>` here?

That seems implementation dependent and I do not want to look it up for the particular MPI implementation that I am using.

> By the way, does `MPI_MAX_ERROR_STRING` include the terminating `\0` character? If you **do** have to use the free store, could you use `unique_ptr<char[]>` instead?

As the [MPI method](https://www.mpich.org/static/docs/latest/www3/MPI_Error_string.html) does not tell anything about that, but return the length, I think one should assume that it is not terminated. Better safe than sorry.

I should have just skipped the error handler...

tests/test_communication_utils.cpp
blattms: I guess these might be included transitively via the boost unit test system or something else?
Anyway I will add them to save time.

blattms: added them nevertheless to speed up things.

tests/test_communication_utils.cpp
blattms: added them nevertheless to speed up things.

tests/test_communication_utils.cpp
blattms: Added nevertheless...

blattms (commented):
All comment should be addressed now.

I have seen people removing multiple header includes, but adding them by request was new to me and feels a bit strange.

blattms:
jenkins build this serial opm-simulators=3029 please

blattms:
jenkins build this serial opm-simulators=3029 please

tests/test_communication_utils.cpp
blattms: Unsing std::vector now

tests/test_communication_utils.cpp
blattms: Well, at least now it is

blattms:
Will try to recover jenkins.

opm/grid/common/CommunicationUtils.hpp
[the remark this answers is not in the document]
blattms: Added nervertheless

blattms:
Note the only change affecting the default flow here, is that we print (`OpmLog::info`) the well distribution now, which might be considered noise. E.g. for SPE9 and flow_distribute_z:

```
Load balancing distributed the wells as follows:
  well name            ranks with perforated cells
---------------------------------------------------
           INJE1       3
         PRODU10       1       0
         PRODU11       0       1
         PRODU12       1       0
         PRODU13       1       0
         PRODU14       1       0
         PRODU15       1       0
         PRODU16       1       0
         PRODU17       1       0
         PRODU18       0       1
         PRODU19       1       0
          PRODU2       1       0
         PRODU20       0       1
         PRODU21       0       1
         PRODU22       0       1
         PRODU23       0       1
         PRODU24       1       0
         PRODU25       1       0
         PRODU26       0       1
          PRODU3       0       1
          PRODU4       1       0
          PRODU5       0       1
          PRODU6       0       1
          PRODU7       0       1
          PRODU8       0       1
          PRODU9       0       1
```

blattms:

blattms:
jenkins build this opm-simulators=3029

blattms:
jenkins build this opm-simulators=3029 please

blattms:
will fire up another build, and merge if green. Silence is interpreted as consent.

blattms:
jenkins build this opm-simulators=3029 please

PR #512, opened 2021-01-27
Verteq2

rbe051:
Merged branch verteq2 from @dr-robertk  into master


blattms:
Welcome to OPM. Thanks a lot for the contribution. I am sure this is very valuable but you could make our life a bit easier to assess and understand the value. Please think about doing this (would be very welcome):

- Please add a more meaningful description. E.g. what is the verteq grid.
- If possible split the PR. One for verteq, one for Polyhedral improvement/fixes etc.
- Maybe you can even squash some of the commits (e.g. "Still not working" looks strange in the master branch)

dr-robertk:
The verteq (Vertical Equilibrium) is a revitalization of a form OPM project called verteq from some years back. It basically allows to colaps a column of grid cells from a corner point grid into a 2d cell while keeping a way of communicating between the two grids. This is used for longterm (>100 years) large scale (large reservoirs) CO2 simulations. The original code from Roland was based on UnstructuredGrid and thus with this PR we try to revitalize some of that code to be used in present day OPM. I think @totto82 can give you the full picture and @atgeirr can fill you in on the history.

totto82 (commented):
Thanks for providing the PR. I know this is not purely your work, so thanks for the effort of trying to get it into the master code. Some cleanup and I think this can be merged. I agree with @blattms  that a rebase -i to squash some of the commits would be helpful. Just give me a hint if you need guidance in what commits that needs be squashed together. Getting the VE code back in shape is a valuable contribution IMO.

PR #515, opened 2021-03-09
fixed: quell unused variable warning without MPI

akva2:
just some janitoring..


akva2:

blattms (approved):
Apart from initialization `procs` is not used anyway. Could also have been dropped completely, but probably not worth the additional effort.

akva2:
did not notice. let's do it properly.

akva2:
failure is due to known issue in opm-models. this does the job it's supposed to do so i'm merging as it is approved.

PR #516, opened 2021-03-22
getting aquifer cells with zero volume work

GitPaean:
even these cells have zero cell volume from the grid geometry, we can still specify pore volume (not handled yet, because opm-common does not support it yet) or numercial aquifer cells with them. We have to keep them in the grid.

### DIFF
--- a/opm/grid/common/GeometryHelpers.hpp
+++ b/opm/grid/common/GeometryHelpers.hpp
@@ -6,7 +6,7 @@
 //
 // Author(s): Atgeirr F Rasmussen <atgeirr@sintef.no>
 //            Halvor M Nilsen     <hnil@sintef.no>
-//            Bjørn Spjelkavik    <bsp@sintef.no>
+//            BjÃ¸rn Spjelkavik    <bsp@sintef.no>
 //
 // $Date$
 //
@@ -148,10 +148,10 @@ namespace Dune
             for (int i = 0; i < num_points; ++i) {
                 Point tet[4] = { cell_centroid, face_centroid, points[i], points[(i+1)%num_points] };
                 double small_volume = std::fabs(simplex_volume(tet));
-                assert(small_volume > 0);
+                assert(small_volume >= 0);
                 tot_volume += small_volume;
             }
-            assert(tot_volume>0);
+            assert(tot_volume>=0);
             return tot_volume;
         }
 
@@ -172,7 +172,7 @@ namespace Dune
             for (int i = 0; i < num_points; ++i) {
                 Point tet[4] = { cell_centroid, face_centroid, points[i], points[(i+1)%num_points] };
                 double small_volume = std::fabs(simplex_volume(tet));
-                assert(small_volume > 0);
+                assert(small_volume >= 0);
                 Point small_centroid = tet[0];
                 for(int j = 1; j < 4; ++j){
                     small_centroid += tet[j];
@@ -182,7 +182,7 @@ namespace Dune
                 tot_volume += small_volume;
             }
             centroid /= tot_volume;
-            assert(tot_volume>0);
+            assert(tot_volume>=0);
             return centroid;
         
         }
--- a/opm/grid/cpgrid/processEclipseFormat.cpp
+++ b/opm/grid/cpgrid/processEclipseFormat.cpp
@@ -99,6 +99,7 @@ namespace Dune
                        const cpgrid::OrientedEntityTable<0, 1>& c2f,
                        const std::vector<std::array<int,8> >& c2p,
                        const std::vector<int>& face_to_output_face,
+                       const std::unordered_map<size_t, double>& aquifer_cell_volumes,
                        cpgrid::EntityVariable<cpgrid::Geometry<3, 3>, 0>& cell_geom,
                        cpgrid::EntityVariable<cpgrid::Geometry<2, 3>, 1>& face_geom,
                        cpgrid::EntityVariable<cpgrid::Geometry<0, 3>, 3>& point_geom,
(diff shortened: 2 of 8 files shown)


GitPaean:
Current approach is that we make sure numerical aquifer cells not removed from the grid processing. 

Potentially, with keywords `PORV` can also used to increase the pore volume of a cell with zero volume geometrically in the grid to make it active and use in the simulation. Based on the checking for pore volume can be more general. I changed back and forth between these two ways. After finishing this sentence, I realized maybe it is good to do this based on pore volume.

GitPaean:
I will test and potentially update this PR with pore volume. The current PR is for testing purpose now.  It might break a lot of things.

GitPaean:
jenkins build this opm-common=2375 opm-simulators=3124

bska:
> opm-common=2375 opm-simulators=3124

Note: As a rule of thumb, you should generally trigger a build check from the "most" upstream repository&mdash;in this case opm-common&mdash;since that's the only way to guarantee that all requisite unit and regression tests are actually run as part of the build check.

GitPaean:
After more thinking, not totally sure how to use `PORV` to activate a geometrically zero volume cell, because the porosity will be `inf`. So this PR will still focus on numerical aquifer cells. 

I will revise this PR to make it ready for review.

GitPaean:
jenkins build this opm-simulators=3124 please

GitPaean:
There might be still more adjustments, but it is good someone can begin commenting on it.

opm/grid/cpgrid/processEclipseFormat.cpp
```diff
@@ -271,7 +279,14 @@ namespace cpgrid
 #ifdef VERBOSE
         std::cout << "Building geometry." << std::endl;
 #endif
-        buildGeom(output, cell_to_face_, cell_to_point_, face_to_output_face, geometry_.geomVector(std::integral_constant<int,0>()),
+        // here we need the cell volumes based on the active index order
+        std::unordered_map<size_t, double> aquifer_cell_volumes_local;
+        for (size_t i = 0; i < this->global_cell_.size(); ++i) {
+            if (aquifer_cell_volumes.count(this->global_cell_[i]) > 0) {
+                aquifer_cell_volumes_local.insert({i, aquifer_cell_volumes.at(this->global_cell_[i])});
+            }
```
bska, line 287: This will search the map twice for the same cell index.  I think I'd personally write this with iterators along the lines of
```C++
for (auto nc = this->global_cell_.size(), i = 0*nc; i < nc; ++i) {
    auto aquCellPos = aquifer_cell_volumes.find(this->global_cell_[i]);
    if (aquCellPos != aquifer_cell_volumes.end()) {
        aquifer_cell_volumes_local.emplace(i, aquCellPos->second);
    }
}
```

bska (approved):
This looks like a reasonable workaround.  Integrating this aspect more tightly into the grid processing will likely require a more bottoms-up approach with greater risk of destabilising the system.

opm/grid/cpgrid/processEclipseFormat.cpp
GitPaean: Yes. You are right. Thanks for pointing it out.

GitPaean:
jenkins build this opm-simulators=3124 please

opm/grid/common/GeometryHelpers.hpp
```diff
@@ -148,10 +148,13 @@ namespace Dune
             for (int i = 0; i < num_points; ++i) {
                 Point tet[4] = { cell_centroid, face_centroid, points[i], points[(i+1)%num_points] };
                 double small_volume = std::fabs(simplex_volume(tet));
-                assert(small_volume > 0);
+                if (small_volume == 0.) {
+                    small_volume = 0.;
+                }
```
blattms, line 153: This seems to have no effect.

opm/grid/cpgrid/processEclipseFormat.cpp
```diff
@@ -254,7 +257,12 @@ namespace cpgrid
         std::cout << "Processing eclipse data." << std::endl;
 #endif
         processed_grid output;
-        process_grdecl(&input_data, z_tolerance, &output);
+        const size_t global_nc = input_data.dims[0] * input_data.dims[1] * input_data.dims[2];
+        std::vector<int> is_aquifer_cell(global_nc, 0);
```
blattms, line 261: Could we set up `aquifer_cell_volumes_local`here and use that instead of `is_aquifer_cell` to judge?

opm/grid/common/GeometryHelpers.hpp
GitPaean: Sorry, that is a debug code. I will remove it.

opm/grid/cpgrid/processEclipseFormat.cpp
GitPaean: It looks like that we can not, because `aquifer_cell_volumes_local` needs indices after processed.

GitPaean:
jenkins build this opm-simulators=3124 please

GitPaean:
jenkins build this opm-simulators=3124 please

GitPaean:
jenkins build this opm-simulators=3124 please

GitPaean:
Not totally sure about the jenkins failures. It looks like the failed cases are irrelevant. 

I will check a few locally to be sure.

GitPaean:
jenkins build this opm-simulators=3124 please

GitPaean:
Finally, jenkins is green without irrelevant failures.  Please merge this PR and OPM/opm-simulators#3124

bska:
jenkins build this opm-simulators=3124 please

bska:
I'll merge this now.  Thanks for getting us closer to being able to run real field cases with numerical aquifers.

PR #517, opened 2021-03-25
using pinched grid for numerical aquifers.

GitPaean:


GitPaean:
serial results look fine while parallel results are wrong.

opm/grid/cpgrid/processEclipseFormat.cpp
```diff
@@ -268,6 +269,22 @@ namespace cpgrid
...
+            std::vector<int> new_actnum(global_nc, 0);
+            for (int i = 0; i < output.number_of_cells; ++i) {
+                new_actnum[output.local_cell_index[i]] = 1;
+            }
+            const auto& ecl_grid = ecl_state.getInputGrid();
+            const auto& fp = ecl_state.fieldProps();
+            aquifer.mutableNumericalAquifers().addAquiferConnections(deck, ecl_grid, new_actnum);
+            const auto& aquifer_nnc = aquifer.numericalAquifers().aquiferConnectionNNCs(ecl_grid, fp);
+            // We need to update the nnc in the ecl_state
+            ecl_state.appendInputNNC(aquifer_nnc);
```
GitPaean, line 282: Only process zero can reach here. It means other processes might not have this updated NNC information.

GitPaean: It is fixed in the ebos side.

GitPaean:
the parallel results looks fine now. fixing all the tests now.

GitPaean:
In this PR, we process the numerical aquifers connection information in the middle of the grid processing, which does not look natural, but I do not think it is an easy task to refactor to do the PINCH, MINPV in the parser.  I am marking this PR to be ready for review.

opm/grid/cpgrid/CpGridData.hpp
```diff
@@ -203,18 +203,17 @@ class CpGridData
...
 #endif
 
     /// Read the Eclipse grid format ('grdecl').
     /// \param input_data the data in grdecl format, declared in preprocess.h.
     /// \param z_tolerance points along a pillar that are closer together in z
     ///        coordinate than this parameter, will be replaced by a single point.
     /// \param remove_ij_boundary if true, will remove (i, j) boundaries. Used internally.
-    void processEclipseFormat(const grdecl& input_data, const std::array<std::set<std::pair<int, int>>, 2>& nnc, double z_tolerance, bool remove_ij_boundary, bool turn_normals = false,
-                              const std::unordered_map<size_t, double>& aquifer_cell_volumes = std::unordered_map<size_t, double>());
+    void processEclipseFormat(const grdecl& input_data, Opm::EclipseState& ecl_state, const Opm::Deck& deck,
```
blattms, line 215: Your really need to document this properly, as the change seems a bit dangerous. We seem to call this function with an empty ecl_state, which is fine as long as there are no numerical aquifers.

The danger is that in the future fellow developer might assume that the ecl_state is always valid and use it for other things. In that case they will have a hard time debugging.

joakim-hove: This looks very dangerous to me - the empty `EclipseState` is purely an artifact of the serialization - it is deeply invalid and should not be used for anything!

GitPaean: I will add the document here. If we do not use `EclipseState` here, we will need the `EclipseGrid`, `FieldPropsManager`, `AquiferConfig` and `inputNNC` of the `EclipseState`. I really do not want to add so many arguments and passing them around. One of the big difficulties when developing this part is that we have five different functions with the same name `processEclipseFormat`, with different arguments, and doing different but related things, and call one other.  Maybe it is good idea to rename the functions to get a little bit more descriptive (maybe I should have done it before developing this PR). Anyway, it is not related to this PR directly. 

> We seem to call this function with an empty ecl_state,

What do you mean here?

GitPaean: The empty `ecl_state` happens in the following code in CpGrid.cpp. 
```c++
    void CpGrid::processEclipseFormat(const grdecl& input_data, double z_tolerance,
                                      bool remove_ij_boundary, bool turn_normals)
    {
        // TODO: just fix the interface, not sure it will be problematic
        Opm::EclipseState ecl_state;
        Opm::Deck deck;
        using NNCMap = std::set<std::pair<int, int>>;
        using NNCMaps = std::array<NNCMap, 2>;
        NNCMaps nnc;
        current_view_data_->processEclipseFormat(input_data, ecl_state, deck, nnc, z_tolerance, remove_ij_boundary, turn_normals);
```
Do you think to use pointers here better than using the references? Then we can just pass in `nullptr` for the above situation.  Please comment.

joakim-hove: Yes I think using a pointer or optional is better here - as I said - things will go fast to pieces if you use a default constructed EclipseState

GitPaean: Okay. I will use pointers here. I think optional to reference is not there yet, right? Correct me if I am wrong here.

blattms: Wouldn't std::optional make a copy of EclipseState? That might be expensive...
You would need to use a std::optional with a std::reference_wrapper<EclipseState> inside to prevent the copy. Not sure whether that really works.

joakim-hove: >You would need to use a std::optional with a std::reference_wrapper inside to prevent the copy. Not sure whether that really works.

That works - I have tried it in another context; but maybe slightly over engineered? I would go with a pointer, but this is not my turf.

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -493,7 +493,10 @@ CpGrid::scatterGrid(EdgeWeightMethod method,
         g.coord = &coord[0];
         g.zcorn = &zcorn[0];
         g.actnum = &actnum[0];
-        current_view_data_->processEclipseFormat(g, {}, 0.0, false, false);
+        using NNCMap = std::set<std::pair<int, int>>;
+        using NNCMaps = std::array<NNCMap, 2>;
+        NNCMaps nnc;
+        current_view_data_->processEclipseFormat(g, nullptr, nullptr, nnc, 0.0, false, false);
```
GitPaean, line 499: not sure why `nnc` can not use `{}` as the original code, will try more.

opm/grid/CpGrid.hpp
```diff
@@ -262,10 +262,10 @@ namespace Dune
         /// \param turn_normals if true, all normals will be turned. This is intended for handling inputs with wrong orientations.
         /// \param clip_z if true, the grid will be clipped so that the top and bottom will be planar.
         /// \param poreVolume pore volumes for use in MINPV processing, if asked for in deck
-        void processEclipseFormat(const Opm::EclipseGrid* ecl_grid, bool periodic_extension, bool turn_normals = false, bool clip_z = false,
-                                  const std::vector<double>& poreVolume = std::vector<double>(),
-                                  const Opm::NNC& = Opm::NNC(),
-                                  const std::unordered_map<size_t, double>& aquifer_cell_volumes = std::unordered_map<size_t, double>());
+        // TODO: add description of the arguments
+        void processEclipseFormat(Opm::EclipseState* ecl_state,
```
GitPaean, line 266: One thing unfortunate due to the change here is that, originally the main input is `EclipseGrid`, but now the main input `EclipseGrid` becomes something hidden in the `EclipseState`. 

Maybe we still can keep `EclipseGrid` as the main input, but also has `EclipseState` there for extra information. But we can not guarantee the `EclipseGrid` is the grid from `EclipseState`.

opm/grid/cpgrid/processEclipseFormat.cpp
```diff
@@ -257,17 +256,44 @@ namespace cpgrid
...
+            const auto& aquifer = ecl_state->aquifer();
+            if (aquifer.hasNumericalAquifer()) {
+                const size_t global_nc = input_data.dims[0] * input_data.dims[1] * input_data.dims[2];
+                std::vector<int> new_actnum(global_nc, 0);
+                for (int i = 0; i < output.number_of_cells; ++i) {
+                    new_actnum[output.local_cell_index[i]] = 1;
+                }
+                const auto& ecl_grid = ecl_state->getInputGrid();
+                const auto& fp = ecl_state->fieldProps();
+                aquifer.mutableNumericalAquifers().addAquiferConnections(*deck, ecl_grid, new_actnum);
```
GitPaean, line 287: @joakim-hove , it is here the `mutableNumericalAquifers()` will be modified.

GitPaean:
rebasing turned out to be a little tedious. So merging is used here to fix the conflicts.

opm/grid/cpgrid/CpGridData.hpp
GitPaean: This change really backfired in the downstream modules. in opm-upscaling module, it only looks at the eclipse_grid. With the same grid deck, you can generate a EclipseGrid and also possible for a EclipseState with an EclipseGrid. But these two grid are different, because since there is no pore volumes, all the grid cells will be marked in active.

opm/grid/CpGrid.hpp
GitPaean: I will recovered to be the main argument is the `EclipseGrid`. `EclipseState` will be secondary argument to provide other information. When there is only `EclipseState` provided, we use the `EclipseGrid` belonging to `EclipseState`.

GitPaean:
This is the major parts related to this effort to get this piece of numerical aquifers in.  Besides some code related to numerical aquifers directly, the main part is interface changes related to functions `processEclipseFormat`.  Please review and suggest. 

The jenkins was triggered from upsctream PR OPM/opm-common#2389 and is green now.

bska (commented):
Looks good to me.  I just had a couple of minor questions.  Once those are addressed, I'm prepared to merge this along with its upstream OPM/opm-common#2389.

opm/grid/cpgrid/processEclipseFormat.cpp
```diff
@@ -248,8 +250,8 @@ namespace cpgrid
 
 
     /// Read the Eclipse grid format ('.grdecl').
-    void CpGridData::processEclipseFormat(const grdecl& input_data, const NNCMaps& nnc, double z_tolerance, bool remove_ij_boundary, bool turn_normals,
-                                          const std::unordered_map<size_t, double>& aquifer_cell_volumes)
+    void CpGridData::processEclipseFormat(const grdecl& input_data, Opm::EclipseState* ecl_state, const Opm::Deck* deck,
+                                          NNCMaps& nnc, double z_tolerance, bool remove_ij_boundary, bool turn_normals)
```
bska, line 254: I didn't notice this before.  Strictly speaking, this particular function is not protected by `HAVE_ECL_INPUT` so it was originally supposed to be callable without types from opm-common.  I'd be happy to remove that restriction, but maybe we should discuss this topic a little more before we merge the PR.

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -541,7 +542,10 @@ CpGrid::scatterGrid(EdgeWeightMethod method,
     void CpGrid::processEclipseFormat(const grdecl& input_data, double z_tolerance,
                                       bool remove_ij_boundary, bool turn_normals)
     {
-        current_view_data_->processEclipseFormat(input_data, {}, z_tolerance, remove_ij_boundary, turn_normals);
+        using NNCMap = std::set<std::pair<int, int>>;
+        using NNCMaps = std::array<NNCMap, 2>;
+        NNCMaps nnc;
+        current_view_data_->processEclipseFormat(input_data, nullptr, nullptr, nnc, z_tolerance, remove_ij_boundary, turn_normals);
```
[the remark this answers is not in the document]
GitPaean: Not totally sure that I understand your question correctly here. 

For this situation, nnc will always be empty. The only reason it can be modified within this function call is numerical aquifers, but we pass in `nullptr` for `EclipseState*`, so no numerical aquifers.  `nnc` created here is purely for interface.  It does not look pretty.  

Suggestion is very welcomed here.

opm/grid/cpgrid/processEclipseFormat.cpp
GitPaean: Sure. I am not that familiar with grid related stuff. 

In my own opinion, the grid processing (getting the final ACTNUM, we should not drop any information before determining the final ACTNUM, so a cell is free to change from inactive to active.) should be done in the early stage of the parsing, so all the numerical aquifers stuff can stay in opm-common.  

The current way is intruding because we do the processing (can remove more cells) after parsing, while parsing needs processed grid and parsing information also intervene the grid processing.

blattms:
Thanks for adding documentation. My main point for requesting it, was that developers can see which are optional or which do only have to have a limited state in certain circumstances. Seems like that is still not clear from the text. Or maybe this did change?

I would like to get the following info from the documentation:
1. If and when I am allowed to pass nullpointers
2. If and when I allowed to pass incomplete objects (e.g. EclState not reflecting the whole model)

Or has that changed meanwhile.
Would be cool if that was be easy to find. Otherwise I concur with @bska's opinion (not even sure whether I have the required knowledge to do so).

blattms:
I am approving now as this could be done in a separate PR (which often means it will slip).

blattms (approved):
Apart from doc improvements this looks fine.

GitPaean:
> If and when I am allowed to pass nullpointers

Yes. Null pointers are allowed, which makes it okay to only have the grid passed in.  When you do not have `ElcipseState`  or ` `Deck` available, you can pass in null pointers. This is referring to the function `void CpGridData::processEclipseFormat` in `processEclipseFormat.cpp`. 

> If and when I allowed to pass incomplete objects (e.g. EclState not reflecting the whole model)

This is hard to tell. For example, if a deck only have the GRID description without providing `PORO` or `PORV` information, all the grid cells will be marked inactive. Then it will probably not work very well.  But I am not sure it is a problem of the `EclipseState` or here.

joakim-hove (changes requested):
This PR uses the raw `Deck` datastructure - outside of opm-common, we must find a way to avoid that. I'm trying to come up with a suggestion.

joakim-hove:
This series of PR's brings in a completely new thing - namely that pinched out cells are activated by the Aquifer code. This is complex code - breaking with an assumption (that cells can only be *de*activated) which has been correct for a long time. This should in my opinion not be merged without proper testing - preferably both unit testing and integration test.

opm/grid/cpgrid/processEclipseFormat.cpp
```diff
@@ -258,17 +260,44 @@ namespace cpgrid
...
         }
 
+        if (ecl_state) {
+            assert(deck);
+            const auto& aquifer = ecl_state->aquifer();
+            if (aquifer.hasNumericalAquifer()) {
+                const size_t global_nc = input_data.dims[0] * input_data.dims[1] * input_data.dims[2];
+                std::vector<int> new_actnum(global_nc, 0);
+                for (int i = 0; i < output.number_of_cells; ++i) {
+                    new_actnum[output.local_cell_index[i]] = 1;
```
joakim-hove, line 287: That this line of code actually "turns on" a cell is so special it deserves a comment.

joakim-hove:
> This PR uses the raw Deck datastructure - outside of opm-common, we must find a way to avoid that. I'm trying to come up with a suggestion.

My only suggestion is to have a temporary vector in the numerical aquiferclass with *all* the connections coming from the `AQUCON` keyword, and then a method e.g. `NumericalAquifer::post_process_connections( new_actnum );`

opm/grid/cpgrid/processEclipseFormat.cpp
GitPaean: Okay. This is not the line turning on a aquifer cell. This line just generates the actnum based on `output`.

GitPaean: Here is the place that forcefully keeps aquifer cells active. 

https://github.com/OPM/opm-grid/blob/a28f0bd0dfdf724ee19b3f975efa2c851cd15060/opm/grid/cpgpreprocess/preprocess.c#L354

GitPaean:
> My only suggestion is to have a temporary vector in the numerical aquiferclass with all the connections coming from the AQUCON keyword, and then a method e.g. NumericalAquifer::post_process_connections( new_actnum );

Done, waiting for Jenkins to pass. At the same time, there is one function `processEclipseFormat` using `const Deck&` as input argument already in `CpGridData.hpp`.

joakim-hove:
> At the same time, there is one function processEclipseFormat using const Deck& as input argument already in CpGridData.hpp.

A little surprising - I think it should be removed.

joakim-hove (approved):

PR #518, opened 2021-03-26
Stop pretending that we support CMake 2.8

blattms:
We actually already require at least CMake 2.8.12 due to the embedded pybind11 (some tests of it are even at 3.0). Anyway as Ubuntu LTS has 3.10.2 I doubt that anything less is tested by us.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,7 +17,7 @@
 # Mandatory call to project
 project(opm-grid C CXX)
 
-cmake_minimum_required (VERSION 2.8)
+cmake_minimum_required (VERSION 3.10)
 # project information is in dune.module. Read this file and set variables.
 # we cannot generate dune.module since it is read by dunecontrol before
 # the build starts, so it makes sense to keep the data there then.


CMakeLists.txt
```diff
@@ -17,7 +17,7 @@
 # Mandatory call to project
 project(opm-grid C CXX)
 
-cmake_minimum_required (VERSION 2.8)
+cmake_minimum_required (VERSION 3.0)
```
bska, line 20: This should probably say `3.10` instead.

bska:
Note: This (opm-grid) still says `3.0` instead of `3.10`, even after the rebase.

bska:
@blattms : Are you going to respond to my comment here?  This opm-grid PR definitely requests CMake version 3.0 when it should request 3.10.

CMakeLists.txt
blattms: Thanks for pointing this out. Fixed,

blattms:
jenkins build this please

blattms:
BUild check successfull in OPM/opm-common#2386

PR #519, opened 2021-04-12
Report the global index cells which have been pinched out

joakim-hove:
To enable reporting of volume lost by pinching

### DIFF
--- a/opm/grid/MinpvProcessor.hpp
+++ b/opm/grid/MinpvProcessor.hpp
@@ -33,6 +33,20 @@ namespace Opm
     class MinpvProcessor
     {
     public:
+
+        struct PinchPair {
+            int cell1;
+            int cell2;
+            int removed_cell;
+
+            PinchPair(int c1, int c2, int rm) :
+                cell1(std::min(c1, c2)),
+                cell2(std::max(c1, c2)),
+                removed_cell(rm)
+            {}
+        };
+
+
         /// \brief Create a processor.
         /// \param[in]   nx   logical cartesian number of cells in I-direction
         /// \param[in]   ny   logical cartesian number of cells in J-direction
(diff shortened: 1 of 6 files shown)


joakim-hove:
jenkins build this please

bska (approved):
Looks good to me.  I'll merge into master.  When do you expect to use the new information computed here?

joakim-hove:
> When do you expect to use the new information computed here?

Now: https://github.com/OPM/opm-simulators/pull/3153

blattms:
I think his PR needs a fellow PR that documents the return value. As it is implemented only rank 0 wil return a non-empty vector, which might not be very obvious without documentation.

opm/grid/MinpvProcessor.hpp
```diff
@@ -205,7 +218,7 @@ namespace Opm
 
                             // Add a connection if the cell above and below is active and has porv > minpv
                             if ((actnum.empty() || (actnum[c_above] && actnum[c_below])) && pv[c_above] > minpvv[c_above] && pv[c_below] > minpvv[c_below]) {
-                                    nnc.insert(std::make_pair(c_above, c_below));
+                                    nnc.emplace_back(c_above, c_below, c);
```
blattms, line 221: The original code is complicated and it has been a long time since I read it carefully., hence I might be wrong:
My interpretation was that this algorithm might remove more than one active cell for each nnc added. But the downstream PR seems to assume that this cannot happen

joakim-hove:
>I think his PR needs a fellow PR that documents the return value

Coming.

opm/grid/MinpvProcessor.hpp
joakim-hove: > The original code is complicated 

You can say that!

> My interpretation was that this algorithm might remove more than one active cell for each nnc added

That is right - in the current code that will be multiple `PinchPair`objects with the same `cell1` and `cell2` value and different values for `removed_cell`. Eventually the `cell1, cell2` pairs are stored in a `std::set<>` which should handle the duplicates gracefully.

blattms: Well maybe I am wrong. What if the top-most cell (k==0) is pinched and hence no nnc created?

PR #522, opened 2021-04-19
Avoid include of Parser.hpp and Deck.hpp

joakim-hove:
The file `OpmParserIncludes.hpp` includes the header `parser/Parser.hpp` which transitively brings in half of opm-common. In particular it allowed for silent usage of `Deck` in: #517. 

With this PR only the "really needed" headers are included with the `OpmParserIncludes.hpp`. If this is merged #517 will need to explicitly include the dependencies from opm-common; which is in my opinion a good thing.

As an aside I would really like to reignite this: https://github.com/OPM/opm-grid/pull/401


joakim-hove:
jenkins build this please

bska:

PR #526, opened 2021-04-26
fixed: include required headers where needed

akva2:
Due to changes upstream.


akva2:
jenkins build this please

blattms (commented):
I am a it surprised. Shouldn't all additional includes be guarded by `#if HAVE_ECL_INPUT`, or is there no difference?

akva2:
uhm, tbh i am not sure. i assumed existing tests not having the guards was excluded on the build system level, but i did not confirm. in any case it is no worse with pr than it was before.

blattms:
Me neither. Merging.

PR #534, opened 2021-05-06
changed: put the EdgeWeightMethod enum in separate file

akva2:
to avoid having to pull in the entire grid class in downstreams


akva2:
jenkins build this please

PR #536, opened 2021-05-18
Remove explicit copy-constructor from class

akva2:
since it already has a explicitly cc, the implicit
assignment operator is deprecated. clang likes to
point this out.


akva2:
jenkins build this please

akva2:
jenkins build this please

PR #537, opened 2021-05-19
Add SubGridPart class and test.

atgeirr:
The SubGridView is intended for enabling local solves. By using a SubGridView instead of the full leaf grid view it will be possible to perform local assembly or other grid-based tasks with no or minimal code changes.

The test is essentially a copy of test_cpgrid.cpp, but it adds testing of iteration over a SubGridView. Furthermore, the test is also run with YaspGrid, to ensure SubGridView can be used with that as well.

I would appreciate any comments, in particular:
 - Is this a proper Dune GridView, i.e. is the concept followed properly?
 - In the iterator class, we store a mutable Entity, which may be a bit expensive for some grids. I could not see a simple alternative, as the construction of an Entity from an EntitySeed yields an object not a reference.

The dune-subgrid module was considered, but this is a more lightweight approach, and I consider it more appropriate for our needs since all the equation assembly code in opm-models is based on grid views.

### DIFF
--- a/opm/grid/CpGrid.hpp
+++ b/opm/grid/CpGrid.hpp
@@ -446,8 +446,8 @@ namespace Dune
 
         /// given an EntitySeed (or EntityPointer) return an entity object
         template <int codim>
-        cpgrid::Entity<codim> entity(const cpgrid::Entity< codim >&) const;
-           
+        cpgrid::Entity<codim> entity(const cpgrid::Entity<codim>& seed) const;
+
         /// @brief Create a grid out of a coarse one and a refinement(LGR) of a selected block-shaped patch of cells from that coarse grid.
         ///
         /// Level0 refers to the coarse grid, assumed to be this-> data_[0]. Level1 refers to the LGR (stored in this->data_[1]).
--- a/opm/grid/common/SubGridPart.hpp
+++ b/opm/grid/common/SubGridPart.hpp
@@ -100,6 +100,16 @@ struct SubGridPartTraits {
 };
 
 
+/// \brief A class to represent a part of a grid, similar to a GridView.
+///
+/// The differences from a GridView are:
+///   - The SubGridPart consists of a set of elements (codim 0 entities),
+///     considered to be Interior, and their neighbours, considered to be Overlap.
+///   - When iterating over intersections on the elements, and accessing the outside()
+///     elements, this can give you access to grid entities that are not in the
+///     SubGridPart itself. This can only happen for intersections in the Overlap part.
+///     For intersections of elements in the Interior part, the outside() element will
+///     be either Interior or Overlap.
 template <class GridImp>
 class SubGridPart
 {
--- a/opm/grid/common/SubGridPart.hpp
+++ b/opm/grid/common/SubGridPart.hpp
@@ -245,8 +245,8 @@ public:
         }
         subset_.resize(subset_.size() + unowned_neighbors.size());
         std::size_t count = num_owned_;
-        for (const auto& [index, seed] : unowned_neighbors) {
-            subset_[count] = seed;
+        for (const auto& neighbor : unowned_neighbors) {
+            subset_[count] = neighbor.second;
             ++count;
         }
         assert(count == subset_.size());
--- a/opm/grid/common/SubGridPart.hpp
+++ b/opm/grid/common/SubGridPart.hpp
@@ -8,8 +8,8 @@
 // Copyright 2021 Dune contributors.
 // Copyright 2021 SINTEF Digital, Mathematics and Cybernetics.
 
-#ifndef OPM_SUBGRIDVIEW_HEADER
-#define OPM_SUBGRIDVIEW_HEADER
+#ifndef OPM_SUBGRIDPART_HEADER
+#define OPM_SUBGRIDPART_HEADER
 
 #include <dune/common/exceptions.hh>
 #include <dune/common/typetraits.hh>
@@ -31,11 +31,11 @@ namespace Dune
 {
 
 template <class GridImp>
-class SubGridView;
+class SubGridPart;
 
 template <class GridImp>
-struct SubGridViewTraits {
-    using GridViewImp = SubGridView<GridImp>;
+struct SubGridPartTraits {
+    using GridPartImp = SubGridPart<GridImp>;
 
     /** \brief type of the grid */
     using Grid = typename std::remove_const<GridImp>::type;
@@ -101,12 +101,12 @@ struct SubGridViewTraits {
 
 
 template <class GridImp>
-class SubGridView
+class SubGridPart
 {
-    using ThisType = SubGridView<GridImp>;
+    using ThisType = SubGridPart<GridImp>;
 
 public:
-    using Traits = SubGridViewTraits<GridImp>;
+    using Traits = SubGridPartTraits<GridImp>;
 
     /** \brief type of the grid */
     using Grid = typename Traits::Grid;
@@ -131,7 +131,7 @@ public:
         class SubIterator
         {
         public:
-            SubIterator(const SubGridView& view, std::size_t index)
+            SubIterator(const SubGridPart& view, std::size_t index)
                 : view_(&view)
                 , index_(index)
             {
@@ -168,7 +168,7 @@ public:
                 return index_ != other.index_;
             }
         private:
-            const SubGridView* view_;
+            const SubGridPart* view_;
             std::size_t index_;
             mutable Entity entity_; // This may be low-performing for grids with large Entity objects.
         };
@@ -191,7 +191,7 @@ public:
     /// Construct a view of the codim 0 entities that can be constructed from the seeds input.
     ///
     /// The seeds input is moved from and will be in a valid but indeterminate state after the call.
-    SubGridView(const Grid& grid,
+    SubGridPart(const Grid& grid,
                 std::vector<typename Codim<0>::Entity::EntitySeed>&& seeds,
                 const bool overlap = true)
         : grid_(&grid)
@@ -269,7 +269,7 @@ public:
     template <int cd>
     typename Codim<cd>::Iterator begin() const
     {
-        static_assert(cd == 0, "Only codimension 0 iterators for SubGridView.");
+        static_assert(cd == 0, "Only codimension 0 iterators for SubGridPart.");
         using Iterator = typename Codim<cd>::Iterator;
         return Iterator(*this, 0);
     }
@@ -278,7 +278,7 @@ public:
     template <int cd>
     typename Codim<cd>::Iterator end() const
     {
-        static_assert(cd == 0, "Only codimension 0 iterators for SubGridView.");
+        static_assert(cd == 0, "Only codimension 0 iterators for SubGridPart.");
         using Iterator = typename Codim<cd>::Iterator;
         return Iterator(*this, subset_.size());
     }
@@ -289,7 +289,7 @@ public:
     template <int cd, PartitionIteratorType pit>
     typename Codim<cd>::template Partition<pit>::Iterator begin() const
     {
-        static_assert(cd == 0, "Only codimension 0 iterators for SubGridView.");
+        static_assert(cd == 0, "Only codimension 0 iterators for SubGridPart.");
         static_assert(pit == Interior_Partition || pit == Overlap_Partition || pit == All_Partition);
         if constexpr (pit == Interior_Partition || pit == All_Partition) {
             return begin<0>();
@@ -304,7 +304,7 @@ public:
     template <int cd, PartitionIteratorType pit>
     typename Codim<cd>::template Partition<pit>::Iterator end() const
     {
-        static_assert(cd == 0, "Only codimension 0 iterators for SubGridView.");
+        static_assert(cd == 0, "Only codimension 0 iterators for SubGridPart.");
         static_assert(pit == Interior_Partition || pit == Overlap_Partition || pit == All_Partition);
         if constexpr (pit == Overlap_Partition || pit == All_Partition) {
             return end<0>();
@@ -368,4 +368,4 @@ private:
 
 } // namespace Dune
 
-#endif // OPM_SUBGRIDVIEW_HEADER
+#endif // OPM_SUBGRIDPART_HEADER
--- a/opm/grid/common/SubGridView.hpp
+++ b/opm/grid/common/SubGridView.hpp
@@ -51,11 +51,39 @@ struct SubGridViewTraits {
     /** \brief type of the collective communication */
     using CollectiveCommunication = typename Grid ::Traits ::CollectiveCommunication;
 
+
+    template <class BaseEntityType>
+    class SubEntity : public BaseEntityType
+    {
+    public:
+        SubEntity()
+            : BaseEntityType()
+            , is_owned_(false)
+        {
+        }
+        SubEntity(const BaseEntityType& base, const bool owned)
+            : BaseEntityType(base)
+            , is_owned_(owned)
+        {
+        }
+        auto partitionType() const
+        {
+            if (is_owned_) {
+                return Dune::InteriorEntity;
+            } else {
+                return Dune::OverlapEntity;
+            }
+        }
+    private:
+        bool is_owned_;
+    };
+
     template <int cd>
     struct Codim {
         using BaseIterator = typename Grid ::Traits ::template Codim<cd>::template Partition<All_Partition>::LeafIterator;
 
-        using Entity = typename Grid ::Traits ::template Codim<cd>::Entity;
+        using BaseEntity = typename Grid ::Traits ::template Codim<cd>::Entity;
+        using Entity = SubEntity<BaseEntity>;
         using EntitySeed = typename Grid ::Traits ::template Codim<cd>::EntitySeed;
 
         using Geometry = typename Grid ::template Codim<cd>::Geometry;
@@ -95,6 +123,7 @@ public:
 
     /** \brief type of the collective communication */
     using CollectiveCommunication = typename Traits ::CollectiveCommunication;
+
 
     /** \brief Codim Structure */
     template <int cd>
@@ -194,11 +223,12 @@ public:
             }
         }
         subset_.resize(subset_.size() + unowned_neighbors.size());
-        int count = num_owned_;
+        std::size_t count = num_owned_;
         for (const auto& [index, seed] : unowned_neighbors) {
             subset_[count] = seed;
             ++count;
         }
+        assert(count == subset_.size());
     }
 
     /** \brief obtain a const reference to the underlying hierarchic grid */
@@ -303,13 +333,15 @@ public:
     }
 
 private:
-    typename Codim<0>::Entity get(std::size_t ii) const
+    using Entity0 = typename Codim<0>::Entity;
+    Entity0 get(std::size_t ii) const
     {
-        return grid_->entity(subset_[ii]);
+        const bool owned = ii < num_owned_;
+        return Entity0(grid_->entity(subset_[ii]), owned);
     }
     const Grid* grid_;
-    std::vector<typename Codim<0>::Entity::EntitySeed> subset_;
-    const int num_owned_;
+    std::vector<typename Entity0::EntitySeed> subset_;
+    const std::size_t num_owned_;
 };
 
 } // namespace Dune
--- a/opm/grid/common/SubGridView.hpp
+++ b/opm/grid/common/SubGridView.hpp
@@ -145,25 +145,6 @@ public:
     enum { dimension = GridImp::dimension };
 
 public:
-    /// Construct a view of the codim 0 entities enumerated in the indices vector.
-    ///
-    /// The indices vector is assumed to be sorted in ascending order.
-    SubGridView(const Grid& grid,
-                const std::vector<int>& indices)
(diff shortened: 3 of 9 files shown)


opm/grid/CpGrid.hpp
```diff
@@ -501,7 +501,7 @@ namespace Dune
         template <int codim>
         cpgrid::Entity<codim> entity( const cpgrid::EntityPointer< codim >& seed ) const
         {
-            return cpgrid::Entity<codim>( *seed );
+            return cpgrid::Entity<codim>( *(this->current_view_data_), *seed );
```
atgeirr, line 504: Note that this is a bugfix that was needed now but never triggered before.

atgeirr:
jenkins build this please

bska (commented):
Given the constraints of EntitySeeds with respect to Entities I think this is a good compromise.  I noticed a couple of missing headers and I'd prefer using `using` for type aliases instead of `typedef`.  Other than that, I don't really have any comments.  We should really think through whether we really need to provide the "grid + index" construction method though.

opm/grid/common/SubGridView.hpp
```diff
@@ -0,0 +1,290 @@
...
+        : grid_(&grid)
+    {
+        assert(std::is_sorted(indices.begin(), indices.end()));
+        subset_.resize(indices.size());
+        const auto beg = grid.template leafbegin<0>();
+        auto it = beg;
+        int previous = 0;
+        for (size_t c = 0; c < indices.size(); ++c) {
+            using EntitySeed = typename Codim<0>::Entity::EntitySeed;
+            std::advance(it, indices[c] - previous);
```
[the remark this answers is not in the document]
atgeirr: It is not that expensive, as it only steps the difference between the previous and next. But yes, it will make one iteration sweep of the entire grid to create one SubGridView. If all you need is a single such view, this constructor is as fast as it gets, the other constructor will need to get the seeds from a similar grid iteration sweep. If you want to construct many sub views, you should try to make all with a single sweep, and for that you need the other constructor. The question is if it is dangerous to keep this one? I am on the side of "not very".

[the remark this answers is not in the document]
bska: > If all you need is a single such view, this constructor is as fast as it gets, the other constructor will need to get the seeds from a similar grid iteration sweep.

True, but that's not really an interesting use case for a subgrid view.

> The question is if it is dangerous to keep this one? I am on the side of "not very".

It Depends[tm].  The mere existence of this constructor will, at least in some small way, encourage its use.  I'd rather that there be a single, canonical constructor and then applications can create free functions to form objects from other types if needed.

atgeirr:
I have now revised the branch, with all requested changes, including deleting the contentious constructor.

atgeirr:
jenkins build this please

bska (approved):
Thanks for the updates.  I think this looks good now.  I'll let @blattms have a look at potential Dune-related improvements, but otherwise this should be merged.

opm/grid/common/SubGridView.hpp
```diff
@@ -0,0 +1,274 @@
...
+            }
+            Iterator operator++(int)
+            {
+                Iterator copy(*this);
+                ++index_;
+                return copy;
+            }
+            bool operator==(const Iterator& other) const
+            {
+                assert(view_ == other.view_);
```
blattms, line 132: I think the assert here should simply be incorporated into the equality comparison. They should be cheap, anyway.

blattms (changes requested):
I think this class might have very surprising semantics to people assuming that this is a real grid view:
1. Communicate will communicate on all entities regardless whether they are part of the gridview or not.
2. It seems like using the intersection iterator you will reach neighbors that are not part of the grid view. This is certainly not ok for a DUNE gridview. I also wonder how your local assembly will behave here.
3. Throwing for some gridview methods  might be quite surprising.

As this is not really a grid view in the DUNE sense (because of above), I would choose another name or fix the issues. Even with another name 2) might become an issue unless worked around in your assembler code. It might give wrong boundary conditions in your assembler.

opm/grid/common/SubGridView.hpp
atgeirr: At first I agreed, but after thinking it over I have come to disagree. Comparing iterators from two different views is a programmer error, and not a normal part of the comparison. Therefore would not the assert() be more appropriate?

atgeirr:
> I think this class might have very surprising semantics to people assuming that this is a real grid view:
> 
> 1. Communicate will communicate on all entities regardless whether they are part of the gridview or not.
> 2. It seems like using the intersection iterator you will reach neighbors that are not part of the grid view. This is certainly not ok for a DUNE gridview. I also wonder how your local assembly will behave here.
> 3. Throwing for some gridview methods  might be quite surprising.
> 
> As this is not really a grid view in the DUNE sense (because of above), I would choose another name or fix the issues. Even with another name 2) might become an issue unless worked around in your assembler code. It might give wrong boundary conditions in your assembler.

Thanks for your comments! I agree with 1 and 2. Number 3 may be forgiven perhaps, but better then to delete the methods. I will make this a WIP PR until I have decided on the appropriate way to approach this.

blattms:
Well, if 1 and 3 are not a problem for your project, then you might just use another name to indicate that this is not a full grid view. If you use a normal assembler (without communication) than 2 is probably the only thing that that needs to be fixed.

For fixing 1 you will probably need some kind of Datahandle wrapper that does not disregard cells not in the grid view when reading and writing.

bska (commented):
I'll defer to @blattms to judge whether this meets the requirements of a Dune gridview.  In terms of the implementation of owned vs. not owned entities I noticed a couple of possible improvements.

opm/grid/common/SubGridView.hpp
```diff
@@ -0,0 +1,349 @@
...
+        using Iterator = typename Codim<cd>::Iterator;
+        return Iterator(*this, 0);
+    }
+
+    /** \brief obtain begin iterator for this view */
+    // template <int cd, PartitionIteratorType pit>
+    // typename Codim<cd>::template Partition<pit>::Iterator begin() const
+    // {
+    //     return grid().template leafbegin<cd, pit>();
+    // }
```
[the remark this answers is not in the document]
atgeirr: I do plan on enabling this, so I would like to keep it in for now. As to other methods that currently throw, I will remove them (but keep a commented-out declaration with documentation of why it is not implemented).

atgeirr:
I have addressed the issues mentioned by @bska, and also made the test program a proper test that will actually trigger a testing failure if messed with rather than just printing errors to cout.

To get back to @blattms' points:
1. Communication. The communicate() method has been removed. (Kept the commented-out declaration for reference.)
2. Reaching outside the view. Now the SubGridView makes a distinction between interior and overlap entities. I am not sure if this makes it actually closer to a grid class than a grid view, as the entity type is now different from the entity type of the underlying grid. You can create a SubGridView with a 1-cell overlap zone or not, and partitionType() as well as iterating over Overlap_Partition and Interior_Partition in addition to the default All_Partition. If you insist on iterating over intersections in the overlap zone you will still be able to reach outside the SubGridView.
3. No methods throw, I have instead deleted any nonfunctional methods.

I think the current code should be quite useful. However it may not be a proper grid view (missing methods, able to get noncontained entities via intersections), so I am ok with renaming it. Perhaps SubGridPart? Suggestions are welcome!

Apart from the name issue I think this code is ready for review. (I will also add a few extra tests.)

opm/grid/common/SubGridView.hpp
```diff
@@ -0,0 +1,349 @@
...
+            if (owned.count(nb.first) == 0) {
+                unowned_neighbors.insert(nb);
+            }
+        }
+        subset_.resize(subset_.size() + unowned_neighbors.size());
+        std::size_t count = num_owned_;
+        for (const auto& [index, seed] : unowned_neighbors) {
+            subset_[count] = seed;
+            ++count;
+        }
```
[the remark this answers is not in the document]
bska: Did you not want to do anything about this (default constructible entity seeds vs. `reserve()` and `push_back()`)?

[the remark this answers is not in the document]
atgeirr: They are default constructible, so I decided to keep the code as is. Sorry for not remarking on this.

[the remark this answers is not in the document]
bska: > They are default constructible, so I decided to keep the code as is.

Cool.  Thanks for the information.

opm/grid/common/SubGridView.hpp
```diff
@@ -0,0 +1,371 @@
...
```
bska, line 241: Come to think of if, this section could be reduced to
```C++
for (const auto& [index, seed] : neighbors) {
    if (owned.count(index) == 0) {
        this->subset_.push_back(seed);
    }
}
this->subset_.shrink_to_fit();
```

atgeirr: That would give a random order to the neighbours, I would like them to have the predictable order (same as global index order). It could also be important to maintain the "owned by process first" order within the subset overlap partition when we go to parallel, although that is not really thought through.

This is also why the unowned_neighbors is a map and not unordered_map.

opm/grid/common/SubGridView.hpp
```diff
@@ -0,0 +1,371 @@
...
+            const Entity& operator*() const
+            {
+                entity_ = this->view_->get(index_);
+                return entity_;
+            }
+            const Entity* operator->() const
+            {
+                entity_ = this->view_->get(index_);
+                return &entity_;
+            }
```
blattms, line 148: This seems very dangerous as another call the function will make the previously returned reference/pointer dangling.

opm/grid/common/SubGridView.hpp
blattms, line 241: Can't we simply do
```
subset_.reserve(subset_.size() + unowned_neighbors.size());
subset_.insert(subset_.end(), unowned_neighbors.begin(), unowned_neighbors.end());
```
to make this more explicit and less code?

opm/grid/common/SubGridView.hpp
```diff
@@ -0,0 +1,371 @@
...
+    {
+        assert(grid_);
+        return *grid_;
+    }
+
+    /** \brief obtain the index set */
+    // const IndexSet& indexSet() const // Not implemented
+
+    /** \brief obtain number of entities in a given codimension */
+    int size(int codim) const
```
blattms, line 256: Only correct for codim 0, but even CpGrid supports this for vertices

opm/grid/common/SubGridView.hpp
```diff
@@ -0,0 +1,371 @@
...
+    using Grid = typename std::remove_const<GridImp>::type;
+
+    /** \brief type of the index set */
+    using IndexSet = typename Grid ::Traits ::LeafIndexSet;
+
+    /** \brief type of the intersection */
+    using Intersection = typename Grid ::Traits ::LeafIntersection;
+
+    /** \brief type of the intersection iterator */
+    using IntersectionIterator = typename Grid ::Traits ::LeafIntersectionIterator;
```
blattms, line 50: Are we still accessing neighbors that are not in the subgrid?

blattms (changes requested):
It seems that iterating over the intersections of an entity is not contained in the subgrid as we seem to be able to end up at cells that are not part of the subgrid.

Please add a test that checks that we never reach an entity from an intersection.

If we do not support fully support all entities and partition types then this should at least be made explicit by the class name and documentation.

opm/grid/common/SubGridView.hpp
bska: > Can't we simply do

Not out of the box, no.  `unowned_neighbors` is a `map<int, EntitySeed>` whereas `subset_` is a `vector<EntitySeed>`.

opm/grid/common/SubGridView.hpp
```diff
@@ -0,0 +1,371 @@
...
+            : BaseEntityType()
+            , is_owned_(false)
+        {
+        }
+        SubEntity(const BaseEntityType& base, const bool owned)
+            : BaseEntityType(base)
+            , is_owned_(owned)
+        {
+        }
+        auto partitionType() const
```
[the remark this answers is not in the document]
atgeirr: Yes. The "view" only deals with codim 0, and since I do not see a need for vertices, I was not planning to add it.

opm/grid/common/SubGridView.hpp
atgeirr: It will not be dangling, but it will suddenly refer to a different object. I agree this is a bit dangerous, but I did not see any way around the requirement for the operator. Suggestions are welcome. For the `operator*()` I guess we could return a copy instead, but that is not possible for `operator->()`.

opm/grid/common/SubGridView.hpp
atgeirr: See previous answer.

opm/grid/common/SubGridView.hpp
atgeirr: Yes.

atgeirr:
> It seems that iterating over the intersections of an entity is not contained in the subgrid as we seem to be able to end up at cells that are not part of the subgrid.

Yes. However now that we are able to distinguish between Interior and Overlap, we can avoid this by not iterating over intersections of Overlap entities. Going any further will require much more I think, turning this class into something with a full topology implementation instead of using the underlying grid.

> Please add a test that checks that we never reach an entity from an intersection.

Since this is indeed possible, there is no point in checking that it is not.

> If we do not support fully support all entities and partition types then this should at least be made explicit by the class name and documentation.

I agree, and suggested the new class name SubGridPart. What do you think?

atgeirr:
> suggested the new class name SubGridPart

Or maybe PartialGridView? That shows that most of the API is following the GridView concept, but not in a complete way.

blattms:
I am a bit surprised about your comments concerning reaching entities not in the view via entities and your recommendation to use special iterators to prevent that. Just to see that I get this correctly:

- I cannot use the View with any generic assembly algorithm as I would reach entities outside of the view and I cannot get a proper index
-  If I want to do an assembly, then I have to (per your recommendation)
   + Create a view with the domain of interest and the domain needs to be the interior + the overlap cells.
   + In need to iterate over the interior cells and can assemble as usual but need to manually keep track of the index of the dof.
   + Then I need to iterate over the overlap cells and need to apply a special assembly as it addtiionally needs to treat intersections with entities outside the domain in a special way.

This seems rather complicated (as I still have to implement a special assembly algorithm). I also do not see a way to determine whether a cell reached is outside the view or not.

But I might just be too tight up in my way of assembling (e.g. for domain decomposition methods) or I might still not getting the use case/purpose of this. Maybe you can describe how your assembly algorithm would work?

I really think the cleanest way would be a custom intersection iterator. You would need to mark entities at the boundary and use special treatment there. Should not be too involved and would make this a lot more usable.
Now you see me a bit surprised. Isn't the purpose of a subgridview, to use it in generic assembling algorithms? (E.g. you can use a LevelGridView just like a LeafGridView because intersections will never b between entitiers of different level index). How is such an algorithm supposed to determine that an intersection is a boundary?

As long as one can reach outside entities from the intersection iterator, I would strongly advice against having GridView in the name as this against the main concept (e.g. a intersectionn from a LevelGridView will never reach entities on another level.).

GridPart seems O.K.

blattms:
Just a thought:

Since we have simulators working with dune-fem and that includes similar things, e.g. [FilteredGridPart](https://www.dune-project.org/doxygen/dune-fem/master/classDune_1_1Fem_1_1FilteredGridPart.html). I know it is fun to implement stuff yourself, but this implementation and maintenance might be more work than anticipated?

atgeirr:
> Since we have simulators working with dune-fem and that includes similar things, e.g. [FilteredGridPart](https://www.dune-project.org/doxygen/dune-fem/master/classDune_1_1Fem_1_1FilteredGridPart.html). I know it is fun to implement stuff yourself, but this implementation and maintenance might be more work than anticipated?

I will take a look. There is not much maintenance to keeping this open, rebasing the branch every now and then has been very uneventful, but I agree we should avoid reimplementing things already in existence.

blattms:
FYI: My dune-fem argument is moot for DUNE >= 2.9

atgeirr:
> FYI: My dune-fem argument is moot for DUNE >= 2.9

What was changed?

blattms:
OPM/opm-simulators#4170

bska:
jenkins build this please

bska:
I had to make the change below in order for this PR to build cleanly.  Probably fallout from the ongoing header cleanup work.  Please rebase onto current master and include the requisite header.

---

```diff
diff --git a/tests/test_subgridview.cpp b/tests/test_subgridview.cpp
index 49d8dca3..d3dfda75 100644
--- a/tests/test_subgridview.cpp
+++ b/tests/test_subgridview.cpp
@@ -33,6 +33,8 @@ using Dune::referenceElement; //grid check assume usage of Dune::Geometry
 // Re-enable warnings.
 #include <opm/grid/utility/platform_dependent/reenable_warnings.h>
 
+#include <opm/input/eclipse/EclipseState/Grid/EclipseGrid.hpp>
+
 #include <opm/input/eclipse/Deck/Deck.hpp>
 #include <opm/input/eclipse/Parser/Parser.hpp>
 ```

**Comment by bska** [comment]
jenkins build this please

**Comment by bska** [comment]
Tiny issue identified in the [CI build](https://ci.opm-project.org/job/opm-grid-PR-builder/331/gcc/new/).  Naming both [the test](https://github.com/OPM/opm-grid/blob/ca58ad3d7ef54de7a64029610bbb36b51892c99b/tests/test_subgridview.cpp#L182) and a [variable](https://github.com/OPM/opm-grid/blob/ca58ad3d7ef54de7a64029610bbb36b51892c99b/tests/test_subgridview.cpp#L202) `deck` in `test_subgridview.cpp` leads to the latter name shadowing the (generated) class/constructor name.  I suggest we rename the test to something like `From_Deck` or similar to avoid that issue.

**Comment by bska** [comment]
jenkins build this please

**Comment by bska** [comment]
jenkins build this please

**Comment by atgeirr** [comment]
I suggest merging this now. I have changed the name to SubGridPart, to make it clear that it is not a proper GridView, and added documentation explaining it, which I think should resolve the major criticism.

**Comment by atgeirr** [comment]
jenkins build this please

File: tests/test_subgridpart.cpp [hunk 11]
```diff
@@ -0,0 +1,250 @@
...
+2 2 2 /
+GRID
+DXV
+2*1 /
+DYV
+2*1 /
+DZ
+8*1 /
+TOPS
+8*100.0 /
```
bska, line 199: I probably doesn't matter a whole lot, but this `8 * 100` should **probably** be `4 * 100`, possibly followed by `4 * 101`.  As written, this `TOPS` vector does not represent a valid tesselation of the geometry.

opm/grid/common/SubGridPart.hpp
```diff
@@ -0,0 +1,381 @@
...
+        // Now that owned is complete, we can eliminate any owned entries.
+        std::map<int, Seed> unowned_neighbors;
+        for (const auto& nb : neighbors) {
+            if (owned.count(nb.first) == 0) {
+                unowned_neighbors.insert(nb);
+            }
+        }
+        subset_.resize(subset_.size() + unowned_neighbors.size());
+        std::size_t count = num_owned_;
+        for (const auto& [index, seed] : unowned_neighbors) {
```
bska, line 248: It appears that `index` is unused here.  At the moment, the CI system does **not** issue a diagnostic for that, but there's  no guarantee that that will remain true if/when we upgrade the system's compiler.  Although the syntax is less palatable this might be a case for using the pair directly:
```C++
for (auto&& neighbour : unowned_neighbours) {
    subset[count] = std::move(neighbour.second);
    ++count;
}
```

opm/grid/common/SubGridPart.hpp
```diff
@@ -0,0 +1,381 @@
...
+            }
+        }
+        // Now that owned is complete, we can eliminate any owned entries.
+        std::map<int, Seed> unowned_neighbors;
+        for (const auto& nb : neighbors) {
+            if (owned.count(nb.first) == 0) {
+                unowned_neighbors.insert(nb);
+            }
+        }
+        subset_.resize(subset_.size() + unowned_neighbors.size());
```
bska, line 246: `Resize()` assumes that `Seed` is default constructible and, typically, quick to assign.  While this is obviously true for `CpGrid`, do we know that the assumption always holds for all relevant grid types?  If not, `reserve()` followed by `push_back()` is more appropriate.

bska (commented):
I think this is okay, except for a couple of potential generalisations.

opm/grid/common/SubGridPart.hpp
atgeirr: I think this is true, as the EntitySeed was introduced to dune to have a lightweight easy-to-use representation of entities. Certainly this holds also for YaspGrid, which is tested here. I therefore suggest to let this stay as it is.

opm/grid/common/SubGridPart.hpp
atgeirr: Done (but without the move).

tests/test_subgridpart.cpp
atgeirr: Done.

PR #540, opened 2021-06-02
Fix NNC creation for horizontal faces with no connected cells.

blattms:
At least for the grid Hummocky in opm-upscaling there (lots) of faces with -1 for both connected cells. This resulted in valgrind
complaining about reading uninitialized values global_cell[fnc[0]] as fnc[0] was -1.

Hence we need to cater for this.

### DIFF
--- a/opm/grid/cpgrid/processEclipseFormat.cpp
+++ b/opm/grid/cpgrid/processEclipseFormat.cpp
@@ -865,17 +865,20 @@ namespace cpgrid
                     if (fnc[1] == -1 ) {
                         // Add the NNC created from the minpv processs
                         // at the bottom of the cell.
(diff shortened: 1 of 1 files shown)


blattms:
It seems a little strange that there are faces with no connected cells at all. Why is that?

blattms:
jenkins build this please

blattms:
The problem is there for Hummocky.grdecl from opm-upscaling.

atgeirr (approved):

atgeirr:
> It seems a little strange that there are faces with no connected cells at all. Why is that?

I do not have a good answer. Maybe the faces were created in a processing step, and then afterwards both adjoining cells were removed due to being degenerate?

PR #552, opened 2021-10-05
User OwnerOverlapCopyCommunicatio to encapsulate cell communication.

blattms:
This will allow to get rid of some copying in opm-simulators (read ParallelIstlInformation) eventually or at least to not add more copying for parallel tracer support.


blattms:
jenkins build this please

blattms:
jenkins build this please

blattms:
jenkins build this please

atgeirr:
Just trying to wrap my head around this a bit. I think that (starting from a very low level...):
 - MC = Dune::MPIHelper::MPICommunicator is a class that either is a dummy, or an actual MPI_Comm.
 - CC = Dune::CollectiveCommunication<MC> (without Collective from Dune 2.7) is a class that can be used to do mpi stuff without calling MPI, and does the correct (noop) thing if there is no MPI.
 - OOCC = Dune::OwnerOverlapCopyCommunication<GlobalId, LocalId> is a class that communicates data on cells (for example), between defined sets (overlap etc.) based on index sets, using a CC object.
 - PII = ParallelIstlInformation has a similar role as OOCC, but contains more options.
 - Each of the four above have been at some place in Dune or Opm code referred to using the template type parameter C or Comm...
 - You seek to get rid of part of the PII by using the OOCC instead, but then you need to give it more information when set up.

Please correct me if I am wrong, or not precise. In any case, I think your goal (as far as I understand it) seems good! I am not sure I understand the how and what of it though.

blattms:
Your description of MC/CC/OCC/PII seems correct.

I simply need an OOCC (with all the information about the cell communication) for the tracers and tried to understand what is actually happening with all these classes. The information that I need is already present in the grid and now exported for more easier usage.

It turns out that ParallelIstlInformation was nearly nearly unused (despite for copying information) and with this change we should be able to get rid of it (and hence also of copying the information already present in the grid and wasting space).
ParallelIstlInformation  holds pointers to the ParallleIndexSet and RemoteIndices in CpGrid. It is used to pass this information to the linear algebra (IstlSolverEbos). When we pass the information we actually create an OOCC there and then copy the information of the ParallleIndexSet and RemoteIndices there and use it for communication. We need the copy as OOCC owns its
members and is not under our control.

Why was ParallelIstlInformation there?
1. Some did not want a DUNE dependency in opm-core where it originally resided. (Hence it replicates OOCC and allows copying the communication information of opm-grid)
2. For the old Newton approach it's additional methods were used in the convergence checks to do less global reductions.. This is now done differently.
3. more historic baggage, maybe?

atgeirr:
Thanks for the explanation!

bska (commented):
I don't have a sufficiently deep understanding of the protocols to make an informed review so I just have one question concerning the differences between a
```C++
const CommunicationType&
```
and a regular (mutable) `CommunicationType&`.  When do you need the latter in a way that is not possible with the former?

atgeirr:

PR #555, opened 2021-10-25
Save/export information about Numerical Aquifer cells in CpGrid.

blattms:
This is done as a sorted range of active indices of aquifer cells. The information is communicated during loadbalance.

For parallel testing we use a data file from opm-tests.

Is there a way to deactivate the serial test? It is rather useless...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -128,7 +128,23 @@ if(MPI_FOUND)
 endif()
 
 if(MPI_FOUND AND HAVE_OPM_TESTS)
+  # opm_add_test would need a driver script to run the
+  # parallel test (tests/run-parallel-unitTest.sh which
+  # is in opm-simulators and needs to be moved) until
+  # then we use opm_add_test and an additional call to add_test
+  # to run the test in parall. ONLY_COMPILE does not work because
+  # somehow -DBOOST_TEST_DYN_LINK is skipped and there is no main.
+  opm_add_test(cpgrid_aquifer_test
+    SOURCES tests/cpgrid/cpgrid_aquifer_test.cpp
+    DEPENDS opmgrid
+    #ONLY_COMPILE
+    PROCESSORS 4
+    CONDITION MPI_FOUND AND Boost_UNIT_TEST_FRAMEWORK_FOUND
+    DRIVER_ARGS -np 4
+    TEST_ARGS -- ${OPM_TESTS_ROOT}/aquifer-num/3D_2AQU_NUM.DATA
+    LIBRARIES opmgrid ${Boost_LIBRARIES})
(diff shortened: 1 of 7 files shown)


ytelses:
Benchmark result overview: 




| Test    | Configuration                                     | Relative |
| ------- | ------------------------------------------------- | -------- |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1        | 1.011    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8        | 1.008    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1        | 0.981    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8        | 0.995    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 1 | 0.997    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 8 | 0.99     |



View result details @ https://www.ytelses.com/opm/?page=result&id=1394

blattms:
@akva2 That suggestion sounded very easy, but I seem to missing all necessary skills. How do I get the to run in parallel?
Using
```
  opm_add_test(cpgrid_aquifer_test
    SOURCES tests/cpgrid/cpgrid_aquifer_test.cpp
    DEPENDS opmgrid
    PROCESSORS 4
    CONDITION MPI_FOUND AND Boost_UNIT_TEST_FRAMEWORK_FOUND
    DRIVER_ARGS -np 4
    TEST_ARGS -- ${OPM_TESTS_ROOT}/aquifer-num/3D_2AQU_NUM.DATA
    LIBRARIES opmgrid ${Boost_UNIT_TEST_FRAMEWORK_LIBRARY})
``
Adding "DRIVER mpirun" at least uses mpirun but passes the binary with the '-e' option which mpirun does not support.
Is there an existing driver script for this?

akva2:
there is a driver but not in omp-grid. see https://github.com/OPM/opm-simulators/blob/master/CMakeLists.txt#L286 and https://github.com/OPM/opm-simulators/blob/master/tests/run-parallel-unitTest.sh

blattms:
Maybe we should move that to opm-common?

akva2:
yes probably. simply has been no need yet. you'd have to do some extra dancing to pull that off though (to pass it through the installed build chain), so nobody would object if you simply copy-pasta for now and then we'll deal with it separately.

blattms:
Well, then I can just use add_test directly instead. Seems less work.

blattms:
Ok, I tried this but the result a47717f  seems worse than what was there before.

totto82:
jenkins build this please

PR #564, opened 2022-01-03
Fix: unconnected wells give undefined behaviour in parallel.

atgeirr:
This situation would segfault (clang, macos) or fail silently (linux) with the current master. The line
```C++
int owner = no_connections_on_proc.begin()->first;
```
would bomb (or not) because the no_connections_on_proc map was empty.

### DIFF
--- a/opm/grid/common/WellConnections.cpp
+++ b/opm/grid/common/WellConnections.cpp
@@ -183,11 +183,16 @@ postProcessPartitioningForWells(std::vector<int>& parts,
 
         for (const auto &well : wells) {
             const auto &connections = well_connections[well_index];
+            if (connections.empty()) {
+                // No connections, nothing to move or worry about.
+                ++well_index;
+                continue;
+            }
             std::map<int, std::size_t> no_connections_on_proc;
             for (auto connection_index : connections) {
                 ++no_connections_on_proc[parts[connection_index]];
             }
-
+            assert(!no_connections_on_proc.empty());
             int owner = no_connections_on_proc.begin()->first;
 
             // \todo remove trigger code for #476 that moves all wells to the last rank


atgeirr:
jenkins build this please

atgeirr:
I'll self-merge this, as it is a simple and (almost) obvious fix.

blattms:
Coming to think of this (this morning) it might not be that simple (or it might).  I would like to understand why and when this happens.

If it is a model without UDQ/ActionX that defines a well without connections, this is a fix. Is that the case here?

The code assumes that `wells` (return value of `Schedule::getWellsatEnd()`) contains all wells with all possible perforations throughout the whole simulation. Hence the question is whether this still holds under all circumstances. It certainly did in the past before UDQ/ActionX. But what happens if a well is defined or changed via UDQ/ActionX? Will it show up as a normal well in the return value of `Schedule::getWellsatEnd()` at the start of the simulation or not? Maybe @joakim-hove can enlighten us.

atgeirr:
> If it is a model without UDQ/ActionX that defines a well without connections, this is a fix. Is that the case here?

It was in the case that triggered this bug for me. The whole well was ACTNUMed out. It was a model created by taking a larger model and disabling some parts the engineer was not interested in I think.

I am also curious as to what happens in the case you outline. We assume that we can build the sparse matrix only once, knowing the full connectivity pattern for the whole simulation ahead of time. I think the well connections only matter if we use `--matrix-add-well-contributions=true`, in that case having connections not seen with getWellsAtEnd() would likely fail. For the default behaviour it should not be a problem I think. (Or maybe hope?)

joakim-hove:
> But what happens if a well is defined or changed via UDQ/ActionX? Will it show up as a normal well in the return value of Schedule::getWellsatEnd() at the start of the simulation or not? 

I am sorry to say, but:

1. If a well is defined with `WELSPECS` in `ACTIONX` it will *not* show up in `Schedule::getWellsatEnd()` at the start of the simulation.
2. If a a well is modified with e.g. `COMPDAT` in `ACTIONX` the version returned from `Schedule::getWellsatEnd()` will not have the final set of connections.

We do inspect the the content of the keywords in the `ACTIONX` block briefly, so that for instance we now which wells will/can appear at a later stage. But only through name.

If we are to support this fully I would assume that the [SimulatorUpdate](https://github.com/OPM/opm-common/blob/master/opm/input/eclipse/Schedule/Action/SimulatorUpdate.hpp) mechanism could be extended with a method to inform the simulator that "Hey - now you need to redistribute the wells" - or something similar. I see a rich potential for confusion.

General comment: flow will internalize "everything" from the deck up front, and then start simulating. Eclipse (I think) does a first pass which validates the input, then afterwards it executes the deck as a reservoir model interpreter. I think the flow model is better, but the difference in execution model certainly hits us regularly, and `ACTIONX` is probably the main offender. Roughly one year ago (?) the `Schedule` class was significantly refactored, to allow a "replay" of the remaining part of the Schedule section when an `ACTIONX` keyword has evaluated to true - that refactoring has worked well for the Schedule data structure itself, but in the interaction between simulator and `Schedule` we still feel some pain due to the difference in execution model between flow and Eclipse.

For the rather exotic problem adressed here maybe a meaningful compromise would be to detect the situation, but not really handle it?

atgeirr:
Thanks @joakim-hove for the answer. Could we do some processing to get the union of all *possible* wells and connections upfront (something like playing all actionx keywords at the end), that we could use? Taking the end Schedule, then "play all actions"?

Alternatively, if we could find a good solution to get CPR performance without the well fill-in in the matrix that would eliminate the need for this. We have tried some ideas but not really found a great solution.

joakim-hove:
> Could we do some processing to get the union of all possible wells and connections upfront (something like playing all actionx keywords at the end), that we could use? Taking the end Schedule, then "play all actions"?

Well - they did get to the moon after all; so it should be possible. But I foresee a mess of special cases best avoided.

blattms:
> Alternatively, if we could find a good solution to get CPR performance without the well fill-in in the matrix that would eliminate the need for this.

I am sorry to say this, but I think we have a rather serious bug here and it will strike in any case (i.e. for distributed wells, too). CPR without fill-in will not eliminate anything as I highly doubt that appearing wells are correctly treated in parallel (if they cross processor borders). Not even sure whether it is restricted to wells. 

> For the rather exotic problem adressed here maybe a meaningful compromise would be to detect the situation, but not really handle it?

Even if some folks think it only strikes for exotic models, it will get used and apparently is. But as a first step that would be good.

> But I foresee a mess of special cases best avoided.

I foresee "mess" in any case case, let's just call it hard exiting work.

joakim-hove:
OK - some additional thoughts on the topic:

1. If I were to address this topic I would focus on extending the `SimulatorUpdate` mechanism mentioned above, rather than attempts of "playing all actions up-front". I think the former is to the greatest extent "going *with* the flow of the implementation".
2. It is valid add a well in an `ACTIONX` block, but most probably the same can be achieved by just opening an already defined well in the `ACTIONX` block - that already works perfectly.

atgeirr:
> 2\. It is valid add a well in an `ACTIONX` block, but most probably the same can be achieved by just opening an already defined well in the `ACTIONX` block - that already works perfectly.

I interpret this as being something the user can/must do in the deck. Could we make it so that all wells exist, and all connections, and the ACTIONX blocks that would create wells or connections instead open the well or connection? Then the user would not have to accommodate this?

joakim-hove:
> I interpret this as being something the user can/must do in the deck. 

Yes - that is right.

> Could we make it so that all wells exist, and all connections, and the ACTIONX blocks that would create wells or connections instead open the well or connection?

I can only repeat my comment about extending the `SimulatorUpdate` functionality which I think is the best alternative here.

PR #565, opened 2022-01-04
Removed unused variable.

blattms:
How I missed this for so long is beyond me. Shouldn't the compiler be warning about it?


blattms:

PR #567, opened 2022-02-15
[cmake] Cleanup prerequisites.

blattms:
Actually there is no direct boost dependency. Hence the one on date_time is not needed at all and the rest is already a dependency of opm-common. No need to list it here again.

### DIFF
--- a/opm-grid-prereqs.cmake
+++ b/opm-grid-prereqs.cmake
@@ -21,8 +21,6 @@ set (opm-grid_DEPS
   # compile with C99 support if available
   "C99"
   # various runtime library enhancements
-  "Boost 1.44.0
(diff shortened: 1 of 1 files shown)


blattms:
jenkins build this please

akva2:
If boost::test carries as a prereq on opm-common that is a bit worrysome but a separate issue.

akva2 (approved):

blattms:
> If boost::test carries as a prereq on opm-common that is a bit worrysome but a separate issue.

Well, I was also a bit surprised that boost::test was REQUIRED by opm-common. Is that what you mean?

akva2:
it used to be that boost::test was only searched for if there were unit tests, and not added as a dependency on the library itself. that's how it should be, there are no reason for requiring library users to pull in boost::test.

PR #568, opened 2022-02-17
Added partition method without grid distribution.

blattms:
This function is intended to create a domain decomposition into N domains that will be used to set up a block-Jacobi type smoother to be used on one process.

Cleaned up version of status in https://github.com/andrthu/opm-grid/tree/partition-without-scatter for better overview.
There seemed to only one relevant commit. Is that right, @andrthu?

### DIFF
--- a/opm/grid/CpGrid.hpp
+++ b/opm/grid/CpGrid.hpp
@@ -828,7 +828,7 @@ namespace Dune
         template<class DataHandle>
         bool loadBalance(DataHandle& data,
 #if DUNE_VERSION_NEWER(DUNE_COMMON, 2, 7)
-			 decltype(data.fixedSize(0,0)) overlapLayers=1, bool useZoltan = true)
+                         decltype(data.fixedSize(0,0)) overlapLayers=1, bool useZoltan = true)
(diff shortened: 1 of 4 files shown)


andrthu:
Yes, this is the only commit.

atgeirr:
If I understand correctly, calling zoltanGraphPartitionGridForJac() will not change anything, but return a vector with one element per cell, each put into a partition from 0 to (numPart - 1)?

When in parallel: Can I call this only on rank 0, before load balancing, or can I call this on any rank after the grid has been distributed to get a sub-partitioning?

This will be useful for ASPIN and similar domain decomposition methods!

blattms:
Seems like the graph uses the grid interface. Hence before loadbalancing
it will use the global grid, and after the local one. You can change that
for the load balanced one by copying CpGrid and switching the view.

It uses ZoltanSerialPartitioner. You implemented that. I would assume that it
runs Zoltan on one process. The question is whether there is a hardcoded check
if the rank is 0 somewhere. I guess you need to dive into the code.

blattms:

PR #569, opened 2022-02-21
remove cellGeomTypes

totto82:
The cellGeomTypes seems to create more problems than it solves. I suggest to remove them for now.

### DIFF
--- a/opm/grid/polyhedralgrid/grid.hh
+++ b/opm/grid/polyhedralgrid/grid.hh
@@ -1142,16 +1142,8 @@ namespace Dune
     {
       if( Seed::codimension == 0 )
       {
-        if( cellGeomTypes_.empty() )
-        {
-          assert( geomTypes( Seed::codimension ).size() == 1 );
-          return geomTypes( Seed::codimension )[ 0 ];
-        }
-        else
-        {
-          assert( static_cast<size_t>(seed.index()) < cellGeomTypes_.size() );
-          return cellGeomTypes_[ seed.index() ];
-        }
+        assert(!geomTypes( Seed::codimension ).empty());
+        return geomTypes( Seed::codimension )[ 0 ];
       }
       else
       {
@@ -1170,6 +1162,7 @@ namespace Dune
         }
 
         // for codim 2 and codim 3 there is only one option
+        assert(!geomTypes( Seed::codimension ).empty());
         return geomTypes( Seed::codimension )[ 0 ];
       }
     }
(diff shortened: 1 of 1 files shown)


totto82:
jenkins build this please

blattms:
Please give a better explanation:
- When does the problem ocurr
- What asert fails or what else goes wrong.

Whith the current description it is hard to see what the problems are and whether the fix is sound.

totto82:
My aim is to be able to run the smeaheia model from the CO2dataShare using simulators in opm-models with polyhedral grid. If GeomTypes is set to something other than `none` the ordering of nodes and possibly the connection map is altered deep inside the dune code in a way that lead to issues when solving the linear system. With this change GeomTypes `cube `and  `simplex `is only used when all cells are simplex or cubes. With this PR and https://github.com/OPM/opm-models/pull/683 the smeaheia model runs fine using simulators in opm-models combined with polyhedral grid.

blattms:
I think we should investigate this a little more to understand where the problem is in the algorithms. It might be that our code is relying on a specific ordering (index set) instead of using mappers as it should be. If that is the case we should also (or rather) fix that as it will reoccur with other grids that support multiple geometry types. What executable are you using in opm-models to run the simulation?

totto82:
> What executable are you using in opm-models to run the simulation?

@trinemykk is currently preparing a version that could be added to the tests in opm-models. I suggest we continue the discussion when the actual model is ready.

dr-robertk:
This looks reasonable. I think the general problem is the vertex numbering and the simplex or cube types never really worked, if I remember correctly. In the long run it would be nice to have those. In the short run this fix is fine.

opm/grid/polyhedralgrid/grid.hh
```diff
@@ -1142,16 +1142,7 @@ namespace Dune
...
-        {
-          assert( geomTypes( Seed::codimension ).size() == 1 );
-          return geomTypes( Seed::codimension )[ 0 ];
-        }
-        else
-        {
-          assert( static_cast<size_t>(seed.index()) < cellGeomTypes_.size() );
-          return cellGeomTypes_[ seed.index() ];
-        }
+        return geomTypes( Seed::codimension )[ 0 ];
```
blattms, line 1145: Shouldn't we keep an assert like `assert( geomTypes( Seed::codimension ).size() == 1 );` or `assert(!geomTypes( Seed::codimension ).empty());`

totto82: I guess so. Will update.

totto82:
jenkins build this please

totto82:
> In the short run this fix is fine

I take this as an approval. @blattms is this ok from your perspective as well?

totto82:
jenkins build this please

PR #570, opened 2022-03-02
[mpi] Never increment end iterator in gatherv.

blattms:
Even if some do not care (but should) as we never decrement, it is undefined behavior according to the STL. Hence let's do proper code.

Kudos to @bska for the hint. (Learning never stops.)


blattms:
jenkins build this please

blattms:
if-branch is indeed more readable. Will change.

blattms:
jenkins build this please

bska (approved):
Thanks.  This looks good to me and the build check is green.  I'll merge into master.

PR #572, opened 2022-03-25
Towards implementing LGR - Allow multiple grids in CpGrid

verveerpj:
Taking small steps toward implementing local grid refinement: Allow storing multiple `CpGridData` objects in a `CpGrid` object.

### DIFF
--- a/opm/grid/CpGrid.hpp
+++ b/opm/grid/CpGrid.hpp
@@ -946,7 +946,7 @@ namespace Dune
 
 
         const std::vector<double>& zcornData() const {
-            return data_->zcornData();
+            return current_view_data_->zcornData();
         }
 
 
@@ -1399,9 +1399,9 @@ namespace Dune
         void scatterData(DataHandle& handle) const
         {
 #if HAVE_MPI
-            if(!distributed_data_)
+            if(distributed_data_.empty())
                 OPM_THROW(std::runtime_error, "Moving Data only allowed with a load balanced grid!");
-            distributed_data_->scatterData(handle, data_.get(), distributed_data_.get(), cellScatterGatherInterface(),
+            distributed_data_[0]->scatterData(handle, data_[0].get(), distributed_data_[0].get(), cellScatterGatherInterface(),
                                            pointScatterGatherInterface());
 #else
             // Suppress warnings for unused argument.
@@ -1419,9 +1419,9 @@ namespace Dune
         void gatherData(DataHandle& handle) const
         {
 #if HAVE_MPI
-            if(!distributed_data_)
+            if(distributed_data_.empty())
                 OPM_THROW(std::runtime_error, "Moving Data only allowed with a load balance grid!");
-            distributed_data_->gatherData(handle, data_.get(), distributed_data_.get());
+            distributed_data_[0]->gatherData(handle, data_[0].get(), distributed_data_[0].get());
 #else
             // Suppress warnings for unused argument.
             (void) handle;
@@ -1486,15 +1486,15 @@ namespace Dune
         /// \brief Switch to the global view.
         void switchToGlobalView()
         {
-            current_view_data_=data_.get();
+            current_view_data_=data_[0].get();
         }
 
         /// \brief Switch to the distributed view.
         void switchToDistributedView()
         {
-            if (! distributed_data_)
+            if (distributed_data_.empty())
                 OPM_THROW(std::logic_error, "No distributed view available in grid");
-            current_view_data_=distributed_data_.get();
+            current_view_data_=distributed_data_[0].get();
         }
         //@}
 
@@ -1584,13 +1584,13 @@ namespace Dune
 
         /** @brief The data stored in the grid.
          *
-         * All the data of the grid is stored there and
-         * calls are forwarded to it.*/
-        std::shared_ptr<cpgrid::CpGridData> data_;
+         * All the data of all grids are stored there and
+         * calls are forwarded to relevant grid.*/
+        std::vector<std::shared_ptr<cpgrid::CpGridData>> data_;
         /** @brief A pointer to data of the current View. */
         cpgrid::CpGridData* current_view_data_;
         /** @brief The data stored for the distributed grid. */
-        std::shared_ptr<cpgrid::CpGridData> distributed_data_;
+        std::vector<std::shared_ptr<cpgrid::CpGridData>> distributed_data_;
         /**
          * @brief Interface for scattering and gathering cell data.
          *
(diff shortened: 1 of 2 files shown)


opm/grid/CpGrid.hpp
```diff
@@ -937,7 +937,7 @@ namespace Dune
 
 
         const std::vector<double>& zcornData() const {
-            return data_->zcornData();
+            return data_[0]->zcornData();
```
blattms, line 940: I think you discovered a bug here. This should probably have operated on current_view_data_ .  should probably check whether this function is used.

blattms (commented):
Let's think a bit about how this well work in the end:
- We read the grid without LGR info `data_[0]`
- The we add each LGR to its own level `data[i]_`
- Then we might do a loadbalance on `data_[0]`
- The we will need to add the LGRs to `distributed_data_[i]` on the relevant processes.

Seems like in the long run both the `distributed_view_` and `current_view_` will need to have the same type as `data_` as we need the whole grid hierarchy.

Maybe we should add a data member `current_leaf_view_data` that holds `current_view_data_[0]` and operate on that?

blattms:
jenkins build this please

opm/grid/CpGrid.hpp
verveerpj: It seems to be only used in `createEclipseGrid` in `GridHelpers.cpp`.

verveerpj:
> Let's think a bit about how this well work in the end:
> 
> * We read the grid without LGR info data_[0]

That should trivially work when we replace all access to `data_` with `data[0]`, right?

> * The we add each LGR to its own level data[i]_

So this would work similar to intialization of the coarse grid, I assume? So it should be possible to re-use code for that?

> * Then we might do a loadbalance on data_[0]
> * The we will need to add the LGRs to distributed_data_[i] on the relevant processes.
> 
> Seems like in the long run both the distributed_view_ and current_view_ will need to have the same type as data_ as we need the whole grid hierarchy.

I will look into making those vectors too already.

> 
> Maybe we should add a data member current_leaf_view_data that holds current_view_data_[0] and operate on that?

So, the simulation would proceed by accessing the grid via `current_leaf_view_data` instead of `current_view_data_`?

Apologies if my questions are very simple-minded, still trying to wrap my head properly about all concepts...

verveerpj:
@blattms Should I be seeing the Jenkins test results? I saw you asked Jenkins to run, but I do not see any test results.

akva2:
results are attached to commit id's. as you force-pushed, the result got lost.. the old build can be found at https://ci.opm-project.org/job/opm-grid-PR-builder/257/

akva2:
jenkins build this please

verveerpj:
Jenkins, build this please

verveerpj:
I am guessing I need some form of permission to start the Jenkins tests?

akva2:
jenkins build this please

akva2:
Will whitelist you tmrw, too clumsy on the phone.

blattms:
> > Let's think a bit about how this well work in the end:
> > 
> > * We read the grid without LGR info data_[0]
> 
> That should trivially work when we replace all access to `data_` with `data[0]`, right?
> 

I do think so. 

> > * The we add each LGR to its own level data[i]_
> 
> So this would work similar to intialization of the coarse grid, I assume? So it should be possible to re-use code for that?
> 

I could be wong, but in my mind this works differently. To me an LGR identifies a group of cells (A rectangular group in the underlying cartesian grid) and rule how to refine these. Hence the levels will be constructed by refinement. I think this is the
las step of the implementation.

> > * Then we might do a loadbalance on data_[0]
> > * The we will need to add the LGRs to distributed_data_[i] on the relevant processes.
> > 
> > Seems like in the long run both the distributed_view_ and current_view_ will need to have the same type as data_ as we need the whole grid hierarchy.
> 
> I will look into making those vectors too already.
> 
> > Maybe we should add a data member current_leaf_view_data that holds current_view_data_[0] and operate on that?
> 
> So, the simulation would proceed by accessing the grid via `current_leaf_view_data` instead of `current_view_data_`, and that would eventually contain the leave (the refined grid)?
> 

Yes, exactly. To the algorithm it looks like a normal grid.
Of course output might need to use the level information. But that is also for a later step.

> Apologies if my questions are very simple-minded, still trying to wrap my head properly about all concepts and terms...

No worries. Personally, I think these are the best questions. Often they tend to showcase problems/issues, that people with deep insight never would realize themselves.

verveerpj:
> > So this would work similar to intialization of the coarse grid, I assume? So it should be possible to re-use code for that?
> 
> I could be wong, but in my mind this works differently. To me an LGR identifies a group of cells (A rectangular group in the underlying cartesian grid) and rule how to refine these. Hence the levels will be constructed by refinement. I think this is the las step of the implementation.

That is then a point where I am not entirely clear on how this will function. Maybe we can have a chat about that.

blattms:
Just send me an email if you want a meeting.

For now I propose to do this step by step. We should get this in once flow works as before with the changes. Just summon jenkins once you think it is or tell us.

Then we prepare for the entities for adaptation, which should also not interfere with the current usage of flow

verveerpj:
jenkins build this please

verveerpj:
> For now I propose to do this step by step. We should get this in once flow works as before with the changes. Just summon jenkins once you think it is or tell us.

This passes the tests, so presumably these changes would not affect flow as it is.

blattms:
In that case, please undraft this and I will get it merged.

verveerpj:
> In that case, please undraft this and I will get it merged.

Done.

As a side note: I see that the merge pull request button is green for me, suggesting that I could merge it. Should that be? Is there not a policy that there should be an approved review?

akva2:
it's green but it won't work.

verveerpj:
> it's green but it won't work.

That is a good thing...

blattms:
by "I will get it merged" I meant that I will check it and will merge if it works.

verveerpj:
> by "I will get it merged" I meant that I will check it and will merge if it works.

I understood that, just was surprised by the green merge button :smile:

blattms:
benchmark please

ytelses:
Benchmark result overview: 




| Test    | Configuration                                     | Relative |
| ------- | ------------------------------------------------- | -------- |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1        | 1.01     |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8        | 1.012    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1        | 1.005    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8        | 0.991    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 1 | 1.006    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 8 | 0.994    |



View result details @ https://www.ytelses.com/opm/?page=result&id=1583

PR #576, opened 2022-04-08
LGR intial steps - Add bottom argument to createCartesian

verveerpj:
Adds a `bottom` argument with a default value equal to zero. Will be used to create refined grids.

### DIFF
--- a/opm/grid/CpGrid.hpp
+++ b/opm/grid/CpGrid.hpp
@@ -305,8 +305,10 @@ namespace Dune
         /// Create a cartesian grid.
         /// \param dims the number of cells in each cartesian direction.
         /// \param cellsize the size of each cell in each dimension.
+        /// \param bottom bottom coordinate of the grid
         void createCartesian(const std::array<int, 3>& dims,
-                             const std::array<double, 3>& cellsize);
+                             const std::array<double, 3>& cellsize,
+                             const std::array<double, 3>& bottom = {0.0, 0.0, 0.0});
 
         /// The logical cartesian size of the global grid.
         /// This function is not part of the Dune grid interface,
--- a/opm/grid/cpgrid/CpGrid.cpp
+++ b/opm/grid/cpgrid/CpGrid.cpp
@@ -467,7 +467,8 @@ CpGrid::scatterGrid(EdgeWeightMethod method,
 
 
     void CpGrid::createCartesian(const std::array<int, 3>& dims,
-                                 const std::array<double, 3>& cellsize)
+                                 const std::array<double, 3>& cellsize,
+                                 const std::array<double, 3>& bottom)
     {
         if ( current_view_data_->ccobj_.rank() != 0 )
         {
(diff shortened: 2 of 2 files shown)


blattms:
The assumption that bottom is the same for every direction might be a bit to restricting. Maybe that should be of type `std::array<double, 3>`?

verveerpj:
> The assumption that bottom is the same for every direction might be a bit to restricting. Maybe that should be of type `std::array<double, 3>`?

I am not sure I understand how that would work. The bottom refers to the z coordinate only, right?

blattms:
oottom` was probably misleading.
With your changes y and y still always start with zero in your case. I meant something like this:
```
std::array<double, 3> origin = { .... };
....
double pillar[6] = { origin[0] + x, origin[1]+ y, origin[2], origin[0] + x, origin[1] + y, top };
....
```
Even that is probably not yet general enough, as (for LGR) the pillars might not be vertical and to the very least not start/end at
the same height. For the final use case the grid is really spanned by at rather arbitrary points.

blattms:
Actually more like
```
double pillar[6] = { bottom[0] + x_bot, bottom[1]+ y_bot, bottom[2]+z_bot, top[0] + x_top, top[1] + y_top, top[2]+z_top };
```
Of course there might be restrictions on pillars allowed in the input format we use, that we might violate here.

verveerpj:
> oottom` was probably misleading. With your changes y and y still always start with zero in your case. I meant something like this:
> 
> ```
> std::array<double, 3> origin = { .... };
> ....
> double pillar[6] = { origin[0] + x, origin[1]+ y, origin[2], origin[0] + x, origin[1] + y, top };
> ....
> ```
 
ah, now I understand. the refined grid is of course also located at an offset laterally.

> Even that is probably not yet general enough, as (for LGR) the pillars might not be vertical and to the very least not start/end at the same height. For the final use case the grid is really spanned by at rather arbitrary points.

Yes, but then we will need a bit more fancy function...

verveerpj:
> 

does the `top` vector not follow from `dims `and `cellsize`?

> Actually more like
> 
> ```
> double pillar[6] = { bottom[0] + x_bot, bottom[1]+ y_bot, bottom[2]+z_bot, top[0] + x_top, top[1] + y_top, top[2]+z_top };
> ```
> 
> Of course there might be restrictions on pillars allowed in the input format we use, that we might violate here.

that would cover the case of pillars that are not straight but slanted. The original function did not seem to do that. Do we need that?

blattms:
Well, I think the original code assumes that the top plane has the same size as the lower one (same for left/right, front/back).  What I was hinting at is that this is not general enough. But like I said I cannot tell what the current underlying algorithms allow. You need to test and may raise the bar iteratively.

verveerpj:
> Well, I think the original code assumes that the top plane has the same size as the lower one (same for left/right, front/back). What I was hinting at is that this is not general enough. But like I said I cannot tell what the current underlying algorithms allow. You need to test and may raise the bar iteratively

> > 
> 
> does the `top` vector not follow from `dims `and `cellsize`?
> 
> > Actually more like
> > ```
> > double pillar[6] = { bottom[0] + x_bot, bottom[1]+ y_bot, bottom[2]+z_bot, top[0] + x_top, top[1] + y_top, top[2]+z_top };
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Of course there might be restrictions on pillars allowed in the input format we use, that we might violate here.
> 
> that would cover the case of pillars that are not straight but slanted. The original function did not seem to do that. Do we need that?

Actually, would `top` not be defined by `bottom + dims * cellsize`, considering that we define an integer number of cells of fixed size?

verveerpj:
Jenkins, build this please

blattms:
Mmh, the commit history looks  a bit messed up now. Not sure what you did. Maybe you can rebase it onto current master to only see the commits you added?

verveerpj:
> Mmh, the commit history looks a bit messed up now. Not sure what you did. Maybe you can rebase it onto current master to only see the commits you added?

Yes, it seems something went wrong during a rebase, did it again.

verveerpj:
Closing this, since we are taking a different direction.

PR #579, opened 2022-04-27
Changed encoding of header files to UTF-8

blattms:
Should fix lintian warning about national-encoding.

Note that Debian is in the process of phasing out support for national
encodings.


blattms:
Fell through the cracks before.

blattms:
thanks.

PR #586, opened 2022-09-16
added: option for externally supplied zoltan parameters

akva2:


akva2:
jenkins build this opm-simulators=4117 please

akva2:
jenkins build this opm-simulators=4117 please

opm/grid/common/ZoltanPartition.cpp
```diff
@@ -536,6 +545,7 @@ class ZoltanSerialPartitioner
     std::vector<ZoltanId> importGlobalGidsVector;
     bool allowDistributedWells;
     int numParts;
+    const std::map<std::string,std::string>& params;
```
bska, line 548: Hm, another reference data member.  We already have those so it's not really a change in semantics, but I guess we need to be **really** careful about lifetimes for these partitioner objects.

akva2: class is only available inside this cpp file and lifetimes are local to a few calls so it is not really an issue. i can also make copies if you want, as these maps will be trivially sized.

bska: > class is only available inside this cpp file and lifetimes are local to a few calls so it is not really an issue.

Ah, right&mdash;I didn't notice that.  Then this isn't a problem.

> can also make copies if you want, as these maps will be trivially sized.

No, there's no need to do that.  By the way, since you're already here, would you mind removing one of the [duplicate assignments](https://github.com/OPM/opm-grid/blob/9ebdd8b52f629e5ca7edd04fb0a91c2a79496a6a/opm/grid/common/ZoltanPartition.cpp#L477-L478) of the `NUM_GLOBAL_PARTS` parameter?  I **guess** that's a leftover from resolving an earlier merge conflict.

akva2: sure, amended.

PR #591, opened 2022-10-18
fix dune 2.6 without mpi

akva2:
As usual I forgot to test the serial build :/


akva2:
shameful self-merge to avoid failing post builder.

PR #588, opened 2022-10-11
Complete refined geometry of a single cell.

aritorto:
Refined face geometries added. Refined corner numbering consistent with
cpgrid. Small corrections on corner indices and addition of face
indices in the construction of refined cells. Extended check for the eight
refined corners that coincide with the parent cell corners.

Next steps:
1. To include topology of the refined cell.
2. To refine a patch of cells.

PR #594, opened 2022-10-20
Topology of refined cells and faces added

aritorto:
We added information about the faces of a cell, points of a face, cells attached to a face, face tags, and face normals.
Based on and includes PR #588

### DIFF
--- a/opm/grid/CpGrid.hpp
+++ b/opm/grid/CpGrid.hpp
@@ -80,6 +80,9 @@
 #include <list>
 #endif
 
+void refine_and_check(const Dune::cpgrid::Geometry<3, 3>&,
+                      const std::array<int, 3>&,
+                      bool);
 namespace Dune
 {
 
@@ -211,6 +214,10 @@ namespace Dune
         friend class cpgrid::CpGridData;
         template<int dim>
         friend cpgrid::Entity<dim> createEntity(const CpGrid&,int,bool);
+        friend
+        void ::refine_and_check(const Dune::cpgrid::Geometry<3, 3>&,
+                                const std::array<int, 3>&,
+                                bool);
 
     public:
 
--- a/opm/grid/CpGrid.hpp
+++ b/opm/grid/CpGrid.hpp
@@ -80,9 +80,6 @@
 #include <list>
 #endif
 
-void refine_and_check(const Dune::cpgrid::Geometry<3, 3>&,
-                      const std::array<int, 3>&,
-                      bool);
 namespace Dune
 {
 
--- a/opm/grid/cpgrid/CpGridData.hpp
+++ b/opm/grid/cpgrid/CpGridData.hpp
@@ -503,6 +503,17 @@ private:
     /** @brief The indicator of the partition type of the entities */
     PartitionTypeIndicator* partition_type_indicator_;
 
+    /*// Refine a patch of cells.
+    void refine_patch(const std::array<int,3>& cells_per_dim,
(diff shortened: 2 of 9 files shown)


blattms:
jenkins build this please

blattms:
jenkins build this please

bska:
I haven't looked at this in any way, but I notice that Jenkins [reports a new warning](https://ci.opm-project.org/job/opm-grid-PR-builder/288/gcc/new/). The `i` counter in the [geometry refinement loop](https://github.com/OPM/opm-grid/blob/473c250c9fdb397395ecf10aef0a8b52c86af311/tests/cpgrid/geometry_test.cpp#L429-L431) is being shadowed in an inner [geometry check](https://github.com/OPM/opm-grid/blob/473c250c9fdb397395ecf10aef0a8b52c86af311/tests/cpgrid/geometry_test.cpp#L494-L498).  The easiest remedy may just be to put the outer `i` definition into a nested scope.

blattms:
Argh, and of course the code with the warning is from me. :see_no_evil: How hard can it be... Fix is coming.

opm/grid/CpGrid.hpp
```diff
@@ -80,6 +80,9 @@
 #include <list>
 #endif
 
+void refine_and_check(const Dune::cpgrid::Geometry<3, 3>&,
+                      const std::array<int, 3>&,
+                      bool);
```
bska, line 85: Hm, a new function in the global namespace?  Do we really need that?

bska (commented):
I think this is a good first step.  We can maybe revise some of the indexing expressions at a later time, but I'd be happy to merge this as is.

The only thing that really caught my eye is the new `refine_and_check()` function in the global namespace.  Is there no other way to accomplish that effect, for instance something like
```C++
namespace Opm { namespace GeometryHelpers {
    void refine_and_check(...);
}} // namespace Opm::GeometryHelpers
```

opm/grid/CpGrid.hpp
blattms: Thanks for noticing
I guess I have introduced this. This function is defined in tests/cpgrid/geometry_test.cpp. As we need to access private members there, we need it to be a friend. Hence the forward declaration here. Now that I reread this I am wondering whether the declaration is needed or the friend declaration inside of CpGrid suffices. I am always struggeling with friends if they come together with template parameters.

blattms: Indeed, this forward declaration is unnecessary and we should remove it,

blattms:
jenkins build this please

aritorto:
The changes of my last commit are:
- In CpGrid.hpp: removing lines 83-85 due to unnecessary forward declaration.
- In Geometry.hpp: 
Variable names have been changed to unify and simplify notation: new variables are called "refined_*", e.g. "refined_faces", "refined_cell_to_point", etc. 
The container "global_refined_face4corners_indices_storage" has been removed since its content is basically in "refined_face_to_point". 
Change of lines 924-931 arises from deleting  "global_refined_face4corners_indices_storage" , therefore those lines consist in a new way to construct the same object using "refined_face_to_point" instead of the (old and) deleted "global_refined_face4corners_indices_storage" .

blattms:

blattms:

blattms:
jenkins build this please

bska (approved):
Thanks a lot for the updates and the stamina in getting the PR ready for merging.  I'll merge this into the master branch now.  Great Work[:tm:] both of you!

PR #595, opened 2022-10-21
Checks for the refined topoloy

blattms:
This is based on #594 and its purpose to more thoroughly check the implemented refinement for one cell. We do that for the case where we refine an axis-parallel cell. For the check we build an equivalent cartesian grid to get the geometry of the other codims and the topology for free. Then we compare the data produced by the refinement with this one.

Note that we do not insist on the exact same numbering of the face as we hope that OPM never would rely on this.

Marked as draft as the other PRs should probably be merged first and we might want to squash the commits of this one.

### DIFF
--- a/opm/grid/CpGrid.hpp
+++ b/opm/grid/CpGrid.hpp
@@ -80,6 +80,9 @@
 #include <list>
 #endif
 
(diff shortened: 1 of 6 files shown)


blattms:
jenkins build this please

blattms:
Fixed version is part of another PR now. Closing.

PR #596, opened 2022-10-24
Allow User-Defined Message Tag in P2PCommunicator::exchange

bska:
That way we're not necessarily restricted to MPI_COMM_WORLD since we don't rely on a [static variable](https://github.com/OPM/opm-grid/blob/b4b7971e4eaaff4830b9ea5254dc39748092f93d/opm/grid/common/p2pcommunicator.hh#L252) to [create message tags](https://github.com/OPM/opm-grid/blob/b4b7971e4eaaff4830b9ea5254dc39748092f93d/opm/grid/common/p2pcommunicator.hh#L255) for each new message.

The first use case for this is [gathering connection information](https://github.com/OPM/opm-simulators/blob/cf6423bfce2fe7b482fee6a02ef5f5aa5329d382/opm/simulators/wells/WellState.cpp#L316) for distributed wells [light of `data::Connection` having variable-length tracer solution components](https://github.com/OPM/opm-common/issues/2921#issuecomment-1285496321).

### DIFF
--- a/opm/grid/common/p2pcommunicator.hh
+++ b/opm/grid/common/p2pcommunicator.hh
@@ -238,6 +238,15 @@ public:
     /** \brief exchange data with peers, handle defines pack and unpack of data */
     virtual void exchange ( DataHandleInterface& ) const;
 
+    /** \brief exchange data with peers, handle defines pack and unpack of
+     * data.  User-defined message tag.
+     *
+     * \param[in,out] handle Message.
+     *
+     * \param[in] tag Message tag.
+     */
+    void exchange( DataHandleInterface& handle, const int tag ) const;
+
     /** \brief exchange data with peers, handle defines pack and unpack of data,
      *  if receive buffers are known from previous run and have not changed
      *  communication could be faster */
--- a/opm/grid/common/p2pcommunicator_impl.hh
+++ b/opm/grid/common/p2pcommunicator_impl.hh
@@ -579,17 +579,26 @@ namespace Dune
   template <class MsgBuffer>
   inline void
   Point2PointCommunicator< MsgBuffer >::
-#if HAVE_MPI
-  exchange( DataHandleInterface& handle) const
-#else
-  exchange( DataHandleInterface&) const
-#endif
+  exchange( [[maybe_unused]] DataHandleInterface& handle,
+            [[maybe_unused]] const int            tag ) const
   {
-    assert( _recvBufferSizes.empty () );
+    assert( this->_recvBufferSizes.empty () );
+
 #if HAVE_MPI
-    NonBlockingExchangeImplementation< ThisType > nonBlockingExchange( *this, getMessageTag() );
+    NonBlockingExchangeImplementation< ThisType > nonBlockingExchange( *this, tag );
     nonBlockingExchange.exchange( handle );
-#endif
+#endif  // HAVE_MPI
+  }
+
+  template <class MsgBuffer>
+  inline void
+  Point2PointCommunicator< MsgBuffer >::
+  exchange( [[maybe_unused]] DataHandleInterface& handle) const
+  {
+    assert( this->_recvBufferSizes.empty () );
+#if HAVE_MPI
+    this->exchange( handle, getMessageTag() );
+#endif  // HAVE_MPI
   }
 
   // --exchange
--- a/tests/p2pcommunicator_test.cc
+++ b/tests/p2pcommunicator_test.cc
@@ -156,6 +156,12 @@ void testCommunicator( const bool output )
     comm.exchange( handle );
   }
 
+  {
+    // use handle with custom message tag to perform the same operations as above
+    DataHandle handle( comm, output );
+    comm.exchange( handle, 1 << 12 );
+  }
+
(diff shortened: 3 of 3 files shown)


bska:
jenkins build this please

blattms:
> That way we're not necessarily restricted to MPI_COMM_WORLD since we don't rely on a [static variable](https://github.com/OPM/opm-grid/blob/b4b7971e4eaaff4830b9ea5254dc39748092f93d/opm/grid/common/p2pcommunicator.hh#L252) to [create message tags](https://github.com/OPM/opm-grid/blob/b4b7971e4eaaff4830b9ea5254dc39748092f93d/opm/grid/common/p2pcommunicator.hh#L255) for each new message.

I don't understand this comment. Where is the relation to MPI_COMM_WORLD?

In my understanding the message tag is used to identify the correct messages on a communicator. Hence we have to make sure that at any given point all messages on a communicator that have the same tag server the same purpose (the same p2p exchange). My conclusion would be that there needs to a different tag for each well and that must be distinct from the one which is used at other places.

Maybe we should allow custom message tags only from a region that is not used by the others?

bska:
> > That way we're not necessarily restricted to MPI_COMM_WORLD since we don't rely on a [static variable](https://github.com/OPM/opm-grid/blob/b4b7971e4eaaff4830b9ea5254dc39748092f93d/opm/grid/common/p2pcommunicator.hh#L252) to [create message tags](https://github.com/OPM/opm-grid/blob/b4b7971e4eaaff4830b9ea5254dc39748092f93d/opm/grid/common/p2pcommunicator.hh#L255) for each new message.
> 
> I don't understand this comment. Where is the relation to MPI_COMM_WORLD?

It's derived/implied from a requirement, possibly erroneously, asserted in the [internal implementation of the non-blocking communication protocol](https://github.com/OPM/opm-grid/blob/b4b7971e4eaaff4830b9ea5254dc39748092f93d/opm/grid/common/p2pcommunicator_impl.hh#L168-L174) that states that all processes engaged in exchanging a message always have the same tag.  That does clearly make sense for an individual message.  On the other hand, if we're using the automatic generation of message tags based on a static member of class `Point2PointCommunication`, then every call to `Point2PointCommuncation::exchange()` must use the same communication object.  In other words, it's not (currently) possible to use a `Point2PointCommuncation` object to communicate amongst a subset of the MPI ranks.  This PR allows the user to define a message tag which bypasses the automatic tag generation and therefore enables communication between that subset of the ranks.

> we have to make sure that at any given point all messages on a communicator that have the same tag serve the same purpose

Yes.  But we do not necessarily have to use distinct tags for separate messages of the same type.  We could use a single tag for well data, another tag for cell data, a third for block data, and a fourth for aquifer data.  The `Point2PointCommuncation` receive object waits for all its senders and does not accept multiple messages from the same sender with the same tag so that's an implied serialisation point.

> My conclusion would be that there needs to a different tag for each well and that must be distinct from the one which is used at other places.

The tag only needs to be distinct within the same communcator object.

blattms:
This gist of my rather confused musings below probably is the following:
Making the tag a normal variable is a good thing, if we at the same time clone the MPI_Communicator in the constructor and use the duplicate inside of Point2PointCommunication. This will safely separate messages of separate instances of the class. For communicating only on a subset of the processes I would still recommend to create a custom MPI_Communicator that only operates on that subset. I don't think a custom message tag is needed with this approach as the communicator is only used for one purpose.

And here come my probably confused thoughts that lead to this insight (I really hope that it is one):

DISLAIMER 1:
I am not against merging this in spirit of "if somebody wants to shoot themselves in the foot, they should be able to do that". If people know what they are doing then they can handle it.

DISCLAIMER 2:
My hunch concerning the tag in Point2PointCommunication is, that it is used to separate subsequent calls to exchange in a safe manner. Meaning produce a deadlock rather then processing wrong data. 
IMHO, the tag should probably not be used for anything else. One could argue that for safety, the MPI_Communicator
should be duplicated in the constructor. When doing that there is no need for the tag to be static.
Clearly, the old design assumes that all ranks create Point2PointCommunication and simultaneously call exchange.

> Yes. But we do not necessarily have to use distinct tags for separate messages of the same type. We could use a single tag for well data, another tag for cell data, a third for block data, and a fourth for aquifer data. The Point2PointCommuncation receive object waits for all its senders and does not accept multiple messages from the same sender with the same tag so that's an implied serialisation point.

I don't see any synchronization points in this code. Unless the calling code happens to have one, bad things could happen, because messages could overtake each other for whatever reason.

blattms:
Just looked at OPM/opm-simulators#4201 as the use case for this. My feeling is that if we clone the communicator and make the message tag non-static then that PR does not need any custom message tags anymore.

bska:
> My feeling is that if we clone the communicator and make the message tag non-static then that PR does not need any custom message tags anymore.

You're right, it does not.  We still can't use `MPI_Gatherv()` though.

blattms:
Would be cool if we would do that then. Would be a huge improvement to the current stage on its own,

blattms:
Cool. :sunglasses:

PR #597, opened 2022-10-28
Allow creation of cartesian grids not starting at origin.

blattms:
Needed for some future tests of the refinement code.

### DIFF
--- a/opm/grid/CpGrid.hpp
+++ b/opm/grid/CpGrid.hpp
@@ -308,9 +308,10 @@ namespace Dune
         /// Create a cartesian grid.
         /// \param dims the number of cells in each cartesian direction.
         /// \param cellsize the size of each cell in each dimension.
-        /// \param shift Number of cell widths the left-most front-most lowest
-        ///             corner of the grid should be shifted to the right, back,
-        ///             and top
+        /// \param shift The origin of the grid, i.e. the corner of the cell with index (0,0,0) where
+        ///              the left, bottom, and top face of that cell intersect, is at the coordinate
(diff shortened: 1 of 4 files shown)


blattms:
I intended to self-merge this, but the method documentation now is so lousy that even I don't understand it. Any suggestions for improvements?

blattms:
jenkins build this please

blattms:
jenkins build this please

opm/grid/CpGrid.hpp
```diff
@@ -308,8 +308,12 @@ namespace Dune
         /// Create a cartesian grid.
         /// \param dims the number of cells in each cartesian direction.
         /// \param cellsize the size of each cell in each dimension.
+        /// \param shift Number of cell widths the left-most front-most lowest
+        ///             corner of the grid should be shifted to the right, back,
+        ///             and top
```
atgeirr, line 313: I think I understood this, but maybe
"moves the origin of the grid in (i, j, k) coordinates relative to the default, so that if `shift == {si, sj, sk}` then the origin of the grid (i.e. the location of the corner with the minimum coordinates) is not (0.0, 0.0, 0.0) but (si * cellsize[0], sj * cellsize[1], sk * cellsize[2])"

Now if that is wrong, I truly did not understand the doc...

blattms: I think we are getting there. Just not sure whether there is an "origin of a grid".

blattms: thanks. I tried to improve the comment based on your input. seems much clearer now.

blattms:
jenkins build this please

PR #599, opened 2022-10-31
Use Separate Tag Series for Each P2P Communicator

bska:
This PR makes the tag generator into a per-object mechanism. This, in turn, means that there can be multiple P2P communicator objects, possibly amongst subsets of MPI processes, that exchange messages without generating conflicting message tags.

Immediate use case is communicating per-connection result data between processes that happen to share a distrbuted well.

This PR supersedes and closes #596.


bska:
jenkins build this please

blattms (approved):

blattms:
Thanks  a lot.

PR #610, opened 2023-01-02
Remove sstream use in exception handling

akva2:
Preparatory work for removing sstream usage in OPM_THROW.


blattms:
why do we want/need to get rid of sstream? IMHO the code here does not really get better by this

akva2:
stream classes are large. the more compile units we can avoid them in, the better. at the very least i want to get the stream class usage under control, as it is now it's everywhere because of being included in all sorts of headers.  at the core of a lot of this is the stream usage in OPM_THROW which is what i'm targeting.

my main motivating is that in many places we would rather use {fmt}, and having to pull in sstream when we have a better alternative is unnecessary. this will be even more desirable once we hit c++20 and can use the standard format lib.

bska (approved):
Thanks a lot. I'll merge into master.

PR #612, opened 2023-01-05
adjust to MinpvMode now being an enum class

akva2:
Downstream of https://github.com/OPM/opm-common/pull/3317


bska (approved):
Thanks a lot.  I'll merge into master.

PR #613, opened 2023-01-05
adjust to PinchMode now being an enum class

akva2:
Downstream of https://github.com/OPM/opm-common/pull/3318


bska (approved):

PR #614, opened 2023-01-12
do not include WellConnections.hpp in header

akva2:
Downstream of https://github.com/OPM/opm-common/pull/3343


bska (approved):

PR #618, opened 2023-01-17
rename OpmParserIncludes.hpp to OpmWellType.hpp 

akva2:
And forward Well

Reduce hotness of Well.hpp by 26% (70 files, 200 instead of 270)


akva2:
jenkins build this opm-simulators=4392 please

bska (approved):

PR #619, opened 2023-01-18
[bugfix] Prevent deadlock for accidentally distributed wells.

blattms:
If distributed wells are not allowed, we try to move all perforated cells those wells to one process. We throw an exception if that was not successful. That exception was only thrown on one process causing OPM flow to hang afterwards. This is fixed by this change that throws on all processes.

Drive-by fix while working on the real stuff.

### DIFF
--- a/opm/grid/common/ZoltanPartition.cpp
+++ b/opm/grid/common/ZoltanPartition.cpp
@@ -116,6 +116,8 @@ makeImportAndExportLists(const Dune::CpGrid& cpgrid,
 
 #ifndef NDEBUG
             int index = 0;
+            std::unordered_set<int> distributed_wells;
+
             for( auto well : gridAndWells->getWellsGraph() )
             {
                 int part=parts[index];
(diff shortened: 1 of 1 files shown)


blattms:
jenkins build this please

opm/grid/common/ZoltanPartition.cpp
```diff
@@ -129,10 +131,18 @@ makeImportAndExportLists(const Dune::CpGrid& cpgrid,
...
-                    OPM_THROW(std::domain_error, "Well is distributed between processes, which should not be the case!");
+                    distributed_wells.insert(index);
                 }
                 ++index;
             }
+            auto num_dist_wells = cc.sum(distributed_wells.size());
+            if (num_dist_wells) {
+                OPM_THROW(std::domain_error,
+                          std::to_string(num_dist_wells)
+                              + "Well is distributed between processes, which should not be the case!");
```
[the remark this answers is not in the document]
blattms: Indeed. Resolved. (One should not push 1 minute before a meeting...)

blattms:
jenkins build this please

opm/grid/common/ZoltanPartition.cpp
```diff
@@ -129,10 +131,19 @@ makeImportAndExportLists(const Dune::CpGrid& cpgrid,
...
+                    distributed_wells.insert(index);
                 }
                 ++index;
             }
+            auto num_dist_wells = cc.sum(distributed_wells.size());
+            if (num_dist_wells) {
+                OPM_THROW(std::domain_error,
+                          std::to_string(num_dist_wells) + " well"
+                              +  ((num_dist_wells >= 2) ? "s are" : " is")
+                              + "distributed between processes, which should not be the case!");
```
bska, line 143: Looks better, but now you get something like
```
1 well isdistributed ...
```
or
```
3 wells aredistributed ...
```
instead...

blattms: thanks.

blattms:
jenkins build this please

bska (approved):

PR #620, opened 2023-01-18
[bugfix] Make sure that few cells are moved for getting rid off distributed wells.

blattms:
Well, max_element expects a less-than functor. Implementing that using `>` is possible but not like this. It resulted in greater-than and moved all cells to a new process (often the last one). Not really what you want for a scalable software.

Fixed by switching to `<`. I guess this is the worst typo since 2011!

Morons producing such typos should be tared and feathered :rofl: Oops, that would me, I guess. :flushed: 

Might be related to #476 (see removed comment)

### DIFF
--- a/opm/grid/common/WellConnections.cpp
+++ b/opm/grid/common/WellConnections.cpp
@@ -201,17 +201,13 @@ postProcessPartitioningForWells(std::vector<int>& parts,
             assert(!no_connections_on_proc.empty());
             int owner = no_connections_on_proc.begin()->first;
 
-            // \todo remove trigger code for #476 that moves all wells to the last rank
             if (no_connections_on_proc.size() > 1) {
                 // partition with the most connections on it becomes new owner
-                int new_owner =
-                    std::max_element(no_connections_on_proc.begin(),
-                                     no_connections_on_proc.end(),
-                                     [](const std::pair<int, std::size_t> &p1,
-                                        const std::pair<int, std::size_t> &p2) {
-                                         return (p1.second > p2.second);
-                                     })
-                    ->first;
+                int new_owner = std::max_element(no_connections_on_proc.begin(),
+                                                 no_connections_on_proc.end(),
+                                                 [](const auto& p1, const auto& p2)
+                                                 { return (p1.second < p2.second); })
+                                    ->first;
                 std::cout << "Manually moving well " << well.name()
                           << " to partition " << new_owner << std::endl;
 


blattms:
jenkins build this please.

blattms:
jenkins build this please.

blattms:
No dot moron! :rofl:

blattms:
jenkins build this please

blattms:
benchmark please

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: drogon - Threads: 1         | 1.016    |
| opm-git | OPM Benchmark: drogon - Threads: 8         | 1.005    |
| opm-git | OPM Benchmark: smeaheia - Threads: 1       | 0.999    |
| opm-git | OPM Benchmark: smeaheia - Threads: 8       | 0.997    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1 | 0.998    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8 | 0.997    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1 | 0.998    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8 | 1.002    |



View result details @ https://www.ytelses.com/opm/?page=result&id=1927

blattms:
that looks fine (like noise)

bska:
> that looks fine (like noise)

Right, but there's one subtle performance issue here that we might consider fixing.  On the one hand it's unlikely to matter a lot because I expect these containers to be fairly small, but on the other we might want to avoid unnecessary costs too.

Since [`no_connections_on_proc`](https://github.com/OPM/opm-grid/blob/780fbd24b5d3b439206dc935c47bd4d103975834/opm/grid/common/WellConnections.cpp#L197) is a `std::map<int,std::size_t>` its `value_type` is
```C++
std::pair<const int, std::size_t>
```
Therefore, since the comparison function in the call to [`std::max_element()`](https://github.com/OPM/opm-grid/blob/780fbd24b5d3b439206dc935c47bd4d103975834/opm/grid/common/WellConnections.cpp#L209-L212) is declared to take two
```C++
const std::pair<int, std::size_t>&
```
we effectively end up copying the entire sequence, twice, as we're traversing the elements.

The simplest way to avoid copying the elements is to use a generic lamda instead,
```C++
[](const auto& p1, const auto& p2)
{
    return p1.second < p2.second;
}
```
but it may just be more readable to switch the parameter types to
```C++
const std::pair<const int, std::size_t>&
```
instead.

blattms:
Indeed (after several minutes I got that it's the word const). Fixed.

blattms:
jenkins build this please

bska (approved):
Thanks a lot for the updates.  I'll merge this into the master branch now.

atgeirr:
The benchmarks seem to indicate no effect on those cases, but they might be cases where few cells are moved anyway. After all, the partitioning weights the well connections strongly.

As for the "worst typo since 2011" I see your bet and raise my wrong 2x2 matrix inversion...

PR #624, opened 2023-02-03
Guard Against Division By Zero in Centroid Calculation

bska:
In degenerate geometries it is possible for a face to reduce to a line in which case the area is zero and dividing by the total area leads to centroids with NaN coordinates.  Return the 'inpoint' in this case.

While here, also apply the same guard to the face normal.


bska:
jenkins build this please

blattms:
what is the `inpoint`?

bska:
> what is the `inpoint`?

Initial guess at face centroid.  Typically the arithmetic average of the face's vertex coordinates.

blattms:
Good. That seems to make sense. I am merging.

PR #625, opened 2023-02-03
Expand Left-Handed vs Right-Handed Inference

bska:
This PR switches the `is_lefthanded()` predicate into a coordinate system classifier.  Rename the function according to its new responsibility.  We determine if the geometry's coordinate system is right-handed (most common) or left-handed, and add a fall-back state of "inconclusive" if the type cannot be fully determined.

In that case, we fall back to a more expensive algorithm that considers every active cell.  The existing classifier tried to form a bounding box for the entire model and then to calculate the triple product of that bounding box.  This approach, while inexpensive when it works, is too fragile.  There are modelling tools which place multiple cell columns in the same physical location if there are no active cells within the column.  This, in turn, leads to zero-sized bounding boxes in the original implementation and an inconclusive result.

The fall-back strategies computes the triple product for each active cell and infers "left-handed" vs. "right-handed" coordinate systems for the complete geometry based on the proportion of left-handed vs. right-handed individual cells.

While here, also make function `process_grdecl()` return a status code of zero if we fail to construct a 'processed_grid' object and update callers to use this new information. Typical causes for failing to construct the grid is

  * Failing to conclusively determine if the grid's coordinate system is right-handed or left-handed.
  * Failing to allocate memory (malloc/realloc)

There are still a couple of malloc failures that end up calling `exit()`, but we will address those in follow-up work.

### DIFF
--- a/opm/grid/cpgpreprocess/preprocess.c
+++ b/opm/grid/cpgpreprocess/preprocess.c
@@ -670,12 +670,24 @@ get_zcorn_sign(int nx, int ny, int nz, const int *actnum,
 }
 
 
+/* ---------------------------------------------------------------------- */
+/* Compute (I,J,K) Cartesian coordinate of a single cell.
+ *
+ * @param[in] nx Number of Cartesian cells in model's X direction
+ * @param[in] ny Number of Cartesian cells in model's Y direction
+ * @param[in] nz Number of Cartesian cells in model's Z direction
+ * @param[in] c Linear Cartesian index, natural ordering, of model cell
+ * @param[out] i Cartesian I coordinate of cell \c c
+ * @param[out] j Cartesian J coordinate of cell \c c
+ * @param[out] k Cartesian K coordinate of cell \c c */
+/* ---------------------------------------------------------------------- */
 static void
 ind2sub(const size_t nx,
         const size_t ny,
         const size_t nz,
         size_t       c ,
         size_t *i, size_t *j, size_t *k)
+/* ---------------------------------------------------------------------- */
 {
     assert (c < (nx * ny * nz));
 
@@ -717,6 +729,14 @@ vert_size(const struct grdecl *in,
 
 
 /* ---------------------------------------------------------------------- */
+/* Compute (X,Y,Z) coordinate of a single vertex on a pillar.
+ *
+ * @param[in] in Corner-point grid structure
+ * @param[in] pillar Pillar ID.  Zero to (nx+1)*(ny+1) - 1, inclusive.
+ * @param[in] z Z coordinate of vertex
+ * @param[out] coord Vertex coordinate.  Expected to point to the start of
+ *   an array of size at least 3. */
+/* ---------------------------------------------------------------------- */
 static void
 vertex_coord(const struct grdecl *in,
              const size_t         pillar,
@@ -727,6 +747,8 @@ vertex_coord(const struct grdecl *in,
     const double *top = &in->coord[6*pillar + 0];
     const double *bot = &in->coord[6*pillar + 3]; /* == top + 3 */
 
+    /* Deem top and bottom pillar points coincident if Z coordinates along
+     * pillar differ by less than 1 micrometre */
     const int coincide = fabs(top[2] - bot[2]) < 1.0e-6;
 
     const double t = coincide
@@ -739,6 +761,23 @@ vertex_coord(const struct grdecl *in,
 }
 
 
+/* ------------------------------------------------------------------------ */
+/* Calculate cross product (I - O) x (J - O) when I and J are points on the
+ * I and J axes respectively, and O is the coordinate system origin.
+ *
+ * @param[in] origin Location of coordinate system's origin.  Expected to
+ *   point to the start of an array of size at least 3.
+ *
+ * @param[in] i_axis Location of point on coordinate system's I axis.
+ *   Expected to point to the start of an array of size at least 3.
+ *
+ * @param[in] j_axis Location of point on coordinate system's J axis.
+ *   Expected to point to the start of an array of size at least 3.
+ *
+ * @param[out] cross Resulting cross product.  Expected to point to the
+ *   start of an array of size at least 3.
+ */
+/* ------------------------------------------------------------------------ */
 static void
 bounding_box_cross_axes(const double *origin,
                         const double *i_axis,
@@ -756,6 +795,25 @@ bounding_box_cross_axes(const double *origin,
 }
 
 
+/* ------------------------------------------------------------------------ */
+/* Calculate triple product ((I - O) x (J - O)) . (K - O) when I,J,K are
+ * points on the I, J, and K axes respectively and O is the coordinate
+ * system origin.
+ *
+ * @param[in] origin Location of coordinate system's origin.  Expected to
+ *   point to the start of an array of size at least 3.
+ *
+ * @param[in] i_axis Location of point on coordinate system's I axis.
+ *   Expected to point to the start of an array of size at least 3.
+ *
+ * @param[in] j_axis Location of point on coordinate system's J axis.
+ *   Expected to point to the start of an array of size at least 3.
+ *
+ * @param[in] k_axis Location of point on coordinate system's K axis.
+ *   Expected to point to the start of an array of size at least 3.
+ *
+ * @return Triple product. */
+/* ------------------------------------------------------------------------ */
 static double
 bounding_box_triple_product(const double *origin,
                             const double *i_axis,
@@ -773,10 +831,17 @@ bounding_box_triple_product(const double *origin,
 
 
 /* ---------------------------------------------------------------------- */
+/* Known coordinate system types.  The 'Inconclusive' type is the
+ * default/failure type when we're not able to infer the actual type. */
+/* ---------------------------------------------------------------------- */
 enum CoordinateSystemType { Inconclusive, RightHanded, LeftHanded };
 /* ---------------------------------------------------------------------- */
 
 
+/* ---------------------------------------------------------------------- */
+/* Classify the coordinate system geometry type based on sign of triple
+ * product.  Positive value is right-handed, negative value is left-handed,
+ * and zero is inconclusive. */
 /* ---------------------------------------------------------------------- */
 static enum CoordinateSystemType
 classify_geometry(const double triple)
@@ -794,6 +859,18 @@ classify_geometry(const double triple)
 }
 
 
+/* ---------------------------------------------------------------------- */
+/* Classify the coordinate system geometry type of a single active cell
+ * based on sign of cell's triple product
+ *
+ * @param[in] in Corner-point geometry.
+ * @param[in] i Cell's Cartesian I index.
+ * @param[in] j Cell's Cartesian J index.
+ * @param[in] k Cell's Cartesian K index.
+ * @param[in] sign ZCORN ordering sign.
+ * @param[in] off ZCORN vertex offsets for each of the cell's vertices.
+ * @return Coordinate system type.
+ */
 /* ---------------------------------------------------------------------- */
 static enum CoordinateSystemType
 get_cell_type(const struct grdecl *in,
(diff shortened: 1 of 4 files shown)


bska:
With this PR I am able to run a field case from a client.  Previously, this field case would fail at grid construction with the very hard to understand diagnostic
```
opm/opm-grid/opm/grid/cpgpreprocess/preprocess.c:774: is_lefthanded: Assertion `fabs(triple) > 0.0' failed.
```
For what it's worth, this PR also enables constructing the grid in the model attached to OPM/opm-simulators#3922, but that model does not run due to a convergence problem (very high `CNV` values, even for very small timestep sizes).  I am investigating the latter problem now.

I am starting this PR in draft mode for two reasons
1. I am not convinced that the new criterion is a lot better than the original one, at least not given the additional incurred computational cost
2. It depends on #624 and must not be merged before that PR is merged.

bska:
jenkins build this please

bska:
benchmark please

atgeirr:
I have only looked very briefly at the code, but I think it makes sense to keep the original logic, and fallback to the new (expensive, by cell) logic only if that is inconclusive. Also, I think we should not do `bool is_lefthanded()` but rather `HandedNess determine_handedness()` where `HandedNess` can be an enum class with values `{ LeftHanded, RightHanded, Inconclusive}` to give more information to the calling code. What do you think?

bska:
> I think we should not do
> ```C
> bool is_lefthanded()
>```
> but rather
> ```C
>HandedNess determine_handedness()
> ```
> where `HandedNess` can be an enum class

I suppose.  We're a little limited by the implementation language, since there is no `enum class` in C.  There is, moreover, just a single caller for the predicate and it doesn't really care about anything other than "is this grid left-handed".  Unless we want to report some diagnostic to the user I'm not really sure what the additional information would be used for.

atgeirr:
> I suppose. We're a little limited by the implementation language, since there is no `enum class` in C. There is, moreover, just a single caller for the predicate and it doesn't really care about anything other than "is this grid left-handed". Unless we want to report some diagnostic to the user I'm not really sure what the additional information would be used for.

I guess an old-school enum will have to do. My wish is not as much motivated by the caller's possible actions, as much as "do not let this function lie". I think the calling site here really wants a right-handed coordinate system, and an inconclusive case should be disallowed? Or is that the situation that we now can and should accept?

bska:
> I think it makes sense to keep the original logic, and fallback to the new (expensive, by cell) logic only if that is inconclusive.

I've pushed an update to this effect, and added logic to propagate a status code (zero is failure, one is success) to the callers of function `process_grdecl()`.  It's not much, but will allow the caller to make some kind of decision if the grid construction fails.  I've also amended the PR description to reflect the new semantics.

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: drogon - Threads: 1         | 0.996    |
| opm-git | OPM Benchmark: drogon - Threads: 8         | 1.066    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1 | 1.002    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8 | 0.976    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1 | 0.996    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8 | 1.001    |



View result details @ https://www.ytelses.com/opm/?page=result&id=1949

bska:
jenkins build this please

atgeirr:
> I've pushed an update to this effect, and added logic to propagate a status code (zero is failure, one is success) to the callers of function `process_grdecl()`. It's not much, but will allow the caller to make some kind of decision if the grid construction fails. I've also amended the PR description to reflect the new semantics.

Thanks, I think the new semantics are welcome, I think this will make some assert() failures due to grid issues into throws?

atgeirr:
The Drogon performance increase of 6% in parallel makes no sense. Any idea? This PR should also not change any results, can you attest to that?

bska:
> I think [the semantic change] will make some `assert()` failures due to grid issues into throws?

Yes, I believe so.

> The Drogon performance increase of 6% in parallel makes no sense. Any idea?

It **might** have been related to the earlier version of the patch that also included #624.  I'll rerun the benchmark to check.  I'll also mark this PR as "ready for review".

bska:
jenkins build this please

bska:
benchmark please

ytelses:
Benchmark result overview: 




| opm-git | OPM Benchmark: drogon - Threads: 1         | 0.983    |
| opm-git | OPM Benchmark: drogon - Threads: 8         | 1.185    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1 | 0.997    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8 | 1.018    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1 | 0.99     |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8 | 0.991    |



View result details @ https://www.ytelses.com/opm/?page=result&id=1953

atgeirr:
> opm-git	OPM Benchmark: drogon - Threads: 8    |   1.185

I want some of the juice Drogon is drinking!

bska:
>> | Test    | Configuration                              | Relative |
>> | ------- | ------------------------------------------ | -------- |
> > opm-git | OPM Benchmark: drogon - Threads: 8    |   1.185
> 
> I want some of the juice Drogon is drinking!

I don't.  It's either PEDs or the test is non-deterministic in some way.

alfbr:
The model has one producing well which is difficult to converge. If for instance one of the simulations closes the well earlier than the other, that easily explains the difference.  I do not think this PR is to blame, so this can be safely merged.

opm/grid/cpgpreprocess/preprocess.c
```diff
@@ -716,29 +717,178 @@ vert_size(const struct grdecl *in,
...
+}
+
+
+static void
+bounding_box_cross_axes(const double *origin,
+                        const double *i_axis,
+                        const double *j_axis,
+                        double       *cross)
+{
+    cross[0] = (i_axis[1] - origin[1]) * (j_axis[2] - origin[2])
```
blattms, line 748: Reuse cross from geometry.c?

bska:
> The [Drogon] model has one producing well which is difficult to converge. If for instance one of the simulations closes the well earlier than the other, that easily explains the difference.

Right, but that makes the test/model close to non-deterministic which is not ideal for a benchmark.

> I do not think this PR is to blame,

Neither do I.

opm/grid/cpgpreprocess/preprocess.c
```diff
@@ -716,29 +717,178 @@ vert_size(const struct grdecl *in,
...
+
+/* ---------------------------------------------------------------------- */
+static enum CoordinateSystemType
+get_cell_type(const struct grdecl *in,
+              const size_t         i,
+              const size_t         j,
+              const size_t         k,
+              const double         sign,
+              const size_t         off[8])
+/* ---------------------------------------------------------------------- */
```
blattms, line 805: Have been biting my tongue, but now I am lost. Some documentation of these function would really help others.

opm/grid/cpgpreprocess/preprocess.c
```diff
@@ -716,29 +717,178 @@ vert_size(const struct grdecl *in,
...
+        (ctype[Inconclusive] > ctype[LeftHanded]))
+    {
+        return Inconclusive;
+    }
+    else if (ctype[RightHanded] > ctype[LeftHanded]) {
+        return RightHanded;
+    }
+    else {
+        return LeftHanded;
+    }
```
blattms, line 872: Totally unclear to me. So it is ok to have different cells as long as there are few inconclusive one. And if only there is one more right-handed cell than left-handed ones everything is assumed to be right-handed?

blattms (commented):
Not a full review as I got lost rather soon. Maybe we should add some documentation?

alfbr:
> > The [Drogon] model has one producing well which is difficult to converge. If for instance one of the simulations closes the well earlier than the other, that easily explains the difference.
> 
> Right, but that makes the test/model close to non-deterministic which is not ideal for a benchmark.

True. However, we have the same problem with the prediction version of Norne (which is broken again?). At the same time, there is a real need for benchmarking of models with tricky prediction controls such as THP. My hope is that we can remedy this by making the well solve more robust in Flow. If there are other benchmark models that can be used instead, then please bring them on :)

atgeirr:
> > > The [Drogon] model has one producing well which is difficult to converge. If for instance one of the simulations closes the well earlier than the other, that easily explains the difference.
> > 
> > 
> > Right, but that makes the test/model close to non-deterministic which is not ideal for a benchmark.

The runs with and without this PR should close/open wells exactly in identical ways, it should not change the results at all. I think this may point to an unintentional change from this PR. Can you run Drogon with this PR to verify that the results are identical?

bska:
> I think [large variations in CPU time] may point to an unintentional change from this PR. Can you run Drogon with this PR to verify that the results are identical?

I didn't look at every vector, but I selected a number of vectors that are moderately hard to match unless the solution is the same.  Note that the drop in TCPU is at the restart step.  We don't generate a continuous TCPU vector across restarts.  Whether or not we should is a discussion for another time.

---

Level | Comparison Plot
------ | -----------
Field |  Plot_2
Well OP5 (hard to converge as [mentioned earlier](https://github.com/OPM/opm-grid/pull/625#issuecomment-1420469412)) | OP5
Simulator |  TCPU

opm/grid/cpgpreprocess/preprocess.c
```diff
@@ -802,16 +980,16 @@ reverse_face_nodes(struct processed_grid *out)
...
+ * Public interface
+ * ---------------------------------------------------------------------- */
+int process_grdecl(const struct grdecl   *in,
+                   double                 tolerance,
+                   const int             *is_aquifer_cell,
+                   struct processed_grid *out,
+                   int                    pinchActive)
 {
-    struct grdecl g;
+    struct grdecl g = {0};
```
blattms, line 992: Out of curiosity: Is this valid C?

opm/grid/cpgpreprocess/preprocess.c
```diff
@@ -716,29 +717,178 @@ vert_size(const struct grdecl *in,
...
+vertex_coord(const struct grdecl *in,
+             const size_t         pillar,
+             const double         z,
+             double              *coord)
+/* ---------------------------------------------------------------------- */
+{
+    const double *top = &in->coord[6*pillar + 0];
+    const double *bot = &in->coord[6*pillar + 3]; /* == top + 3 */
+
+    const int coincide = fabs(top[2] - bot[2]) < 1.0e-6;
```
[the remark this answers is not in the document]
bska: > is 1e-6 really the correct choice? Seems large/coarse. How about 100*DBL_EPSILON?

Sure, but these are SI values.  I'm prepared to declare two points as coincident if they differ by less than $1\mu\mathrm{m}$.

opm/grid/cpgpreprocess/preprocess.c
bska: > Reuse cross from geometry.c?

Possibly.  The one downside to doing that is it'd make the function callable in the general library since we don't control symbol visibility.  I suppose we could extract the function as `cpg_cross_product()` or something along those lines.

opm/grid/cpgpreprocess/preprocess.c
bska: > So it is ok to have different cells as long as there are few inconclusive one. And if only there is one more right-handed cell than left-handed ones everything is assumed to be right-handed?

All very good points, and I hinted at this in an [earlier comment](https://github.com/OPM/opm-grid/pull/625#issuecomment-1415686665).  It's mostly based on a "likelihood"/empirical argument, but there's no theory behind it.  It might be better, but probably more expensive, to try to form a model bounding box for the columns with active cells only.  For now I think this is "good enough", but I am quite prepared to be shown otherwise.

opm/grid/cpgpreprocess/preprocess.c
bska: > Is this valid C?

Yes.  It's the portable way of zeroing out an object of `struct` type.  Integers will be `0`, floating-point values will be `0.0` and pointers will be `NULL`.  That may not what's needed in every case, but it is what I want here.

opm/grid/cpgpreprocess/preprocess.c
bska: > Some documentation of these function would really help others.

I've pushed an update which adds some Doxygen-like comments to most of the internal functions.

PR #626, opened 2023-02-03
Create grid with LGR (Cartesian type)

aritorto:
Given a Cartesian grid, we build an LGR from a selected patch of cells and the corresponding leaf view. Parent-children information is stored but not incorporated in the CpGridData object. 

Next step: make the parent-children information usable/available in Entity.hpp.

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -89,6 +89,7 @@ list (APPEND TEST_SOURCE_FILES
   tests/cpgrid/entity_test.cpp
   tests/cpgrid/facetag_test.cpp
   tests/cpgrid/geometry_test.cpp
+  tests/cpgrid/grid_lgr_test.cpp
   tests/cpgrid/orientedentitytable_test.cpp
   tests/cpgrid/partition_iterator_test.cpp
   tests/cpgrid/shifted_cart_test.cpp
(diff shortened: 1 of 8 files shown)


bska:
jenkins build this please

blattms:
Converted to draft until the last commit is moved to a new branch.

blattms:
Thanks a lot for cleaning up.

blattms:
jenkins build this please

aritorto:
jenkins build this please

aritorto:
To describe the patch of cells that gets refined, we use `startIJK` and `endIJK `Cartesian triple indices. 

`startIJK` refers also to the first cell with zero based indexing.  However, `endIJK` is "one after the last". To avoid confusion with similar parameter in the eclipse file which refers to the last cell to include in the patch, clarifying comments have been added to `endIJK` description.

blattms:
Thanks a lot for the great work. I am merging this now. The functionality is not used currently and should not interfere with the simulator.
There will be coming some cleanup PRs (e.g. moving implementations to the *cpp files).
At the moment we are fighting with circuclar includes when adding the parent/children solutions. Seems to be rather complicated to get rid of those as all information is stored in CpGridData and Entities have basically no state. We will probably need to do some benchmarking and having this in master helps a lot.

bska:
Sorry, @aritorto and @blattms.  I had to revert this PR in #629 because the sequential build failed.  Please fix the underlying issue and resubmit the PR.

blattms:
Weird, I think I used to get notifications about that. Maybe it is considered spam now.
Anyway, that is due to an inconsistency in the serial communication object in DUNE 2.7 missing a conversion to the communicator. Fixed in later versions of DUNE it seems.

PR #628, opened 2023-02-17
[cleanup] Entity::numFaces() removed 

aritorto:
Unnecessary member numFaces() removed, as well as the namespace Detail containing only it.


aritorto:
jenkins build this please

bska:
I don't immediately understand this.  Are you saying that no code calls
```C++
Entity<0>::subEntities(1)
```
in a CpGrid?

aritorto:
Code might call Entity<0>::subEntities(1) which will return 0 since CpGrid does not support codim-1 entities, see, e.g., that there is no typedef in lines 1638-1648:

https://github.com/OPM/opm-grid/blob//master/opm/grid/CpGrid.hpp#L1638-L1648

It supports only Entity<0> ("cells/elemets") and Entity<3> ("points/vertices/corners").  Additionally, Dune::Capabilities::hasEntity<CpGrid,1>::value is false in general.  In conclusion, the method  Entity<0>::subEntity(1) throws an exception, while Entity<0>::subEntitites(1) returns 0. 

I hope this helps. Alternatively, we could ignore this PR and work on adding  codim-1 entities, if it's needed/convenient.

blattms:
thanks a lot for rebasing. Will rerun the checks.

blattms:
jenkins build this please

blattms:

atgeirr:
> Alternatively, we could ignore this PR and work on adding codim-1 entities, if it's needed/convenient.

I think that our grids' faces are not a good fit for the codim 1 entity concept of Dune, but they do fit the intersection concept. So we should not add 1-enitites. At some point early on when we created the CpGrid class this was a little muddled, and we treated the grid for a few purposes as having codim 1 entities (cheating...)

PR #632, opened 2023-02-27
Lgr names added

aritorto:
New member data in CpGrid class, to map the lgr names and their levels (lgr_names_). When creating a grid with refinement of a patch of cells from the global grid ("level 0"), via createGridWithLgr() method, a string-parameter is passed to call the resulting LGR. Tested for level 0 ("GLOBAL") and level 1 (given/parameter).  A method getLgrNameToLevel() has been added as well.

### DIFF
--- a/opm/grid/CpGrid.hpp
+++ b/opm/grid/CpGrid.hpp
@@ -90,7 +90,8 @@ void refine_and_check(const Dune::cpgrid::Geometry<3, 3>&,
 void refinePatch_and_check(Dune::CpGrid&,
                            const std::vector<std::array<int,3>>&,
                            const std::vector<std::array<int,3>>&,
-                           const std::vector<std::array<int,3>>&);
+                           const std::vector<std::array<int,3>>&,
+                           const std::vector<std::string>&);
 
 void refinePatch_and_check(const std::array<int,3>&,
                            const std::array<int,3>&,
@@ -227,7 +228,8 @@ namespace Dune
         void ::refinePatch_and_check(Dune::CpGrid&,
                                      const std::vector<std::array<int,3>>&,
                                      const std::vector<std::array<int,3>>&,
-                                     const std::vector<std::array<int,3>>&);
+                                     const std::vector<std::array<int,3>>&,
+                                     const std::vector<std::string>&);
         friend
         void ::refinePatch_and_check(const std::array<int,3>&,
                                      const std::array<int,3>&,
@@ -460,8 +462,9 @@ namespace Dune
         /// @param [in] startIJK                 Cartesian triplet index where the patch starts.
         /// @param [in] endIJK                   Cartesian triplet index where the patch ends.
         ///                                      Last cell part of the lgr will be {endijk[0]-1, ... endIJK[2]-1}.
+        /// @param [in] lgr_name                 Name (std::string) for the lgr/level1
         void addLgrUpdateLeafView(const std::array<int,3>& cells_per_dim, const std::array<int,3>& startIJK,
-                                  const std::array<int,3>& endIJK);
+                                  const std::array<int,3>& endIJK,  const std::string& lgr_name);
 
         /// @brief Create a grid out of a coarse one and (at most) 2 refinements(LGRs) of selected block-shaped disjoint patches
         ///        of cells from that coarse grid.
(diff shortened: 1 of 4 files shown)


aritorto:
jenkins build this please

opm/grid/CpGrid.hpp
```diff
@@ -826,8 +828,12 @@ namespace Dune
             //  Add Leaf View to data_.
             (this-> data_).push_back(leaf_view_ptr);
             current_view_data_ = data_[2].get();
+            lgr_names_[lgr_name] = 1;
```
goncalvesmachadoc, line 831: Does this means that all LGR's will be at the same leaf view level?

goncalvesmachadoc: Do you need the grid index information in the opm-grid?

aritorto: So far, only one refinement (lgr) is supported via refinePatch() (selection a patch of cells, blocked-shaped, from the global grid, and refine them). We assume the coarse grid to be level 0 (meaning data_[0] of the CpGrid object). The LGR generated via refinement is stored in data_[1] "level 1". The leafview is available in data_[2] but not considered "level 2", since it's the leafview (the cells coming from the coarse grid and that haven't been refined, plus the cells coming from the LGR).

Hope this helps.

aritorto: I'm not sure what grid index means, but I can say that the level index of an LGR (level_ member in a CpGridData object; currently, only 0 [for "GLOBAL"] or 1 [for "LGR", the refinement of a patch of cells from global grid] ) will be needed later to get the correct grid level for an LGR. 

For example, to access the LGR with name "LGR1":

// Find level for LGR
auto level = grid.getLgrNameToLevel().find("LGR1");

// Get corresponding grid view
const auto& lgr = grid.getLevelGridView(level);

Hope this helps.

goncalvesmachadoc: In the commercial simulator each grid has an index. Global is 0, LGR1 would be 1, LGR2 would be 2, and so on...I am thinking if this mapping is done in opm-common, maybe you wouldn't need to know the names here. And for outputting, I believe we may only need grid indexes also.

aritorto: I see, then they seems to represent the same feature (the grid index and what it's call here -- opm-grid -- level index). I can imagine that the map connecting lgr names and integers representing grid/level indices could be useful when the names help the user to distinguish between different lgrs. Example, instead of having "LGR1" and "LGR2", one might prefer to have "LGR_bottom_right_cell", "LGR_left_boundary_corner_cell". 

So far that only one lgr is supported, I agree that it's not very useful. We can also remove it, in case this type of information is not so relevant.

goncalvesmachadoc: Yes, the names itself are not relevant for the code, the grids are numbered in the order they are entered in the input file I think. Outputting the names may be more user-friendly, but I believe we can only visualize the grid indexes in ResInsight (will double check it now to be sure). @bska Do you know what information goes in the binaries? The names may be useful tough for error/warning messages.

blattms: Well, the information needs to be somewhere. Where is probably not that important as long the simulator is able to
retrieve the correct level for an LGR. That can be done perfectly well even without this PR as long as the LGRs and hence the levels are constructed in the right (or a deterministic) order by the simulator. I guess this would be natural anyway. The parser is probably making that information available in the correct order anyway, right?

aritorto:
jenkins build this please

blattms:
I did put this into draft mode as it partly based on #643 and should be merged after it.

aritorto:

aritorto:

aritorto:

aritorto:

aritorto:

blattms:
The [failure](https://ci.opm-project.org/job/opm-grid-PR-builder/397/console)
```
Building downstream opm-upscaling=master configuration=serial
~/workspace/opm-grid-PR-builder@2 ~/workspace/opm-grid-PR-builder@2
hint: Using 'master' as the name for the initial branch. This default branch name
hint: is subject to change. To configure the initial branch name to use in all
hint: of your new repositories, which will suppress this warning, call:
hint: 
hint: 	git config --global init.defaultBranch <name>
hint: 
hint: Names commonly chosen instead of 'master' are 'main', 'trunk' and
hint: 'development'. The just-created branch can be renamed via this command:
hint: 
hint: 	git branch -m <name>
Initialized empty Git repository in /var/lib/jenkins/workspace/opm-grid-PR-builder@2/deps/opm-upscaling/.git/
error: RPC failed; HTTP 500 curl 22 The requested URL returned error: 500
fatal: expected flush after ref listing
error: pathspec 'branch_to_build' did not match any file(s) known to git
fatal: ambiguous argument 'HEAD': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
~/workspace/opm-grid-PR-builder@2
~/workspace/opm-grid-PR-builder@2 ~/workspace/opm-grid-PR-builder@2
CMake Warning:
  Ignoring extra path from command line:

   "/var/lib/jenkins/workspace/opm-grid-PR-builder@2/deps/opm-upscaling"


CMake Error: The source directory "/var/lib/jenkins/workspace/opm-grid-PR-builder@2/deps/opm-upscaling" does not appear to contain CMakeLists.txt.
Specify --help for usage, or press the help button on the CMake GUI.
Build step 'Execute shell' marked build as failure
Archiving artifacts
Recording test results
ERROR: Step ‘Publish JUnit test result report’ failed: No test report files were found. Configuration error?
[GNU C Compiler (gcc)] Skipping execution of recorder since overall result is 'FAILURE'
```
is at least not due to us. I'll retry jenkins. If that fails, @akva2 needs to take a look

blattms:

PR #634, opened 2023-03-08
changed: disable shifted_cart_test, grid_lgr_test and geometry_test with boost <= 1.53

akva2:
They fail to build due to using symbols only in newer boosts. Any modern system should not be affected, it is only relevant for the rhel packaging, so simply disable the tests. Not worth the hassle to rewrite them.

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -88,11 +88,8 @@ list (APPEND TEST_SOURCE_FILES
   tests/cpgrid/entityrep_test.cpp
   tests/cpgrid/entity_test.cpp
   tests/cpgrid/facetag_test.cpp
-  tests/cpgrid/geometry_test.cpp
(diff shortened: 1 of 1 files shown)


akva2:
jenkins build this please

blattms:
Seems like this does not work as expected. On my system (boost 1.74) these tests are not run anymore.

bska:
> `BOOST_VERSION` seems not to be set, but `Boost_VERSION`

True.  At least in more recent versions of [`FindBoost.cmake`](https://cmake.org/cmake/help/latest/module/FindBoost.html), the convention is that `BOOST_X` is an input/hint variable set by the user&mdash;the entity configuring the build&mdash;while `Boost_X` is a result variable that gets defined/set by `FindBoost`.  The two exceptions to this rule are `Boost_NO_SYSTEM_PATHS` and `Boost_ADDITIONAL_VERSIONS`.

PR #635, opened 2023-03-13
Mark Independent Partitioning Routine 'const'

bska:
The `zoltanPartitionWithoutScatter()` member function does not alter the `CpGrid` instance so we can mark it as `const` to make it usable in more contexts.


bska:
jenkins build this please

bska:
Marking this as ready for review since the build check succeeds.

akva2 (approved):

PR #637, opened 2023-03-21
[cmake] Fix version test for Boost that skip compiling some tests.

blattms:
On some systems it is Boost_VERSION and not BOOST_VERSION it seems.
Fixes fallout from #634

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -102,7 +102,7 @@ list (APPEND TEST_SOURCE_FILES
   tests/test_compressed_cartesian_mapping.cpp
 	)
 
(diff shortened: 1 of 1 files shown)


blattms:
jenkins build this please

bska:
> merging as this fixes the problem on my system and prevents full tests of other PRs.

I don't understand what you mean by "prevents" here.  Do you mean "the problem prevents" or "the PR prevents"?  In the former case, then by all means go ahead and merge.  In the latter case, merging is less clear cut.

blattms:
Indeed, I should have written" the problem prevents tests from being compiled and executed", which was a bit unfortunate.

PR #638, opened 2023-03-23
MinpvProcessor: process cells in a vertical column subsequently.

blattms:
The current implementation might be wrong for cases with multiple cells in the prospective pinch-out column (e.g. all cells below the thickness theshold, but the sum is not).
Hence we need to be able to process those cells once only and move kk index to the next cell not being piched out directly.

It seems like for multiple neighbouring cells begin pinched out the current algorithm will also add the same NNC multiple times.

For PINCH option 4 will need to be able to compute the harmonic average of the transmissibilities of the cells beween the active ones. As those cells are not represented in he grid we will need to compute the cell centers, face centers, and face areas of the pinched out cells here and make them available. This might also be easier with the new loop.

Playing it careful and creating it in draft mode for now.

### DIFF
--- a/opm/grid/MinpvProcessor.hpp
+++ b/opm/grid/MinpvProcessor.hpp
@@ -134,9 +134,9 @@ namespace Opm
         }
 
         // Main loop.
-        for (int kk = 0; kk < dims_[2]; ++kk) {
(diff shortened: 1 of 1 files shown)


blattms:
jenkins build this please

blattms:
jenkins build this please

PR #639, opened 2023-03-24
MinpvProcessor: process cells in a vertical column subsequently

blattms:
Supersedes #638 which was not pushed to my fork but the original repoitory.

This refactoring is needed as preparation to support more of the various PINCH cases. 

Last but not least honoring parameter 3 of PINCH which prescribes a maximum threshold of the vertical gap for creating NNCs. Calculating the gap thickness seems rather hard with the current approach.

For PINCH option 4 we will need to be able to compute the harmonic average of the transmissibilities of the cells beween the active ones. As those cells are not represented in the grid we will need to compute the cell centers, face centers, and face areas of the pinched out cells here and make them available. For that we need the uncollapsed geometry of these cells. Getting that seems rather hard with the current approach.

~In case of NOGAP we will now only create an NNC if the cells deactivated due to small porevolume have thickness less than the tolerance. This thickness check was missing before.~ I correct myself: The check was there.

It seems like this refactoring also makes the code a bit easier to understand.

Please note that we changed the logic how to detect inactive cells. It seemed like it would not work correctly if actnum is empty (which it apparently never is at least for CpGrid).

Please note further that we keep a peculiarities of the current code. In  particular the way of collapsing cells. This differs depending on the value of mergeMinPVCells.
    
If that is true the whole column will always be collapsed such that all corners coincide with upper corners of the first cell of the layer (and hence the lower corners of the cell above it). If it is false and there are two (or more) active cells that get     deactivated because of PINCH, then the part of the column starting with the  second cell that gets deactivated is collapsed to the upper corners  of that cell (and not the ones of the first deactivated cell).

An illustration what happens with mergeMinPVCells being false and two active cells that get deactivated because of PINCH:
```
_______                            _______
|     | high pv                    |     | high pv
|_____|                            |_____|
|     |  active low pv                     <---- Collapsed next 2 cells to coincide with lower 
|_____|                ------>                  vertixes of high pv cell
|_____| thin inactive cell          ______ 
|     |  active low pv                     <---- Collapsed last 2 cells
|_____|                                         to upper coordinates of low pv cell
|_____| thin inactive cell
```

Personally It seems more logical (and for the new algorithm that would be easier) to collapse all cells such that they coincide with lower corners of the upper-most cell. This is also what would happen if mergeMinPVCells would be true.

Here is an illustration with another cells at the bottom, that is and stays active:
```
|     | high pv                    |     | high pv
|_____|                            |_____|
|     |  low pv                    |     |  <---- Collapsed next 4 cells to coincide with lower 
|_____|                ------>     |     |        vertixes of high pv cell and extended the next
|_____| thin cell                  |     |        cell to add the volum of the pinched layers.
|     |  low pv                    |     |  
|_____|                            |     |  
|_____| thin cell                  |     |  
|     |                            |     |
|_____| high pv                    |_____|
```

Please note that in the simulator mergeMinPVCells is hardcoded to always be false!

If we ever change that then we also need to think about where to put the deactivated value. In the original algorithm it might be added to an inactive cell that is thicker than the threshold. Or to now cell if there is no cell below.

### DIFF
--- a/opm/grid/MinpvProcessor.hpp
+++ b/opm/grid/MinpvProcessor.hpp
@@ -134,9 +134,9 @@ namespace Opm
         }
 
         // Main loop.
-        for (int kk = 0; kk < dims_[2]; ++kk) {
-            for (int jj = 0; jj < dims_[1]; ++jj) {
-                for (int ii = 0; ii < dims_[0]; ++ii) {
+        for (int jj = 0; jj < dims_[1]; ++jj) {
+            for (int ii = 0; ii < dims_[0]; ++ii) {
+                for (int kk = 0; kk < dims_[2]; ++kk) {
                     const int c = ii + dims_[0] * (jj + dims_[1] * kk);
                     if (pv[c] < minpvv[c] && (actnum.empty() || actnum[c])) {
                         // Move deeper (higher k) coordinates to lower k coordinates.
--- a/opm/grid/MinpvProcessor.hpp
+++ b/opm/grid/MinpvProcessor.hpp
@@ -156,8 +156,8 @@ namespace Opm
 
                         int c_below = ii + dims_[0] * (jj + dims_[1] * (kk_iter));
                         // bypass inactive cells with thickness less then the tolerance
-                        while ( ((actnum.empty() || !actnum[c_below]) && (thickness[c_below] <= z_tolerance))  ){
-                            // move these cell to the posistion of the first cell to make the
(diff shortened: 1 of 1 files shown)


blattms:
jenkins build this please

blattms:
jenkins build this please

blattms:
jenkins build this please

blattms:
Shameless self-merge as tests are green. :grimacing:

PR #640, opened 2023-03-28
Support option 3 (MAX_EMPTY_GAP) of PINCH keyword. 

blattms:
In that case an NNC will only be generated if the thickness of the gap due to the pinchout is less than the threshold giiven.

Whether using the added thickness as the gap is up for discussion. Might not be what the commercial guys are using (maximum distance along the vertical columns where the corners live), but better than no support.

Needs OPM/opm-common#3456

Based on and includes #639.

PR #641, opened 2023-03-28
Support skipping NNCs when PINCH option 4 is ALL.

blattms:
Note that this only supports a special case for option 4 ALL:

If one of the pinched out cells between the two active cells has zero transmissibility in Z-direction then the transimissibility between these cells is assumed to be zero and no NNC will be created.

This is the only case really supported with this. General support will need to calculate the harmonic average of all the transmissibilities of the cells in-between. As those cells are not represented by CpGrid, we will need to export some geometry information from the MinpvProcessor, such as face volume, normal, distance between cell and face center. Then hopefully this can be used by the simulator to calculate these transmissibilities.

This based on and includes #640.
Needs upstream OPM/opm-common#3456

Edit (2023-03-29): In addition we now throw an exception for the yet unsupported cases (option 4 is ALL and an NNC would be created). In that case the transmissibilities would be wrong as we are calculating them from the top and bottom cell and not the harmonic average over all pinched out cells.

### DIFF
--- a/opm/grid/MinpvProcessor.hpp
+++ b/opm/grid/MinpvProcessor.hpp
@@ -156,6 +156,7 @@ namespace Opm
         for (int jj = 0; jj < dims_[1]; ++jj) {
             for (int ii = 0; ii < dims_[0]; ++ii) {
                 for (int kk = 0; kk < dims_[2]; ++kk) {
+                    bool option4ALLSupported = false;
                     const int c = ii + dims_[0] * (jj + dims_[1] * kk);
                     if (pv[c] < minpvv[c] && (actnum.empty() || actnum[c])) {
                         // Cell is made inactive due to MINPV
@@ -182,9 +183,15 @@ namespace Opm
                         bool thin_inactive = !active && thin;
                         bool low_pv_active = pv[c_below] < minpvv[c_below] && active;
                         // In the case of PichNOGAP this cell must be thin to allow NNCs, too.
+                        // In addition skip NNC if PINCH option 4 is ALL and we know that Z transmissibilty will
+                        // be zero because of multz or permz
                         bool nnc_allowed = (!pinchNOGAP || thickness[c] <= z_tolerance)
                             && (!pinchOption4ALL || (permz[c] != 0.0 && multz(c) != 0.0) );
 
+                        if (pinchOption4ALL)
+                        {
+                            option4ALLSupported = option4ALLSupported || permz[c] == 0 || multz(c) == 0;
+                        }
 
                         while ( (thin_inactive || low_pv_active) && kk_iter < dims_[2] )
                         {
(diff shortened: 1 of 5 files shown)


blattms:
jenkins build this opm-common=3456 please

bska:
> jenkins build this opm-common=3456 please

That request will probably fail since OPM/opm-common#3456 was already merged.

blattms:
jenkins build this please

bska:
> Let's see. Jenkins might be rather smart or git when it comes to double merging twice.

Git is able to identify commits/branches that have already been merged, but Jenkins is not.  If you specify `opm-common=3456` then Jenkins will expect to find that branch/PR in the master GitHub repository and give up when it no longer exists.

blattms:
jenkins build this please

blattms:
jenkins build this please

opm/grid/MinpvProcessor.hpp
```diff
@@ -132,11 +147,16 @@ namespace Opm
...
+            OPM_THROW(std::runtime_error,
+                      "If option 4 of PINCH keyword is ALL, then the deck needs to specify PERMZ or PERMX");
+            }
 
         // Main loop.
         for (int jj = 0; jj < dims_[1]; ++jj) {
             for (int ii = 0; ii < dims_[0]; ++ii) {
                 for (int kk = 0; kk < dims_[2]; ++kk) {
+                    bool option4ALLSupported = false;
                     const int c = ii + dims_[0] * (jj + dims_[1] * kk);
```
totto82, line 160: Could you add a commenting line here about the purpose of this boolean. It is not clear from the name. I had to go down to line 315 to understand it. That would also make it easier to understand the code.

tests/test_minpvprocessor.cpp
```diff
@@ -59,21 +59,21 @@ BOOST_AUTO_TEST_CASE(Pinch)
          3.5, 3.5, 3.5, 3.5,
          6, 6, 6, 6
         };
-    auto minpv_result = mp1.process(thickness, z_threshold, pv, minpvv, actnum, fill_removed_cells, z1.data(), pinch_no_gap);
+    auto minpv_result = mp1.process(thickness, z_threshold, 1e20, pv, minpvv, actnum, fill_removed_cells, z1.data(), pinch_no_gap);
```
totto82, line 62: I suggest you add a test where maxgap actually make a difference.

totto82 (commented):
Thanks for the contribution. After testing this code could be merged.

totto82:
Did you test this for the relevant test cases in  https://github.com/OPM/opm-tests/tree/master/pinch

tests/test_minpvprocessor.cpp
blattms: Good point. I added one

blattms (commented):
Thanks for the review, I have addressed everything.
Will undraft this.

blattms:
jenkins build this please

blattms:
I have a relevant test case (T1B2_GAP) provided by @tskille that this PR fixes, but that is not yet part of opm-tests.

PR #642, opened 2023-03-30
fixed: remove const qualifiers on value return types

akva2:
these are bad for RVO

### DIFF
--- a/opm/grid/cpgrid/CpGridData.hpp
+++ b/opm/grid/cpgrid/CpGridData.hpp
@@ -296,7 +296,7 @@ private:
 
     ///
     /// @return patch_dim Patch dimension {#cells in x-direction, #cells in y-direction, #cells in z-direction}.
-    const std::array<int,3> getPatchDim(const std::array<int,3>& startIJK, const std::array<int,3>& endIJK) const
+    std::array<int,3> getPatchDim(const std::array<int,3>& startIJK, const std::array<int,3>& endIJK) const
     {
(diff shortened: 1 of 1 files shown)


akva2:
jenkins build this please

akva2:
RVO is maybe a bit imprecise. The standard rvo  is okay but const disables move semantics so further ellisions is not possible.

bska:
>> May I ask why this influences RVO?
>
> The standard rvo is okay but const disables move semantics so further ellisions is not possible.

Correct, it's not so much RVO directly as it is move semantics and pr-values (or x-values).  If a function returns a `const T` with `T` being an object type, e.g. a built-in type like `double` or a user-defined class type like `std::array<>`, then the return value is not an x-value and the r-value category-based operators/constructors don't activate.

bska:
> Are you guys sure we are moving here? Seems like for all these functions the return objects are created (without name) in the return statement. Isn't this exactly where unnamed RVO should kick in?

It's not RVO.  Code like
```C++
class X { /* ... */ };
const X f();
X g();
// ...
const auto x = f();
const auto y = g();
```
works perfectly and will construct both `x` and `y` in place.  However
```C++
std::vector<X> x;
// ...
x.push_back(f()); // (1)
x.push_back(g()); // (2)
```
will call `push_back(const X&)` for (1) and `push_back(X&&)` for (2).  That might matter.

akva2:
closing as it will be handled elsewhere.

PR #643, opened 2023-04-06
Support multiple LGRs and update LeafGridView 

aritorto:
Cartesian grid required. Based on PR #636.

### DIFF
--- a/opm/grid/cpgrid/CpGrid.cpp
+++ b/opm/grid/cpgrid/CpGrid.cpp
@@ -1362,13 +1362,10 @@ template cpgrid::Entity<3> Dune::createEntity(const CpGrid&, int, bool);
 template cpgrid::Entity<1> Dune::createEntity(const CpGrid&, int, bool); // needed in distribution_test.cpp 
 
 
-void CpGrid::addLgrUpdateLeafView(std::array<int,3> cells_per_dim, std::array<int,3> startIJK, std::array<int,3> endIJK)
+void CpGrid::addLgrUpdateLeafView(const std::array<int,3>& cells_per_dim, const std::array<int,3>& startIJK,
+                                  const std::array<int,3>& endIJK)
 {
-    const std::vector<std::array<int,3>> cells_per_dim_vec{{cells_per_dim}};
-    const std::vector<std::array<int,3>> startIJK_vec{{startIJK}};
-    const std::vector<std::array<int,3>> endIJK_vec{{endIJK}};
-    
-    this -> addLgrsUpdateLeafView(cells_per_dim_vec, startIJK_vec, endIJK_vec);
+    this -> addLgrsUpdateLeafView({cells_per_dim}, {startIJK}, {endIJK});
 }
 
 
@@ -1377,48 +1374,56 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
                                    const std::vector<std::array<int,3>>& endIJK_vec)
 {
     if (!distributed_data_.empty()){
-        OPM_THROW(std::logic_error, "Grid has been distributed. Cannot create LGRs.");
+        if (comm().rank()==0){
+            OPM_THROW(std::logic_error, "Adding LGRs to a distributed grid is not supported, yet.");
+        }
+        else{
+            OPM_THROW_NOLOG(std::logic_error, "Adding LGRs to a distributed grid is not supported, yet.");
+        }
     }
-    if (startIJK_vec.size() > 0){
-        if (!(*data_[0]).disjointPatches(startIJK_vec, endIJK_vec)){
-        OPM_THROW(std::logic_error, "Patches are not disjoint.");
+    if (startIJK_vec.size() > 0 && !(*data_[0]).disjointPatches(startIJK_vec, endIJK_vec)){
+        if (comm().rank()==0){
+            OPM_THROW(std::logic_error, "LGRs are not disjoint.");
+        }
+        else{
+            OPM_THROW_NOLOG(std::logic_error, "LGRs are not disjoint.");
         }
     }
     // Total amount of patches:
     const int& num_patches = startIJK_vec.size();
     assert(cells_per_dim_vec.size() == startIJK_vec.size());
     assert(cells_per_dim_vec.size() == endIJK_vec.size());
-    std::vector<int> all_patch_corners; 
-    std::vector<int> all_patch_faces; 
+    std::vector<int> all_patch_corners;
+    std::vector<int> all_patch_faces;
     std::vector<int> all_patch_cells;
-    std::vector<std::vector<std::array<int,2>>> all_patch_boundary_old_to_new_corners;
-    std::vector<std::vector<std::tuple<int,std::vector<int>>>> all_patch_boundary_old_to_new_faces;
+    
+    // Map to relate boundary patches corners with their equivalent refined/new-born ones. {0,oldCornerIdx} -> {level,newCornerIdx}
+    std::map<std::array<int,2>, std::array<int,2>> old_to_new_boundaryPatchCorners;
+    // Map to relate boundary patch faces with their children refined/new-born ones. {0,oldFaceIdx} -> {level,{newFaceIdx0, ...}}
+    std::map<std::array<int,2>, std::tuple<int, std::vector<int>>> old_to_new_boundaryPatchFaces;
+    
     std::vector<std::vector<std::tuple<int,std::vector<int>>>> all_patch_parent_to_children_cells;
     std::vector<std::vector<std::array<int,2>>> all_patch_child_to_parent_cells;
-    all_patch_boundary_old_to_new_corners.reserve(num_patches);
-    all_patch_boundary_old_to_new_faces.reserve(num_patches);
     all_patch_parent_to_children_cells.resize(num_patches);
     all_patch_child_to_parent_cells.resize(num_patches);
-    // For level0, attach children to each parent cell. For no parents, entry {-1, {-1}} representing {no level, {no children}}
+    
+    // For level0, attach children to each parent cell. For no parents, entry {-1, {}} representing {no level, {no children}}
     auto& l0_parent_to_children_cells = (*data_[0]).parent_to_children_cells_;
-    l0_parent_to_children_cells.resize(data_[0]-> size(0));
-    const std::vector<int> no_children = {-1};    
-    const std::tuple<int,std::vector<int>> has_no_children = std::make_tuple(-1, no_children);
-    for (int cell = 0; cell < data_[0]-> size(0); ++cell){
-        l0_parent_to_children_cells[cell] = has_no_children;
-    }
+    l0_parent_to_children_cells.resize(data_[0]-> size(0), std::make_tuple(-1, std::vector<int>()));
     // Get patches corner, face, and cell indices.
     for (int patch = 0; patch < num_patches; ++patch){
-        const auto& [corners, faces, cells] = (*(this->data_[0])).getPatchGeomIndices(startIJK_vec[patch], endIJK_vec[patch]);
-        all_patch_corners.reserve(corners.size());
+        const auto& corners  = (*(this->data_[0])).getPatchCorners(startIJK_vec[patch], endIJK_vec[patch]);
+        const auto& faces = (*(this->data_[0])).getPatchFaces(startIJK_vec[patch], endIJK_vec[patch]);
+        const auto& cells = (*(this->data_[0])).getPatchCells(startIJK_vec[patch], endIJK_vec[patch]);
+        all_patch_corners.reserve(all_patch_corners.size() + corners.size());
         for (const auto& corner : corners){
             all_patch_corners.push_back(corner);
         }
-        all_patch_faces.reserve(faces.size());
+        all_patch_faces.reserve(all_patch_faces.size() + faces.size());
         for (const auto& face : faces){
             all_patch_faces.push_back(face);
         }
-        all_patch_cells.reserve(cells.size());
+        all_patch_cells.reserve(all_patch_cells.size() + cells.size());
         for (const auto& cell : cells){
             all_patch_cells.push_back(cell);
         }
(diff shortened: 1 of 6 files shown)


aritorto:

aritorto:

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1362,73 +1362,117 @@ template cpgrid::Entity<3> Dune::createEntity(const CpGrid&, int, bool);
...
-            OPM_THROW(std::logic_error, "Adding LGRs to a distributed grid is not supported, yet.");
-        }
-        else
-        {
-            OPM_THROW_NOLOG(std::logic_error, "Adding LGRs to a distributed grid is not supported, yet.");
+        OPM_THROW(std::logic_error, "Grid has been distributed. Cannot create LGRs.");
+    }
+    if (startIJK_vec.size() > 0){
+        if (!(*data_[0]).disjointPatches(startIJK_vec, endIJK_vec)){
+        OPM_THROW(std::logic_error, "Patches are not disjoint.");
```
blattms, line 1384: Here we should use the same approach as above:
```
if (startIJK_vec.size() > 0) && !(*data_[0]).disjointPatches(startIJK_vec, endIJK_vec){
        if (comm().rank()==0)
        {
            OPM_THROW(std::logic_error, "LGRs are not disjoint.");
        }
        else
        {
            OPM_THROW_NOLOG(std::logic_error, "LGRs are not disjoint.");
        }
}
```

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1362,73 +1362,117 @@ template cpgrid::Entity<3> Dune::createEntity(const CpGrid&, int, bool);
...
-        l0_parent_to_children_cells[cell] = no_children;
-    }
-    // For level1/LGR, attach to each child-cell its parent.
-    std::vector<std::array<int,2>>& l1_child_to_parent_cells = (*data_[1]).child_to_parent_cells_;
-    // For cells with no parent, we set {-1,-1}. Entries with actual parents will be rewritten.
-    const std::array<int,2>& no_parent = {-1,-1};
-    l1_child_to_parent_cells.resize(data_[1]-> size(0));
-    for (int cell = 0; cell < data_[1]->size(0); ++cell){
-        l1_child_to_parent_cells[cell] = no_parent;
+    const std::vector<int> no_children = {-1};
```
blattms, line 1405: Why isn't this an empty vector (representing no children)?

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1362,73 +1362,117 @@ template cpgrid::Entity<3> Dune::createEntity(const CpGrid&, int, bool);
...
-        l0_parent_to_children_cells[parent] = {1, children}; // {level LGR, {child0, child1, ...}}
-        for (const auto& child : children){
-            l1_child_to_parent_cells[child] = {0, parent}; // {level of parent cell, parent cell index in that level}
+    // Get patches corner, face, and cell indices.
+    for (int patch = 0; patch < num_patches; ++patch){
+        const auto& [corners, faces, cells] = (*(this->data_[0])).getPatchGeomIndices(startIJK_vec[patch], endIJK_vec[patch]);
+        all_patch_corners.reserve(corners.size());
+        for (const auto& corner : corners){
+            all_patch_corners.push_back(corner);
         }
```
blattms, line 1416: For the second patch the reserve call probably has little effect as the vector is not empty anymore. 

If that would be fixed then I think this does the same as if you would use the [insert](https://en.cppreference.com/w/cpp/container/vector/insert) function of std::vector `all_patch_corners.insert(all_patch_corners.end(),corners.begin(), corners.end());`. Maybe it is better to use that one)?

Similar for other for face, etc below.

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1362,73 +1362,117 @@ template cpgrid::Entity<3> Dune::createEntity(const CpGrid&, int, bool);
...
+        (this-> data_).push_back(level_ptr);
+        (*data_[patch +1]).level_data_ptr_ = &(this -> data_); // shifted +1 since level0 is coarse grid. Levels are 1,2,..., num_patches.
+        (*data_[patch +1]).level_ = patch +1;
+        //
+        // POPULATING (*data_[0]).parent_to_children_cells_
+        // POPULATING (*data_[patch +1]).child_to_parent_cells_
+        // Extension of container all_patch_child_to_parent_cells, adding {-1,-1} entries for each NO-CHILD cell, in each level/patch.
+        (all_patch_child_to_parent_cells[patch]).resize(child_to_parent_cells.size()); // To consider all cells in the LGR.
+        // True parent cells have entries: {level of the LGR the parent cell has its children, {child0, child1, ...}}
+        // False parent cells (NO PARENT CELLS) have {-1,{-1}} entries.
```
blattms, line 1439: Ah, I see. Similar to above my question is: Would `{-1, {}}` also work? Seems cleaner.

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1362,73 +1362,117 @@ template cpgrid::Entity<3> Dune::createEntity(const CpGrid&, int, bool);
...
-    std::vector<std::array<int,2>>& l1_child_to_parent_cells = (*data_[1]).child_to_parent_cells_;
-    // For cells with no parent, we set {-1,-1}. Entries with actual parents will be rewritten.
-    const std::array<int,2>& no_parent = {-1,-1};
-    l1_child_to_parent_cells.resize(data_[1]-> size(0));
-    for (int cell = 0; cell < data_[1]->size(0); ++cell){
-        l1_child_to_parent_cells[cell] = no_parent;
+    const std::vector<int> no_children = {-1};    
+    const std::tuple<int,std::vector<int>> has_no_children = std::make_tuple(-1, no_children);
+    for (int cell = 0; cell < data_[0]-> size(0); ++cell){
+        l0_parent_to_children_cells[cell] = has_no_children;
```
blattms, line 1408: I think we should use the default directly during resize above (line 1404), i.e. with my suggestion above
```
l0_parent_to_children_cells.resize(data_[0]-> size(0), {{-1, {}});
```

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1362,73 +1362,117 @@ template cpgrid::Entity<3> Dune::createEntity(const CpGrid&, int, bool);
...
+            = (*(this-> data_[0])).refinePatch(cells_per_dim_vec[patch], startIJK_vec[patch], endIJK_vec[patch]);
+        // Add each LGR to "data".
+        (this-> data_).push_back(level_ptr);
+        (*data_[patch +1]).level_data_ptr_ = &(this -> data_); // shifted +1 since level0 is coarse grid. Levels are 1,2,..., num_patches.
+        (*data_[patch +1]).level_ = patch +1;
+        //
+        // POPULATING (*data_[0]).parent_to_children_cells_
+        // POPULATING (*data_[patch +1]).child_to_parent_cells_
+        // Extension of container all_patch_child_to_parent_cells, adding {-1,-1} entries for each NO-CHILD cell, in each level/patch.
+        (all_patch_child_to_parent_cells[patch]).resize(child_to_parent_cells.size()); // To consider all cells in the LGR.
```
blattms, line 1437: Looking at this line and line I wonder whether we really need the all_patch_child_to_parent_cells container.
We can substitute `all_patch_child_to_parent_cells[patch]` with `level_child_to_parent_cells` an operate directly on that one. That would be defined by
```
auto& level_child_to_parent_cells = this-> data_.back()..child_to_parent_cells_ ;
```

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1362,73 +1362,117 @@ template cpgrid::Entity<3> Dune::createEntity(const CpGrid&, int, bool);
...
+        // Add this family information [child_to_parent_cells_] in each CpGridData object representing each LGR
+        (*data_[patch +1]).child_to_parent_cells_ =  all_patch_child_to_parent_cells[patch];
+        // shifted +1 since "GLOBAL" is level0 (coarse grid). Levels are 1,2,..., num_patches.
+        //
+        all_patch_boundary_old_to_new_corners.push_back(boundary_old_to_new_corners);
+        all_patch_boundary_old_to_new_faces.push_back(boundary_old_to_new_faces);
+    } // end-patch-forloop
+    // LEVEL 0, definition/declaration of some members:
+    (*data_[0]).level_data_ptr_ = &(this -> data_);
+    (*data_[0]).level_ = 0;
```
blattms, line 1464: I would hope that this not needed and already correctly initialized previously. Isn't that the case?

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1362,73 +1362,117 @@ template cpgrid::Entity<3> Dune::createEntity(const CpGrid&, int, bool);
...
     // Relation between level and leafview cell indices.
-    std::map<int,int>& l1_to_leaf_cells = (*data_[1]).level_to_leaf_cells_;
-    //
+    std::map<int,int>& l0_to_leaf_cells = (*data_[0]).level_to_leaf_cells_;
     // To store the leaf view (mixed grid: with (non parents) coarse and (children) refined entities).
     typedef Dune::FieldVector<double,3> PointType;
-    std::vector<std::shared_ptr<Dune::cpgrid::CpGridData>>& leaf_data = this -> data_;
+        std::vector<std::shared_ptr<Dune::cpgrid::CpGridData>>& leaf_data = this -> data_;
```
blattms, line 1469: Indentation seems wrong.

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1453,66 +1497,75 @@ void CpGrid::addLgrUpdateLeafView(const std::array<int,3>& cells_per_dim, const
...
-        old_to_new_boundaryPatchCorners[{0, boundary_old_to_new_corners[corner][0]}] = {1, boundary_old_to_new_corners[corner][1]};
-        boundary_patch_corners.push_back(boundary_old_to_new_corners[corner][0]);
+    std::vector<std::vector<int>> all_boundary_corners; // CHANGE NAME
+    all_boundary_corners.resize(num_patches);
+    for (int patch = 0; patch < num_patches; ++patch){
+        all_boundary_corners[patch].reserve(all_patch_boundary_old_to_new_corners[patch].size());
+        for (int corner = 0; corner < static_cast<int>(all_patch_boundary_old_to_new_corners[patch].size()); ++corner) {
+            old_to_new_boundaryPatchCorners[{0, all_patch_boundary_old_to_new_corners[patch][corner][0]}]
+                = {patch +1, all_patch_boundary_old_to_new_corners[patch][corner][1]};
+            // shifted +1 since coarse grid is level 0, levels/LGRs are 1,2, ..., num_patches.
```
blattms, line 1541: Comment does not fit next line.

blattms (commented):
thanks a lot.
First part of review until CpGrid.cpp: 1597

opm/grid/cpgrid/CpGrid.cpp
aritorto: l1_child_to_parent_cells is not empty in the end since it contains for each cell of the LGR (level 1, coming from refining a patch from level 0),  l1_child_to_parent_cells[child cell index] = {0, parent cell index} (0 representing level 0). The deleted lines are not needed since this container gets populated later with an actual parent for each entry. In contrary to leaf_child_to_parent_cells which may have entries with no parent, those are "by defualt set as" {-1, -1} representing "no level" and "no parent"

aritorto: I meant <any-level-from-1-to-last-lgr>.child_to_parent_cells_ behave as I described above (each cell has an actual parent in level 0, so each entry looks like {0, parent cell index}), and <leafview>.child_to_parent_cells_ may have entries {-1,-1} (instead of empty entries) when the cell does not have a parent.

opm/grid/cpgrid/CpGrid.cpp
aritorto: the comment refers to "patch +1" and not to  "..[patch]". I can modify it, in any case.

opm/grid/cpgrid/CpGrid.cpp
aritorto: I did the following modification:

`all_patch_corners.reserve(all_patch_corners.size() + corners.size());
        for (const auto& corner : corners){
            all_patch_corners.push_back(corner);
        }`
To enlarge the capacity of the vector in each step.

opm/grid/cpgrid/CpGrid.cpp
aritorto: Yes, {-1,{}} would work for "no-parent-cell" entries in parent_to_children_cells_

opm/grid/cpgrid/CpGrid.cpp
aritorto: For each actual level/LGR, child_to_parent_cells_ has entries {0 (level), parent cell index (in level 0) of that child-cell in that lgr}. For the LeafView, child_to_parent_cells_ has entries {-1,-1} when the cell "was born" in level 0 and has been not involved in any LGR, and {0, parent cell index in level 0} for the cells coming for each LGR. Since the leafview cell index and the "each-level/lgr" index of each cell are different, containers 
(*data_[level]).child_to_parent_cells_
and 
(*data_["leafview"]).child_to_parent_cells_
have different sizes and different entries. 

In the current code, the necessary information to populate child_to_parent_cells_ for the LeafView is deduced from the information created  previously for each level. It is possible to rewrite the code in such a way that the information is taken in the other direction (deduced from LeafView to create the each-level-information).  I would leave it like it is now.

opm/grid/cpgrid/CpGrid.cpp
aritorto: level_data_ptr_ was already correctly initialized previously. It wasn't the case for level_. Now both are correctly initialized so those lines are gone in the current code.

opm/grid/cpgrid/CpGrid.cpp
aritorto: corrected now

opm/grid/cpgrid/CpGrid.cpp
aritorto: Done!

opm/grid/cpgrid/CpGrid.cpp
aritorto: Added!

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1362,73 +1362,117 @@ template cpgrid::Entity<3> Dune::createEntity(const CpGrid&, int, bool);
...
 {
     if (!distributed_data_.empty()){
-        if (comm().rank()==0)
-        {
-            OPM_THROW(std::logic_error, "Adding LGRs to a distributed grid is not supported, yet.");
-        }
-        else
-        {
-            OPM_THROW_NOLOG(std::logic_error, "Adding LGRs to a distributed grid is not supported, yet.");
+        OPM_THROW(std::logic_error, "Grid has been distributed. Cannot create LGRs.");
```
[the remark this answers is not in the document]
aritorto: Corrected now.

opm/grid/cpgrid/CpGrid.cpp
aritorto: Similar modification done for chid_to_parent_cells_ for the leafview:

` // For cells that do not have a parent, we set {-1,-1} by defualt and rewrite later for actual children
    leaf_child_to_parent_cells.resize(cell_count, std::array<int,2>({-1,-1}));`

aritorto:

PR #646, opened 2023-04-19
Suppress some issues with geometryInFather().

atgeirr:
The current implementation cannot be instantiated, and must be suppressed until fixed. Also a few other issues, such as a wrong constructor call, indicates that the code is not instantiated at all under normal circumstances, and are so far only seen with dune-fem. However, the code is still wrong and must be fixed.

The geometryInFather() implementation has multiple issues that must be resolved before activating, I am available for discussion if desired.


atgeirr:

blattms:
Please open an issue with error log.

blattms:
Probably needs to be backported to the release.

PR #649, opened 2023-04-24
[bugfix] Prevent out-of-bounds indexing during pinch processing.

blattms:
If we are processing a cell at the bottom of the grid, then there will be no lower neighbor at all and certainly not for an NNC.

Previously we would compute the index of a cell in the next non-existing layer and sometimes would index containers actnum and thickness out of bounds.


blattms:
jenkins build this please

akva2 (approved):
Looks good to me

PR #651, opened 2023-05-02
changed: use Boost_VERSION_STRING

akva2:
Boost_VERSION is the version string x.y.z in some versions of cmake, and the boost xxyyzz-number in some cmake versions. Boost_VERSION_STRING is always the expected x.y.z format

sorry to  have to revisit this yet again. the cmake in epel seems to be a bit of a moving target.
it now sets the Boost_XXX variables as expected from modern FindBoost. That's good. However,
while it claims to be cmake 3.17 and the find module documentation says that Boost_VERSION is the same as Boost_VERSION_STRING starting in cmake 3.15, it is not on epel. it is still the Boost_VERSION_MACRO (ie internal boost version number). So I have to change it again.

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -104,7 +104,7 @@ list (APPEND TEST_SOURCE_FILES
   tests/test_compressed_cartesian_mapping.cpp
 	)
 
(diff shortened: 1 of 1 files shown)


akva2:
jenkins build this please

PR #653, opened 2023-05-02
[fix] Entity::geometryInFather() and HierarchicIterators fixed and tested.

aritorto:


aritorto:
jenkis build this serial please

PR #655, opened 2023-05-08
Reduce grid_equal() implementation verbosity

bska:
In particular, refactor the explicit calls to comparators out to a separate helper function, equal_vector<>(), that dispatches to std::equal() or Opm::cmp::array_equal<>() depending on the nature of the element type.  Integral types use the former and floating point types use the latter.

### DIFF
--- a/opm/grid/UnstructuredGrid.c
+++ b/opm/grid/UnstructuredGrid.c
@@ -17,7 +17,6 @@
   along with OPM.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-#include "config.h"
 #include <opm/grid/UnstructuredGrid.h>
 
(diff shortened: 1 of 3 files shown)


bska:
I've had this on my system for quite a while and now's a good time, I think, to get it into master.  We don't really use `UnstructuredGrid` anymore, but that said I still think this is easier to read.

bska:
jenkins build this please opm-common=3511

bska:
jenkins build this opm-common=3511 please

atgeirr:
Thanks, merging. The UnstructuredGrid struct is used by the PolyhedralGrid class, so it makes sense to maintain it.

PR #658, opened 2023-05-24
Correctly support item 3 of PINCH (MAX_EMPTY_GAP).

blattms:
MAX_EMPTY_GAP specifies the maximum thickness of gap between active cells allowed for creating NNCs. The default value is 1.0e20 in any unit system.

If PINCH is not specified, then the EclipseGrid will specify 1e20 in SI units. This should be big enough to account for infinity in a simulation and hence this implementation should work.

Fixes the following cases in opm-tests: T2A_NOPINCH, T2A_GAP, T2A_NOGAP

### DIFF
--- a/opm/grid/MinpvProcessor.hpp
+++ b/opm/grid/MinpvProcessor.hpp
@@ -76,6 +76,7 @@ namespace Opm
         /// will have the zcorn numbers changed so they are zero-thickness. Any
         /// cell below will be changed to include the deleted volume if mergeMinPCCells is true
         /// els the volume will be lost
+        /// \param[in]       tolerance_unique_points Tolerance used to identify points based on their cooridinates.
         Result process(const std::vector<double>& thickness,
                        const double z_tolerance,
                        const double max_gap,
@@ -87,7 +88,8 @@ namespace Opm
                        const bool pinchNOGAP = false,
                        const bool pinchOption4ALL = false,
                        const std::vector<double>& permz = {},
-                       const std::function<double(int)>& multZ = [](int){ return 0;}) const;
+                       const std::function<double(int)>& multZ = [](int){ return 0;},
+                       const double tolerance_unique_points = 0) const;
     private:
         double computeGap(const std::array<double,8>& coord_above, const std::array<double,8>& coord_below) const;
         std::array<int,8> cornerIndices(const int i, const int j, const int k) const;
@@ -130,7 +132,8 @@ namespace Opm
                                                           const bool pinchNOGAP,
                                                           const bool pinchOption4ALL,
                                                           const std::vector<double>& permz,
-                                                          const std::function<double(int)>& multz) const
+                                                          const std::function<double(int)>& multz,
+                                                          const double tolerance_unique_points) const
     {
         // Algorithm:
         // 1. Process each column of cells (with same i and j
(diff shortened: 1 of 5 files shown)


blattms:
jenkins build this please

blattms:
Sorry, seems like I am suffering from a premature-push-and-create-pr syndrome these days. :laughing:  Let's try again.

blattms:
Let's wait until the provided working pinch cases are under regression, see OPM/opm-simulators#4671

opm/grid/MinpvProcessor.hpp
```diff
@@ -332,6 +353,28 @@ namespace Opm
...
+                            if ((actnum.empty() || actnum[c_below]) && pv[c_below] > minpvv[c_below])
+                            {
+                                // Check MAX_GAP threshold
+                                std::array<double, 8> cz = getCellZcorn(ii, jj, kk, zcorn);
+                                std::array<double, 8> cz_below = getCellZcorn(ii, jj, kk_below, zcorn);
+
+                                if (computeGap(cz, cz_below) < max_gap) {
+                                    result.add_nnc(c, c_below);
+                                }
+                            }
```
blattms, line 375: Yeah, this is quite some overkill as we will add an NNC where there often will be a vertical connection between two active cells anyway.
This should probably move into the direct CpGrid generation where we decide whether there is a face between two cells based on whether their vertical faces coincide.

blattms: Alternatively, we could check here whether there is actually a gap to close with an NNC here. That would keep PINCH separated from the topology construction.

blattms: Fixed in [be8b720](https://github.com/OPM/opm-grid/pull/658/commits/be8b720f6fe0d59e2cc829b7a749d46fafe273a0) where we check whether there is actually a gap.

blattms:
The issue with recording NNCs where there would be a connection anyway is fixed now. Let's see whether this breaks other tests.

blattms:
jenkins build this please

blattms:
I will self-merge this tonight.

bska (approved):
I **think** this is correct, and the new test case does provide evidence for its correctness, but `MinpvProcessor::process()` is getting quite long by now and there's a lot of mutable state.  I think we should consider refactoring the cases slightly at some point.

For now I'll merge into master.

GitPaean:
Most likely this PR breaks the running of the master branch?

GitPaean:
reverting this PR for testing purpose.

opm/grid/MinpvProcessor.hpp
```diff
@@ -333,6 +357,34 @@ namespace Opm
...
+
+                            if ((actnum.empty() || actnum[c_below]) && pv[c_below] > minpvv[c_below])
+                            {
+                                // Check MAX_GAP threshold
+                                std::array<double, 8> cz = getCellZcorn(ii, jj, kk, zcorn);
+                                std::array<double, 8> cz_below = getCellZcorn(ii, jj, kk_below, zcorn);
+                                bool vertically_connected = true; // If true a connection will be there anyway -> Skip NNC
+
+                                for(int i = 0; i < 4; ++i) {
+                                    vertically_connected = vertically_connected && std::abs(cz_below[i] - cz[4+1])
```
blattms, line 378: :flushed: Yeah, this is probably not what we should do or I even wanted to do (simple typo of `4+1` instead of `4+i` with big consequences). Seems like my (simple) tests are still lacking...

There you sit in gdb staring add the grid preprocessing and wonder about the magic in uniquepoints.c and then you have to realize (accidentally) that it is here and something embarrassing like this.

I'll create a new PR...

PR #659, opened 2023-05-26
Correctly process inactive thin cell not next to pinched out cells.

blattms:
Pinch processing only started when active cells were pinched out due  to pore volume being below the threshold. Over Thin inactive cells, that were processed in the same iteration might, NNCs might be created. The problem was that we never created an NNC over thin inactive cells without a pinched out neighbor.
    
With this patch we do that now and collapse these inactive cells, too.

Fixes test cases T4A1_GAP and T4A1_NOGAP.

This will conflict with #658 and needs to be rebased once that is merged.

### DIFF
--- a/opm/grid/MinpvProcessor.cpp
+++ b/opm/grid/MinpvProcessor.cpp
@@ -243,8 +243,8 @@ MinpvProcessor::process(const std::vector<double>& thickness,
                             kk = kk_iter;
                             continue;
                         }
-                        // top or bottom cell not active, hence no ncc is created
-                        if (!actnum.empty() && (!actnum[c] ||  !actnum[c_below])) {
+                        // bottom cell not active, hence no ncc is created
+                        if (!actnum.empty() && !actnum[c_below]) {
                             kk = kk_iter;
                             continue;
                         }
--- a/opm/grid/MinpvProcessor.cpp
+++ b/opm/grid/MinpvProcessor.cpp
@@ -243,13 +243,14 @@ MinpvProcessor::process(const std::vector<double>& thickness,
                             kk = kk_iter;
                             continue;
                         }
-                        // bottom cell not active, hence no ncc is created
+                        // bottom cell not active, hence no nnc is created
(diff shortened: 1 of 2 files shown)


blattms:
jenkins build this please

blattms:
Sigh, seems like this this breaks some of the NOPINCH cases.

blattms:
Let's wait until the provided working pinch cases are under regression, see OPM/opm-simulators#4671

blattms:
jenkins build this please

blattms:
rebased, just waiting for OPM/opm-simulators#4722

blattms:
jenkins build this please

bska:
Is there more work coming here?

blattms:
Not in this PR.

Concerning PINCH there is still one scenario that is not supported (and throws an error). But that is much more complicated and for another PR.

bska:
>> Is there more work coming here?
>
> Not in this PR.

Cool!  I'll do final review here, then.

> Concerning PINCH there is still one scenario that is not supported (and throws an error). But that is much more complicated and for another PR.

Noted.  I look forward to seeing that when you're ready.

opm/grid/MinpvProcessor.cpp
```diff
@@ -243,8 +243,8 @@ MinpvProcessor::process(const std::vector<double>& thickness,
                             kk = kk_iter;
                             continue;
                         }
-                        // top or bottom cell not active, hence no ncc is created
-                        if (!actnum.empty() && (!actnum[c] ||  !actnum[c_below])) {
+                        // bottom cell not active, hence no ncc is created
```
akva2, line 246: nnc

bska (approved):
Thanks.  I think this looks good.  Other than the minor misprint identified earlier ("ncc" :arrow_right: "nnc"), I don't see any issues here.

blattms:
Well seems like I am challenged by NNCs... even in branch names :see_no_evil: Will fix and self-merge

blattms:
jenkins build this please

PR #662, opened 2023-05-31
Add some more compile units

akva2:
Likely very low impact but every little drop.


akva2:
jenkins build this please

PR #663, opened 2023-06-01
Looking up data of leaf cells from level 0 grid supported

aritorto:
Given an element in the LeafView and a container of properties, the property of its origin-cell (parent cell, if exists) is returned, in the case of (multiple LGRs) CpGrid. For general grids, the given properties are assumed to be on the leafview. 

Fix CpGridData::disjointPatches(), test added.

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -110,8 +110,11 @@ list (APPEND TEST_SOURCE_FILES
 
 if(Boost_VERSION_STRING VERSION_GREATER 1.53)
 	list(APPEND TEST_SOURCE_FILES
+	  tests/cpgrid/disjointPatches_test.cpp
 	  tests/cpgrid/geometry_test.cpp
 	  tests/cpgrid/grid_lgr_test.cpp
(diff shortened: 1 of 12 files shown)


aritorto:

aritorto:

bska (commented):
Am I reading this correctly, that the new `LookupData<>` template isn't used anywhere within this PR&mdash;other than in its unit tests, of course?

aritorto:
Correct. The idea is to use it in opm-simulators/ebos, e.g. eclgenericthresholdpressure.cc, eclgenericproblem.cc (WIP on branch lookupdata-generalgrid, in opm-simulators).

bska:
>> Am I reading this correctly, that the new `LookupData<>` template isn't used anywhere within this PR
>
> Correct. The idea is to use it in opm-simulators/ebos, e.g. eclgenericthresholdpressure.cc, eclgenericproblem.cc

Ah, okay.  In that case I think this is good and I'll merge into master.  Sorry about the delay and thanks a lot for doing this work!

PR #664, opened 2023-06-13
Revert "Correctly support item 3 of PINCH (MAX_EMPTY_GAP)."

GitPaean:
Reverts OPM/opm-grid#658

### DIFF
--- a/opm/grid/MinpvProcessor.hpp
+++ b/opm/grid/MinpvProcessor.hpp
@@ -27,7 +27,6 @@
 #include <functional>
 #include <map>
 #include <vector>
-#include <numeric>
 
 namespace Opm
 {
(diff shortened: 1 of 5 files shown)


GitPaean:
jenkins build this please

GitPaean:
I am okay with a fix from the current master branch or merging this PR before fixing the actual problem. But I think the jenkins test should be recovered as soon as possible.

atgeirr:
@blattms Do you understand what went wrong, and can you fix this quickly? If it can be done within an hour or so we should avoid the revert, otherwise we should revert and redo the PR (with the as of now unknown fix).

blattms:
When was that VAPPARS test added?

No I have no idea what happens here and why. I would need to take a look in the debugger and that takes time. I will try to take a look now.
Feel free to revert, but then I expect a timely review for the next PR. Lately, they have been hanging for quite some while....

bska:
> but then I expect a timely review for the next PR. Lately, they have been hanging for quite some while....

Pot meet kettle...

blattms:
Yeah, maybe it would be good idea to just ask for reviews on PRs. (At least that is what I often check if I am time pressed). That would make it easier to find the raisins between messages about e.g. branches just being force-pushed because of rebasing or impatient people announcing "I will self-merge tonight".

PR #665, opened 2023-06-14
Second try to support  item 3 of PINCH (MAX_EMPTY_GAP)

blattms:
This #658 again with a  fix for the embarassing fallout due to a typo. My apologies for the fallout. Also added a test that would have found that issue.

Original text:
"
MAX_EMPTY_GAP specifies the maximum thickness of gap between active cells allowed for creating NNCs. The default value is 1.0e20 in any unit system.

If PINCH is not specified, then the EclipseGrid will specify 1e20 in SI units. This should be big enough to account for infinity in a simulation and hence this implementation should work.

Fixes the following cases in opm-tests: T2A_NOPINCH, T2A_GAP, T2A_NOGAP
"

### DIFF
--- a/opm/grid/MinpvProcessor.hpp
+++ b/opm/grid/MinpvProcessor.hpp
@@ -375,7 +375,7 @@ namespace Opm
                                 bool vertically_connected = true; // If true a connection will be there anyway -> Skip NNC
 
                                 for(int i = 0; i < 4; ++i) {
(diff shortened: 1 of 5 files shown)


blattms:
benchmark please

blattms:
jenkins build this please

bska (approved):
Thanks a lot for the updates.  I'll merge into master.

blattms:
Perfect. Thanks a lot.

PR #670, opened 2023-07-18
LookUpData and LookUpCartesianData classes added

aritorto:
LookUpData and LookUpCartesianData are classes to search data via element Index and Cartesian Index respectively. Given a feature vector and an element in the grid view, the data of such an element can be accessed through instances of these classes. 

For CpGrid, the feature vector is assumed to be given for level 0 grid. 
For any other type of grid, the feature vector is assumed to be given in its grid leaf view. 
These classes will be used in opm-simulators, to access data.

Tested for CpGrid and PolyhedralGrid.


aritorto:

aritorto:

aritorto:

opm/grid/LookUpData.hh
```diff
@@ -14,51 +14,242 @@
...
+    //                            [For CpGrid, the feature vector is given for level 0.]
+    template<typename FeatureType>
+    FeatureType operator()(const int& elemIdx, const std::vector<FeatureType>& feature_vec) const;
+
+    // getOriginIndexFromEntity(Entity) For general grids: retunrs the same Entity index.
+    template<typename EntityType, typename GridType = Grid>
+    typename std::enable_if_t<!std::is_same_v<GridType,Dune::CpGrid>,int> getOriginIndexFromEntity(const EntityType& elem) const;
+
+    // getOriginIndexFromEntity(Entity) For CpGrid: returns index of origin cell (parent/equivalent cell when no father) in level 0.
+    template<typename EntityType, typename GridType = Grid>
```
blattms, line 80: There seems to be no need for the template parameter GridType. You seem to be misusing it for `using GridType = Grid`. The same holds below.

Maybe we can skip it and either add a `using` directive or simply, rename the template parameter of the class or use Grid instead of GridType in `is_same_v

opm/grid/LookUpData.hh
```diff
@@ -14,51 +14,242 @@
...
 {
-template <typename GridType>
+class CpGrid;
+
+namespace cpgrid
+{
+template<int codim> class Entity;
+}
+
+// LookUpData class - To search data via element index
```
blattms, line 49: I think it would make things easier if we put LookupData into the Opm namespace. Hence close the other one here and open Opm.

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -559,12 +559,26 @@ void CpGrid::createCartesian(const std::array<int, 3>& dims,
 
 const std::array<int, 3>& CpGrid::logicalCartesianSize() const
 {
-    return current_view_data_->logical_cartesian_size_;
+    // Temporary. For a grid with LGRs, we set the logical cartesian size of the LeafGridView as the one for level 0.
+    //            Goal: CartesianIndexMapper well-defined for CpGrid LeafView with LGRs.
+    if (current_view_data_ == this-> data_.back().get()){
+        return this -> data_[0] -> logical_cartesian_size_;
+    }
+    else{
```
blattms, line 567: Maybe we should make this future-proof (for the parallel version) and use `this -> distributed_data_[0]->...` in the else branch? Same below

blattms (commented):
This looks pretty good. I added some comments.

One thing that strikes me, is that now we kind of two specializations and duplicate code:
1. In the unspecialized version you have the enable_if constructs to do the right thing also in the case that the grid is CpGrid.
2. in addition you have the specialized version for CpGrid.

Hence we kind of have implemented things twice for CpGrid. Maybe we should get rid of one? E.g. keep the enable_if stuff and remove the specialization or the other way around.

We should add a test using PolyhedralGrid with LookupData.

blattms:
There is  a test for PolyhedralGrid in tests/test_polyhedralgrid.cpp where you can see how to construct such a grid.

opm/grid/LookUpData.hh
aritorto: I don't understand yet why this auxiliary template parameter is needed, but if I remove it, the code breaks. The error messages said "cannot overload ...". If I add it only in one method, another error shows up: "no type named ‘type’ in ‘struct std::enable_if<false, int>". 
So, for the moment, I keep this version with the extra auxiliary template parameter, coinciding with 'Grid', that allows the overloading of each method that behaves differently for CpGrid and other type of Grids.

opm/grid/LookUpData.hh
aritorto: Done!

opm/grid/cpgrid/CpGrid.cpp
aritorto: This suggestion breaks the following tests: distribution, distribution_test_parallel,  and lookupdata tests. I'm keeping data_[0] for now. I can have a deeper look at the potential modifications in other files, to make this change [data_ replaced by distributed_data_] work.

opm/grid/LookUpData.hh
blattms: I see. We have discussed this privately and as the average-developer joe is not not bothered by it (no need to provide the template parameter), we can leave it as is from my side.

But I think we should do two things:
1. Add a `static_assert` that Grid is the expected type, to prevent people from using a custom grid type
2. Add documentation why it is like it is. Otherwise many people will wonder about this. This can either be done with doxygen using `\tparam` or as a regular comment about the doxygen docu.
3. Add a short comment above the class outlining that we use enable_if instead of specializing 

This would be very cool to have and help the colleagues.

blattms (commented):
I am just going through the changes regarding previous review comments. I'll have another deeper look later today.

Thanks a lot, especially for the check with PolyhedralGrid. This looks already very good on a few first looks.

Before merging, we might want to get rid off the "WIP" commit, but let's do that last, maybe.

tests/test_lookupdata_polyhedral.cpp
```diff
@@ -0,0 +1,200 @@
...
+        BOOST_CHECK(idx == (lookUpData.getOriginIndex<Dune::PolyhedralGrid<3,3>>(idx)));
+        BOOST_CHECK(featureInElemIDX == featureInElem);
+        BOOST_CHECK(featureInElemDoubleIDX == featureInElemDouble);
+        BOOST_CHECK(featureInElemCartesianIDX == featureInElemCartesian);
+        BOOST_CHECK(featureInElemDoubleCartesianIDX == featureInElemDoubleCartesian);
+    }
+
+}
+
+/*TO BE ADDED ONCE GLOBALCELL ISSUE IS SOLVED FOR POLYHEDRALGRID.
```
blattms, line 123: How about removing this commented out code?

blattms (commented):
Looks good apart from some minor things. Please note that the comments about doxygen apply to several places.

One question, though.
There are three tests:
1. lookupdataCpGrid_test.cpp
2.  lookupdata_general_test.cpp
3. test_lookupdate_polyhedral_test.cpp

Does the second one differ from the other two (or is it a remnant from the previous specialisation approach)?

aritorto:

blattms:
Why on earth did I not merge this one? It is independent of the changes in opm-simulators. One more jenkins run and in it goes.

blattms:
jenkins build this please

PR #673, opened 2023-08-08
removing the duplicated AlignAfterOpenBracket entries in .clang-format

GitPaean:
removing a warning when using an IDE. 

same with https://github.com/OPM/opm-common/pull/3616 and will self-merge.


GitPaean:
jenkins build this please

PR #675, opened 2023-08-24
Reclaim Memory Resources for Parallel Interface Map

bska:
We were not calling the `InterfaceInformation::free()` function for the scatter and gather interfaces backing the cell and point communication in parallel.  The situation was mitigated by the fact that the `CpGrid` object is destroyed very late in an exiting run at which point the operating system would reclaim the memory anyway.

Implementation taken from equivalent code in `CpGridData`.

### DIFF
--- a/opm/grid/cpgrid/CpGrid.cpp
+++ b/opm/grid/cpgrid/CpGrid.cpp
@@ -115,6 +115,42 @@ void setupRecvInterface(const std::vector<std::tuple<int, int, char, int> >& lis
     }
 }
 #endif // HAVE_MPI
+
+/// Release memory resources from CpGrid::InterfaceMap.  Used as custom
+/// deleter for std::shared_ptr<InterfaceMap>.
+struct FreeInterfaces
+{
+#if !HAVE_MPI
+
+    /// Release memory resources for InterfaceMap object
+    ///
+    /// \param[in] interfaces Object for which to release memory resources.
+    void operator()([[maybe_unused]] Dune::CpGrid::InterfaceMap* interfaces) const
+    {
+        // Nothing to do in the sequential case as the CpGrid::InterfaceMap
+        // handles interface deletion in its destructor in this case.
+    }
+
+#else // HAVE_MPI
+
+    /// Release memory resources for InterfaceMap object
+    ///
+    /// \param[in] interfaces Object for which to release memory resources.
+    void operator()(Dune::CpGrid::InterfaceMap* interfaces) const
+    {
+        if (interfaces == nullptr) {
+            return;
+        }
+
+        for (auto& interface : *interfaces) {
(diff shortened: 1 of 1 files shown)


bska:
jenkins build this please

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -115,6 +115,30 @@ void setupRecvInterface(const std::vector<std::tuple<int, int, char, int> >& lis
     }
 }
 #endif // HAVE_MPI
+
+/// Reclaim memory resources from CpGrid::InterfaceMap.  Used as custom
+/// deleter for std::shared_ptr<InterfaceMap>.
+struct FreeInterfaces
+{
+    void operator()([[maybe_unused]] Dune::CpGrid::InterfaceMap* interfaces) const
```
akva2, line 123: I think #if around the definition rather than in the body would be clearer (and lets you skip maybe_unused when we know it won't be used which i personally prefer). It better communicates the do-nothing-without-mpi.

bska: > I think #if around the definition rather than in the body would be clearer

You mean as in
```C++
struct FreeIntefaces
{
#if MPI
    void operator()(...)
    {
        ...
    }
#endif
};
```
?  I don't think that'll work without some other changes.  In that case we'd need to either distinguish between the MPI and the non-MPI cases at [the call site](https://github.com/OPM/opm-grid/blob/3de8a0309ef3a622544f4d1fd1acc3bb257d409d/opm/grid/cpgrid/CpGrid.cpp#L150-L151) so as not to provide a custom deleter in the sequential case, or we'd need an alternative implementation of `FreeInterfaces::operator()` for the sequential case.

akva2: i mean
```c++
#if HAVE_MPI
void operator(Dune::CpGrid::InterfaceMap* interfaces) const
{
..
}
#else
void operator()Dune::CpGrid::InterfaceMap*) const {}
#endif
```

bska: > i mean [an alternative implementation of `FreeInterfaces::operator()` for the sequential case.]

Right.  To be honest I'd prefer not to do that.

akva2: fair enough.

bska:
I've switched to using a separate `FreeInterfaces::operator()(InterfaceMap*) const` function for the sequential case.

bska:
jenkins build this please

akva2 (approved):
changed your mind or trying to please me? i was not at all insisting. in any case this looks (or the original version) looks good to me.

atgeirr:
Looks fine to me as well, merging.

PR #676, opened 2023-08-31
Modified centroids lookup: separated function members 

aritorto:
In #672, a new structure to search for cell centroids was added, however  its application generated issues in OPM/opm-simulators#4778. 

The new approach is to have two separated function members to compute centroids: one via Eclipse grid and the other for CpGrid. 

The issue seems to be related to the fact that, even when the Grid template parameter of EclBaseVanguard is a CpGrid, centroids might be computed via its associated EclipseGrid, when rank is zero. This is just my interpretation of several errors messages, which can be wrong.


aritorto:

PR #680, opened 2023-10-12
Search Cartesian Index of origin cell

aritorto:
New methods added (in LookUpCartesianData) to search the Cartesian Index of the origin cell (parent/equivalent cell in level 0) for each cell in the leaf grid view. 
The methods are tested for CpGrid with/without LGRs and for PolyhedralGrid. 

This will play a key role in upcoming modifications in opm-simulators/ebos files, to support LGRs.


aritorto:

PR #681, opened 2023-10-20
Lookup of field properties via strings (LGRs data supported)

aritorto:
The class to search for field properties (to be used in opm-simulators/ebos) now takes into account that field properties can be also given for some LGRs (refined level). 

New methods added to search (int/double) field properties via FieldPropsManager and strings. Small test case added for PolyhedralGrid and CpGrid.

### DIFF
--- a/opm/grid/LookUpData.hh
+++ b/opm/grid/LookUpData.hh
@@ -72,14 +72,6 @@ public:
     {
     }
 
-    /// \brief: Get field propertry for an element in the leaf grid view, from a vector.
-    ///
-    ///         For general grids, the field property vector is assumed to be given for the gridView_.
-    ///         For CpGrid, the field property vector is assumed to be given for level 0 when isFieldPropLgr_ == false,
-    ///         and for certain LGR/level > 0 when isFieldPropLgr_ == true.
-    template<typename EntityType, typename FieldPropType>
-    FieldPropType operator()(const EntityType& elem, const std::vector<FieldPropType>& fieldProp) const;
-
     /// \brief: Get field propertry for an element in the leaf grid view, from a vector and element index.
     ///
     ///         For general grids, the field property vector is assumed to be given for the gridView_.
@@ -87,6 +79,15 @@ public:
     ///         and for certain LGR/level > 0 when isFieldPropLgr_ == true.
     template<typename FieldPropType>
     FieldPropType operator()(const int& elemIdx, const std::vector<FieldPropType>& fieldProp) const;
+
+    /// \brief: Get field propertry for an element in the leaf grid view, from a vector.
+    ///
+    ///         For general grids, the field property vector is assumed to be given for the gridView_.
+    ///         For CpGrid, the field property vector is assumed to be given for level 0 when isFieldPropLgr_ == false,
+    ///         and for certain LGR/level > 0 when isFieldPropLgr_ == true.
+    template<typename EntityType, typename FieldPropType>
+    typename std::enable_if_t<!std::is_same_v<EntityType, unsigned int>, FieldPropType>
+    operator()(const EntityType& elem, const std::vector<FieldPropType>& fieldProp) const;
 
     /// \brief: Get property of type double from field properties manager by name, via element or its index.
     template<typename ElemOrIndex>
@@ -100,15 +101,15 @@ public:
                      const std::string& propString,
                      const ElemOrIndex& elemOrIndex) const;
 
-    /// \brief: Retunrs the same element index for all grids different from CpGrid.
+    /// \brief: Return the same element index for all grids different from CpGrid.
     ///
     /// \tparam     GridType    Auxiliary type to overload the method, distinguishing general grids from CpGrid, with std::enable_if.
     ///                         Default: GridType = Grid.
     template<typename EntityType, typename GridType = Grid>
-    typename std::enable_if_t<!std::is_same_v<GridType,Dune::CpGrid>,int>
+    typename std::enable_if_t<!std::is_same_v<GridType,Dune::CpGrid> && !std::is_same_v<EntityType, unsigned int>,int>
     getFieldPropIdx(const EntityType& elem) const;
 
-    /// \brief: Returns index to search for the field propertries, for CpGrids.
+    /// \brief: Return index to search for the field propertries, for CpGrids.
     ///
     ///         When isFieldPropInLgr_ == false : fieldPropIdx == Index of the origin cell (parent/equivalent cell when element
     ///                                           has nofather) in level 0.
@@ -117,10 +118,10 @@ public:
     /// \tparam     GridType    Auxiliary type to overload the method, distinguishing general grids from CpGrid, with std::enable_if.
     ///                         Default: GridType = Grid.
     template<typename EntityType, typename GridType = Grid>
-    typename std::enable_if_t<std::is_same_v<GridType,Dune::CpGrid>,int>
+    typename std::enable_if_t<std::is_same_v<GridType,Dune::CpGrid> && !std::is_same_v<EntityType, unsigned int>,int>
     getFieldPropIdx(const EntityType& elem) const;
 
-    /// \brief: Retunrs the same element index for all grids different from CpGrid.
+    /// \brief: Return the same element index for all grids different from CpGrid.
     ///
     /// \tparam     GridType    Auxiliary type to overload the method, distinguishing general grids from CpGrid, with std::enable_if.
     ///                         Default: GridType = Grid.
@@ -128,7 +129,7 @@ public:
     typename std::enable_if_t<!std::is_same_v<GridType,Dune::CpGrid>,int>
     getFieldPropIdx(const int& elemIdx) const;
 
-    /// \brief: Returns the index to search for the field properties, for CpGrids.
+    /// \brief: Return the index to search for the field properties, for CpGrids.
     ///
     ///         When isFieldPropInLgr_ == false : fieldPropIdx == Index of the origin cell (parent/equivalent cell when element
     ///                                           has nofather) in level 0.
@@ -175,21 +176,22 @@ public:
     {
     }
 
-    /// \brief: Get field propertry for an element in the leaf grid view, from a vector, via Cartesian Index.
-    ///
-    ///         For general grids, the field property vector is assumed to be given for the gridView_.
-    ///         For CpGrid, the field property vector is assumed to be given for level 0 when isFieldPropLgr_ == false,
-    ///         and for certain LGR/level > 0 when isFieldPropLgr_ == true.
-    template<typename EntityType, typename FieldPropType>
-    FieldPropType operator()(const EntityType& elem,const std::vector<FieldPropType>& fieldProp) const;
-
-    /// \brief: Get field propertry for an element in the leaf grid view, from a vector, via Cartesian Index.
+    /// \brief: Get field property for an element in the leaf grid view, from a vector, via Cartesian Index.
     ///
     ///         For general grids, the field property vector is assumed to be given for the gridView_.
     ///         For CpGrid, the field property vector is assumed to be given for level 0 when isFieldPropLgr_ == false,
     ///         and for certain LGR/level > 0 when isFieldPropLgr_ == true.
     template<typename FieldPropType>
     FieldPropType operator()(const int& elemIdx, const std::vector<FieldPropType>& fieldProp) const;
+
+    /// \brief: Get field property for an element in the leaf grid view, from a vector, via Cartesian Index.
+    ///
+    ///         For general grids, the field property vector is assumed to be given for the gridView_.
+    ///         For CpGrid, the field property vector is assumed to be given for level 0 when isFieldPropLgr_ == false,
+    ///         and for certain LGR/level > 0 when isFieldPropLgr_ == true.
+    template<typename EntityType, typename FieldPropType>
+    typename std::enable_if_t<!std::is_same_v<EntityType, unsigned int>, FieldPropType>
+    operator()(const EntityType& elem,const std::vector<FieldPropType>& fieldProp) const;
 
     /// \brief: Get property of type double from field properties manager by name, via element or its index.
     template<typename ElemOrIndex>
@@ -203,15 +205,15 @@ public:
                      const std::string& propString,
                      const ElemOrIndex& elemOrIndex) const;
 
-    /// \brief: Retunrs the same element index for all grids different from CpGrid.
+    /// \brief: Return the same element index for all grids different from CpGrid.
     ///
     /// \tparam     GridType    Auxiliary type to overload the method, distinguishing general grids from CpGrid, with std::enable_if.
     ///                         Default: GridType = Grid.
     template<typename EntityType, typename GridType = Grid>
-    typename std::enable_if_t<!std::is_same_v<GridType,Dune::CpGrid>,int>
+    typename std::enable_if_t<!std::is_same_v<GridType,Dune::CpGrid>  && !std::is_same_v<EntityType, unsigned int>,int>
     getFieldPropCartesianIdx(const EntityType& elem) const;
 
-    /// \brief: Returns index to search for the field propertries, for CpGrids.
+    /// \brief: Return index to search for the field propertries, for CpGrids.
     ///
     ///         When isFieldPropInLgr_ == false : fieldPropIdx == Index of the origin cell (parent/equivalent cell when element
     ///                                           has nofather) in level 0.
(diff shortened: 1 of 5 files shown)


aritorto:

aritorto:
jenkins build this opm-simulators=4942 serial please

blattms:
I did not have a thorough look. That will come later.

Just my immediate recaction from a short glimpse at this:
Very cool. Especially the test with FieldPropertyManager.  Would it make sense to also have a similar test (same deck  string, EclipseGrid and FieldPropertyManager), but CpGrid with an LGR?

aritorto:
There is one example/case with string, eclipse grid, and field props manager, but the same string cannot be used (therefore the example for CpGrid has a different one). The reason is that some necessary keywords in the string were missing in the deck-string used for PolyhedralGrid.

aritorto:
I can try to add one. I hope the missing connection between opm-common LGRs and opm-grid LGRs is not needed. Thanks!

opm/grid/LookUpData.hh
```diff
@@ -154,104 +232,160 @@ template<typename Grid, typename GridView>
...
         gridView_(gridView),
         elemMapper_(gridView, Dune::mcmgElementLayout()),
-        cartMapper_(&mapper)
+        cartMapper_(&mapper),
+        isFieldPropInLgr_(isFieldPropInLgr)
     {
     }
 
-    /// \brief: Call operator taking an EntityObject and a FeatureVector.
+    /// \brief: Call operator taking an EntityObject, a FieldPropVector, and a level.
```
blattms, line 253: No level, right?

opm/grid/LookUpData.hh
```diff
@@ -261,131 +395,222 @@ protected:
...
 
 template<typename Grid, typename GridView>
 template<typename EntityType, typename GridType>
 typename std::enable_if_t<std::is_same_v<GridType,Dune::CpGrid>,int>
-Opm::LookUpData<Grid,GridView>::getOriginIndexFromEntity(const EntityType& elem) const
+Opm::LookUpData<Grid,GridView>::getFieldPropIdxFromEntity(const EntityType& elem) const
 {
     static_assert(std::is_same_v<Grid,GridType>);
-    return elem.getOrigin().index();
+    if (isFieldPropInLgr_ && elem.level()) { // level > 0 == true ; level == 0 == false
```
blattms, line 471: One question because of the `elemlevel()` check: Wouldn't `elem.getLgrElem()` also do the correct thing if the entity is not refined?

opm/grid/cpgrid/Entity.hpp
```diff
@@ -566,6 +574,38 @@ Dune::cpgrid::Entity<0> Dune::cpgrid::Entity<codim>::getOrigin() const
...
+    // Check that the element belongs to the leaf grid view
+    if (!(pgrid_ -> leaf_to_level_cells_.empty())) // entity on the LeafGridView
+    {
+        const auto entityLevel = this -> level(); // pgrid_->leaf_to_level_cells_[this->index()][0]
+        const int& entityLgrIdx = pgrid_->leaf_to_level_cells_[this->index()][1]; // leaf_to_level_cells_ [leaf idx] = {level, cell idx}
+        const auto& lgr_grid = (*(pgrid_ -> level_data_ptr_))[entityLevel].get();
+        return Dune::cpgrid::Entity<0>( *lgr_grid, entityLgrIdx, true);
+    }
+    else {
+        throw std::invalid_argument("The entity provided does not belong to the leaf grid view. ");
```
blattms, line 589: Would it make sense to return `*this` here? 

My understanding is that we return the entity of a level grid view that is equivalent to he one on the level. If there is no refinement, then they are the same

opm/grid/cpgrid/Entity.hpp
```diff
@@ -266,6 +266,14 @@ class Entity : public EntityRep<codim>
     /// Otherwise, returns itself. 
     Entity<0> getOrigin() const;
 
+    /// getLgrElem()   ONLY FOR Entities in the leaf grid view.
+    /// Returns refined entity in level > 0 (equivalent to the leaf grid view entity), if the entity was born in any LGR.
+    Entity<0> getLgrElem() const;
```
blattms, line 271: This returns the equivalent Element on the level grid view. What about naming the function getLevelElem?

In that case we might also use "\brief Get equivalent element on the level grid view on the level grid view"

opm/grid/cpgrid/Entity.hpp
```diff
@@ -566,6 +574,38 @@ Dune::cpgrid::Entity<0> Dune::cpgrid::Entity<codim>::getOrigin() const
     }
 }
 
+template<int codim>
+Dune::cpgrid::Entity<0> Dune::cpgrid::Entity<codim>::getLgrElem() const
+{
+    // Check that the element belongs to the leaf grid view
```
blattms, line 580: If this only work for leaf elements then we should throw if the element is not on the leaf

opm/grid/cpgrid/Entity.hpp
```diff
@@ -566,6 +574,38 @@ Dune::cpgrid::Entity<0> Dune::cpgrid::Entity<codim>::getOrigin() const
...
+    }
+    else {
+        throw std::invalid_argument("The entity provided does not belong to the leaf grid view. ");
+    }
+}
+
+template<int codim>
+int Dune::cpgrid::Entity<codim>::getLgrCartesianIdx() const
+{
+    // Check that the element belongs to the leaf grid view
```
blattms, line 596: If this only work for leaf elements then we should throw if the element is not on the leaf

opm/grid/cpgrid/Entity.hpp
```diff
@@ -266,6 +266,14 @@ class Entity : public EntityRep<codim>
     /// Otherwise, returns itself. 
     Entity<0> getOrigin() const;
 
+    /// getLgrElem()   ONLY FOR Entities in the leaf grid view.
+    /// Returns refined entity in level > 0 (equivalent to the leaf grid view entity), if the entity was born in any LGR.
+    Entity<0> getLgrElem() const;
+
+    /// getLgrCartesianIdx()   ONLY FOR Entities in the leaf grid view.
```
blattms, line 273: getLevelCartesianIndex?
"\brief Get cartesian index on the level grid view"

tests/cpgrid/lookupdataCpGrid_test.cpp
```diff
@@ -229,3 +266,121 @@ BOOST_AUTO_TEST_CASE(no_lgrs_grid)
...
+1*2 1*2 1*2 1*2 1*2
+/
+TOPS
+40*1
+/
+ACTNUM
+5*1
+/
+PORO
+5*0.15
```
blattms, line 366: As we are checking the values in our tests, we should make sure that they differ for different cells, by using e.g.
```
PORO
1.0 2.0 3.0 4.0 5.0
/
```

If you decide to also check an in propery, then you could use EQLNUM in addition by adding this at the bottom
```
PROPS

EQLNUM
1 2 3 4 5
/
```

```

**Review by blattms — CHANGES_REQUESTED** [review]
thanks a lot for the good work. Also for already adding the test with LGRS. This looks already really nice.

I now did a decent review and have some comments. Hopefully I was not too picky.

File: opm/grid/LookUpData.hh [hunk 9]
```diff
@@ -63,88 +70,159 @@ template <typename Grid, typename GridView>
...
+    /// \brief: fieldPropDouble() taking a FieldPropsManager, std::string, and an element.
+    ///                           Return field property of type double of the entity, via (ACTIVE) INDEX
+    ///
+    /// \tparam     EntityType          Element type.
+    /// \param [in] FieldPropsManager
+    /// \param [in] propString          String to lookup the field property in the FieldPropsManager
+    /// \param [in] element             EntityType object.
+    /// \return field property of the given leaf grid view element.
+    template<typename EntityType>
+    double fieldPropDoubleFromEntity(const FieldPropsManager& fieldPropsManager,
```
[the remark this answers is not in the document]
aritorto: The compiler takes "int" (when you pass an element index as an argument) as apossible "EntityType". To avoid that, I use different name methods.

opm/grid/LookUpData.hh
```diff
@@ -34,11 +34,18 @@
...
 
+#include <string>
 #include <type_traits>
 #include <vector>
 
+namespace Opm
+{
+class FieldPropsManager;
+}
+
```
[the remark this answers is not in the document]
aritorto: No, we don't. It was there before I realized we need the header since we are invoking methods of that class. I remove Entity forward declaration too, for the same reason. Thanks!

opm/grid/LookUpData.hh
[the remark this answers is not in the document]
aritorto: Dropped!

opm/grid/cpgrid/Entity.hpp
aritorto: Exception is thrown in lines 586-588:
 else {
        throw std::invalid_argument("The entity provided does not belong to the leaf grid view. ");
    }

opm/grid/cpgrid/Entity.hpp
aritorto: I did it before reading this comment. Same spirit of getLevelElem(). Thanks!

tests/cpgrid/lookupdataCpGrid_test.cpp
aritorto: It seems that EQLNUM is not supported, or I did something wrong before this other change.

aritorto:
jenkins build this opm-simulators=4942 serial please

opm/grid/LookUpData.hh
```diff
@@ -63,88 +59,95 @@ template <typename Grid, typename GridView>
...
     ///                         Default: GridType = Grid.
-    /// \param [in] element     EntityType object.
-    /// \return element origin index  Index of the origin cell (parent/equivalent cell when element has no father) in level 0.
     template<typename EntityType, typename GridType = Grid>
-    typename std::enable_if_t<std::is_same_v<GridType,Dune::CpGrid>,int> getOriginIndexFromEntity(const EntityType& elem) const;
+    typename std::enable_if_t<std::is_same_v<GridType,Dune::CpGrid>,int>
+    getFieldPropIdx(const EntityType& elem) const;
 
-    /// \brief: For general grids, it retunrs the same element index.
+    /// \brief: Retunrs the same element index for all grids different from CpGrid.
```
blattms, line 123: Typo: Retunrs should probably be Returns. (Maybe even from me? :flushed:)

opm/grid/LookUpData.hh
```diff
@@ -63,88 +59,95 @@ template <typename Grid, typename GridView>
...
     ///                         Default: GridType = Grid.
-    /// \param [in] element index
-    /// \return element index
     template<typename GridType>
-    typename std::enable_if_t<!std::is_same_v<GridType,Dune::CpGrid>,int> getOriginIndex(const int& elemIdx) const;
+    typename std::enable_if_t<!std::is_same_v<GridType,Dune::CpGrid>,int>
+    getFieldPropIdx(const int& elemIdx) const;
 
-    /// \brief: For CpGrid, it returns index of origin cell (parent/equivalent cell when elem has no father) in level 0.
+    /// \brief: Returns the index to search for the field properties, for CpGrids.
```
blattms, line 131: Should be "Returns", please.

opm/grid/LookUpData.hh
```diff
@@ -154,104 +157,95 @@ template<typename Grid, typename GridView>
...
-    /// \tparam     FeatureType Type of the property of the element, e.g. int, double, float, etc.
-    /// \param [in] element     EntityType object.
-    /// \param [in] feature_vec Vector with each entry, the feature of an element of
-    ///                         the gridView_ [for general grids], or level 0 for CpGrid.
-    /// \return feature of the given element.
-    template<typename EntityType, typename FeatureType>
-    FeatureType operator()(const EntityType& elem,const std::vector<FeatureType>& feature_vec) const;
-
-    /// \brief: Call operator taking an Index and a FeatureVector.
+    /// \brief: Get field propertry for an element in the leaf grid view, from a vector, via Cartesian Index.
```
blattms, line 178: "propertry" should be "property"

opm/grid/LookUpData.hh
```diff
@@ -154,104 +157,95 @@ template<typename Grid, typename GridView>
...
-    ///         Return feature of the entity, via CARTESIAN INDEX
-    ///         For general grids, the feature vector is given for the gridView_.
-    ///         For CpGrid, the feature vector is given for level 0.
+    ///         For general grids, the field property vector is assumed to be given for the gridView_.
+    ///         For CpGrid, the field property vector is assumed to be given for level 0 when isFieldPropLgr_ == false,
+    ///         and for certain LGR/level > 0 when isFieldPropLgr_ == true.
+    template<typename EntityType, typename FieldPropType>
+    FieldPropType operator()(const EntityType& elem,const std::vector<FieldPropType>& fieldProp) const;
+
+    /// \brief: Get field propertry for an element in the leaf grid view, from a vector, via Cartesian Index.
```
blattms, line 186: "property"

opm/grid/LookUpData.hh
```diff
@@ -154,104 +157,95 @@ template<typename Grid, typename GridView>
...
+                           const std::string& propString,
+                           const ElemOrIndex& elemOrIndex) const;
+
+    /// \brief: Get property of type int from field properties manager by name, via element or its index.
+    template<typename ElemOrIndex>
+    int fieldPropInt(const FieldPropsManager& fieldPropsManager,
+                     const std::string& propString,
+                     const ElemOrIndex& elemOrIndex) const;
+
+    /// \brief: Retunrs the same element index for all grids different from CpGrid.
```
blattms, line 206: "Returns"

opm/grid/LookUpData.hh
```diff
@@ -154,104 +157,95 @@ template<typename Grid, typename GridView>
...
     ///                         Default: GridType = Grid.
-    /// \param [in] element     EntityType object.
-    /// \return element origin index  Index of the origin cell (parent/equivalent cell when element has no father) in level 0.
     template<typename EntityType, typename GridType = Grid>
-    typename std::enable_if_t<std::is_same_v<GridType,Dune::CpGrid>,int> getOriginIndexFromEntity(const EntityType& elem) const;
+    typename std::enable_if_t<std::is_same_v<GridType,Dune::CpGrid>,int>
+    getFieldPropCartesianIdx(const EntityType& elem) const;
 
-    /// \brief: For general grids, it retunrs the same element index.
+    /// \brief: Retunrs the same element index for all grids different from CpGrid.
```
blattms, line 225: "Returns"

opm/grid/LookUpData.hh
blattms: the answer seems to be no,

opm/grid/LookUpData.hh
```diff
@@ -261,131 +255,186 @@ protected:
...
 typename std::enable_if_t<std::is_same_v<GridType,Dune::CpGrid>,int>
-Opm::LookUpData<Grid,GridView>::getOriginIndexFromEntity(const EntityType& elem) const
+Opm::LookUpData<Grid,GridView>::getFieldPropIdx(const EntityType& elem) const
 {
     static_assert(std::is_same_v<Grid,GridType>);
-    return elem.getOrigin().index();
+    static_assert(std::is_same_v<EntityType,Dune::cpgrid::Entity<0>>);
+    if (isFieldPropInLgr_ && elem.level()) { // level > 0 == true ; level == 0 == false
+        // In case some LGRs do not have refined field properties, the next line need to be modified.
+        return elem.getLevelElem().index(); // getLgrElem() returns equivalent refined Entity in the LGR the leafGridView cell was born.
```
blattms, line 316: Not a big one, but the method name used in comment is wrong now. Maybe just skip it?

opm/grid/cpgrid/Entity.hpp
blattms: Indeed there is an exception. 

To me it was not clear that `!(pgrid_ -> leaf_to_level_cells_.empty())` is the check that we are on the leaf grid view. Isn't this checking more than that? It seems like for a grid with no refinement (only one level) `pgrid_ -> leaf_to_level_cells_` would be empty too.

If that is intentional then we should probably adjust the comment and the message of the exception

opm/grid/cpgrid/Entity.hpp
blattms: With my current knowledge this comment is just dumb. Resolving this one.

blattms (commented):
Thanks a lot for the changes.

I made some further comments and posed some questions-

bska:
jenkins build this opm-simulators=4942 serial please

opm/grid/LookUpData.hh
aritorto: Skipped

aritorto:
In CpGridData.cpp:55,70   level_data_ptr_ = &data;  so that's not a problem. in the definition of getLevelCartesianIdx(). However, when you invoke getLevelElem() then, for cells born in LGRs, you need to access leaf_to_level_cells_, which is empty for grids without LGRs. I improved the comments on getLevelElem(), which throws for cells not belonging to the leaf grid view.

aritorto:
jenkins build this opm-simulators=4942 serial please

opm/grid/LookUpData.hh
aritorto: this one looks correct

aritorto:
jenkins build this opm-simulators=4942 serial please

PR #684, opened 2023-11-02
[bugfix] Deprecated getCollectiveCommunication() in cpgrid_aquifer_test

aritorto:
After DUNE version 2.7, the method getCollectiveCommunication() was changed to getCommunication(). The getCollectiveCommunication() issues the deprecation warning and assumes all process take part in the communication. However, this might not be the case, namely, only a subset might be used.

In CpGrid Aquifer test, getCollectiveCommunication() was deprecated. The fix consists in replacing it by getCommunication().


aritorto:
jenkins build this please

PR #683, opened 2023-10-26
Assign field property on Leaf Grid View

aritorto:
Given a double/int field property, the methods assignFielPropDoubleOnLeaf(...) and assignFieldPropIntOnLeaf(...) return the vector<double/int> with the field property for all cells in the leaf grid view.

PR #685, opened 2023-11-02
Delete default value in assignFieldPropDoubleOnLeaf

aritorto:
Deleting the unnecessary function argument "default value" from the method that assigns field properties of type double, on the Leaf Grid View.  Followed up of OPM/opm-grid#683.


aritorto:

blattms:
Thanks a lot. Highly appreciated.
And sorry for suggesting it in the first place. :flushed:

PR #687, opened 2023-11-07
cartesianToCompressed changed to unordered multimap

aritorto:
When a CpGrid has LGRs, a Cartesian index (key)  is associated to more than one compressed index (value). Therefore, this unordered_map now has been replaced by an unordered_multimap, to support multiple value with a same key.


aritorto:

opm/grid/utility/cartesianToCompressed.cpp
```diff
@@ -23,31 +24,30 @@
...
+        for (int i = 0; i < num_cells; ++i) {
+            retval.insert(std::pair<int,int>(global_cell[i], i));
+        }
+    } else {
+        for (int i = 0; i < num_cells; ++i) {
+            retval.insert(std::pair<int,int>(i, i));
         }
-        return retval;
     }
+    return retval;
```
bska, line 49: If you're changing this anyway, now would be a good time to switch to
```C++
retval.emplace(x, y)
```
instead of
```C++
retval.insert(std::pair<int,int>(x, y))
```
They do the same thing, effectively, but the former (`.emplace()`) is (slightly) faster since it constructs the value in-place.

tests/test_compressed_cartesian_mapping.cpp
```diff
@@ -26,25 +27,107 @@
...
     const std::vector<int> global_cell{0, 1, 2, 3, 5, 6, 8, 9};
+    const int num_cells = global_cell.size(); // 8
+
+    const std::unordered_multimap<int, int> cartesian_to_compressed =
+        Opm::cartesianToCompressed(num_cells, global_cell.data());
+
+    for (const auto pair : cartesian_to_compressed)
+    {
+        std::cout << pair.first << " " << pair.second << '\n';
+    }
```
bska, line 44: Is this code left over from development?  If so, I think we should remove it.

bska (commented):
Just a couple of minor points now, as an initial review.  I haven't thought through the full implications of using using `unordered_multimap<>` here.  Another option **might** be
`unordered_map<int,std::vector<int>>` instead.

tests/test_compressed_cartesian_mapping.cpp
aritorto: Opps! No, that was supposed to be removed. Thanks for noticing!

opm/grid/utility/cartesianToCompressed.cpp
aritorto: Thanks for your feedback! Indeed, unordered_map<int, std::vector<int>> is an alternative too. I'll check it out.

PR #688, opened 2023-11-07
Avoid NNC in LGRs

aritorto:
To prevent NNC (no neighbor connections) in coarse cells that will get refined. A test with cases containing NNCs is pendent.

### DIFF
--- a/opm/grid/cpgrid/CpGrid.cpp
+++ b/opm/grid/cpgrid/CpGrid.cpp
@@ -1429,6 +1429,8 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
                                    const std::vector<std::array<int,3>>& endIJK_vec,
                                    const std::vector<std::string>& lgr_name_vec)
 {
+    // Check startIJK_vec and endIJK_vec have same size, and "startIJK[patch][coordinate] < endIJK[patch][coordinate]"
+    (*data_[0]).validStartEndIJKs(startIJK_vec, endIJK_vec);
     if (!distributed_data_.empty()){
         if (comm().rank()==0){
             OPM_THROW(std::logic_error, "Adding LGRs to a distributed grid is not supported, yet.");
@@ -1437,6 +1439,7 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
             OPM_THROW_NOLOG(std::logic_error, "Adding LGRs to a distributed grid is not supported, yet.");
         }
     }
+    // Check LGRs are disjoint
     if (startIJK_vec.size() > 0 && !(*data_[0]).disjointPatches(startIJK_vec, endIJK_vec)){
         if (comm().rank()==0){
             OPM_THROW(std::logic_error, "LGRs are not disjoint.");
@@ -1444,6 +1447,23 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
         else{
             OPM_THROW_NOLOG(std::logic_error, "LGRs are not disjoint.");
         }
+    }
+    // Check grid is Cartesian
+    const std::array<int,3>& coarseGrid_dim =  (*data_[0]).logical_cartesian_size_;
+    long unsigned int coarseGridXYZ = coarseGrid_dim[0]*coarseGrid_dim[1]*coarseGrid_dim[2];
+    if ((*data_[0]).global_cell_.size() != coarseGridXYZ){
+        if (comm().rank()==0){
+            OPM_THROW(std::logic_error, "Grid is not Cartesian. This type of refinement is not supported yet.");
+        }
+        else {
+            // No cells on rank > 0
+            return;
+        }
+    }
+    // Check all the cells to be refined have no NNC (no neighbouring connections).
+    std::vector<int> all_patch_cells = (*data_[0]).getPatchesCells(startIJK_vec, endIJK_vec);
+    if ((*data_[0]).hasNNCs(all_patch_cells)){
+        OPM_THROW(std::logic_error, "NNC face on a cell containing LGR is not supported yet.");
     }
     // Total amount of patches:
     const int& num_patches = startIJK_vec.size();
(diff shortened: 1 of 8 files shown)


aritorto:
jenkins build this please

opm/grid/cpgrid/CpGridData.cpp
```diff
@@ -1859,6 +1859,49 @@ CpGridData::getPatchesCells(const std::vector<std::array<int,3>>& startIJK_vec,
...
+            {
+            case I_FACE:
+                // left / right
+            case J_FACE:
+                // front / back
+            case K_FACE:
+                // bottom / top
+            case NNC_FACE:
+                noNNC = false;
+                OPM_THROW(std::logic_error, "NNC face on a patch cell. Not supported yet!");
```
goncalvesmachadoc, line 1885: I think would be clearer to the user something like: "NNC face on a cell containing LGR is not supported yet."

aritorto: Indeed, 'patch' is not clear. Thanks!

aritorto:

aritorto:

aritorto:

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1429,6 +1429,19 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
...
+    const std::array<int,3>& coarseGrid_dim =  (*data_[0]).logical_cartesian_size_;
+    long unsigned int coarseGridXYZ = coarseGrid_dim[0]*coarseGrid_dim[1]*coarseGrid_dim[2];
+    if ((*data_[0]).global_cell_.size() != coarseGridXYZ){
+        if (comm().rank()==0){
+            OPM_THROW(std::logic_error, "Grid is not Cartesian. This type of refinement is not supported yet.");
+        }
+        else {
+            OPM_THROW_NOLOG(std::logic_error, "Grid is not Cartesian. This type of refinement is not supported yet.");
+        }
+    }
```
blattms, line 1442: I should have thought about this and mentioned it before:
We should move this check below the next one (lines 1445-1452). Then we are sure loadBalance was not called in a parallel run. 
The check should only be done on rank 0:
```
 if (comm().rank()==0){
    if ((*data_[0]).global_cell_.size() != coarseGridXYZ){
            OPM_THROW(std::logic_error, "Grid is not Cartesian. This type of refinement is not supported yet.");
        }
    }
```
As the other processes will have an empty grid. (`(*data_[0]).global_cell_.size()==0`).

opm/grid/cpgrid/CpGridData.cpp
```diff
@@ -1947,7 +1978,8 @@ CpGridData::refineSingleCell(const std::array<int,3>& cells_per_dim, const int&
     const cpgrid::Geometry<3,3>& parent_cell = (*(geometry_.geomVector(std::integral_constant<int,0>())))[EntityRep<0>(parent_idx, true)];
     // Get parent cell corners.
     const std::array<int,8>& parent_to_point = this->cell_to_point_[parent_idx];
-    if (parent_to_point.size() != 8){
+    const std::set<int> nonRepeated_parentCorners(parent_to_point.begin(), parent_to_point.end());
+    if (nonRepeated_parentCorners.size() != 8){
```
blattms, line 1982: Nice.

opm/grid/cpgrid/CpGridData.hpp
```diff
@@ -77,6 +77,7 @@
 namespace Opm
 {
 class EclipseState;
+class NNC;
```
blattms, line 80: Do we need this?

blattms (commented):
Thanks a lot. Just a few comments from a preliminary look.

opm/grid/cpgrid/CpGridData.hpp
aritorto: We needed it in CpGrid.hpp, and I added it also here assuming we also need it here (Opm::NNC is the type of one of the arguments of the function defined in the new test file, and not having it declared in CpGrid.hpp was generating errors.)

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1463,6 +1476,15 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
     std::vector<int> all_patch_corners = (*data_[0]).getPatchCorners(startIJK_vec[0], endIJK_vec[0]);
     std::vector<int> all_patch_faces = (*data_[0]).getPatchFaces(startIJK_vec[0], endIJK_vec[0]);
     std::vector<int> all_patch_cells = (*data_[0]).getPatchCells(startIJK_vec[0], endIJK_vec[0]);
+    // Check all the cells to be refined have no NNC (no neighbouring connections).
+    if ((*data_[0]).hasNNCs(all_patch_cells)){
+        if (comm().rank()==0){
+            OPM_THROW(std::logic_error, "NNC face on a cell containing LGR is not supported yet.");
+        }
+        else{
+            OPM_THROW(std::logic_error, "NNC face on a cell containing LGR is not supported yet.");
```
blattms, line 1485: Somehow my comment vanished or I am simply to dumb to comment. Next try:

we do the same here in if and else branch. Maybe just skip the if-else and just throw?

aritorto: I see. In  any case, there is a typo. The second throw should have been OPM_THROW_NOLOG (as the other if-else checks).

opm/grid/cpgrid/CpGridData.hpp
aritorto: As we discussed, this is related to (the many) forward and **friend** declarations of free functions defined in test files, due to the need of accessing certain private member(s) of the class. It will be solved/modified in a separate PR.

blattms: I see. Thanks for the explanation,

aritorto:

tests/cpgrid/avoidNNCinLGRsCpGrid_test.cpp
```diff
@@ -0,0 +1,169 @@
...
+    const auto& hasNNC =  (*data[0]).hasNNCs(all_lgrs_cells);
+    std::cout << std::boolalpha;
+    std::cout << hasNNC << '\n';
+
+    if(hasNNC){
+        BOOST_CHECK_THROW(grid.addLgrsUpdateLeafView(cells_per_dim_vec, startIJK_vec, endIJK_vec, lgr_name_vec), std::logic_error);
+    }
+    else{
+        grid.addLgrsUpdateLeafView(cells_per_dim_vec, startIJK_vec, endIJK_vec, lgr_name_vec);
+    }
```
blattms, line 98: I am a bit skeptical about this approach:

We are using hasNNcs(...) to decide what to test, but hasNNcs(...) is called inside addLgrsUpdateLeafView to decide whether to throw. Seems like we do not really check anything. 

I think we should rather for each test know when whether or not there is an NNC to one of the cells that will be refined and pass another argument (e.g. hasNNCs) to the testCase function.

Maybe we should also add a testCase that won't throw in addition if there is none, yet.

aritorto: Yes! The reason of the if-statement is due to having 2 cases where it should throw and, the last one, that has an NNC, but on cells not contain in any LGR. We can also remove that case, or re-design the test without the if-statement and adding an extra argument in the function, too.

blattms: What I meant is that this decision should not be based on calling the hasNNC methods, but there should be a parameter that tells the test whether this should throw or not.

Then for e.g. NNCoutsideLgrs this parameter would be false:
```
testCase(deckString, nnc,  cells_per_dim_vec, startIJK_vec, endIJK_vec, lgr_name_vec, false);
```
In the current version if hasNNCs is wrong (returns true if there are none) this will not be detected.

aritorto:

PR #692, opened 2023-11-15
Refactor intersection index inside and document what it does

blattms:
Instead of switch-case we compute the index directly for Non-NNcs.


blattms:
jenkins build this please

opm/grid/cpgrid/Intersection.cpp
```diff
@@ -127,28 +127,19 @@ void Intersection::increment()
...
-        // The Dune grid interface is essentially meaningless
-        // for non-neighbouring "intersections".
-        return -1;
-    default:
-        OPM_THROW(std::runtime_error,
-                  "Unhandled face tag: " + std::to_string(tag));
+    const auto& face = faces_of_cell_[subindex_];
+    const auto tag = pgrid_->face_tag_[face];
+    if (tag == NNC_FACE) {
+        return 2 * static_cast<int>(tag);
```
atgeirr, line 137: Should this not be -1?

blattms: thanks for catching this. Yes it should be. :see_no_evil:

blattms:
Added another commit that reused indexInInside in boundaryId to reduce code.

blattms:
jenkins build this please

blattms:
Checks a are green. Do you have more comments?

aritorto:
Apart from those comments on documentation ("idx of the underlying Cartesian grid" -> "idx of the unit cube"), no further comments from my side

blattms:
jenkins build this please

opm/grid/cpgrid/Intersection.cpp
```diff
@@ -127,28 +111,18 @@ void Intersection::increment()
...
+    const auto& face = faces_of_cell_[subindex_];
+    const auto tag = pgrid_->face_tag_[face];
+    if (tag == NNC_FACE) {
         return -1;
-    default:
-        OPM_THROW(std::runtime_error,
-                  "Unhandled face tag: " + std::to_string(tag));
+    }
+    else {
+        return 2 * static_cast<int>(tag) +
```
atgeirr, line 124: While this should work right now, it depends on the (not guaranteed to be stable) values of the face_tag enum, which actually has no defined values, but depends on the defaults. I therefore think the existing code is a bit safer.

opm/grid/cpgrid/Intersection.hpp
```diff
@@ -191,12 +191,28 @@ namespace Dune
...
 
-            /// Local index of codim 1 entity in the inside() entity
+            /// \brief Local index of codim 1 entity in the inside() entity
             /// where intersection is contained in.
+            ///
+            /// Note that CpGrid does not support codim-1 entities and return
+            /// the index of the face of the underlying cartesian grid assuming that
+            /// all faces are perpendicular to the axis:
+            /// For faces perpendicular to x-axis: 0 (normal points contrary to axis) 1 (else)
+            /// For faces perpendicular to x-axis: 2 (normal points contrary to axis) 3 (else)
```
atgeirr, line 201: This should be y-axis (and z-axis below), similar for the next block.

Also: the return values are determined by topological orientation, not by the direction the normal points in.

PR #693, opened 2023-11-17
fixed: explicitly check against double precision 0.0

akva2:
For some reason I didn't really care to investigate further, Boost 1.66 ends up in an integer path and shit meets f*n.


bska (approved):
Very good.  These could arguably use `BOOST_CHECK_CLOSE` instead, but that's a bigger semantic change.  I'll merge into master.

PR #694, opened 2023-11-20
Get compressed size for level 0

aritorto:
When reading relative permeability defined by the input file, (currently) done for a grid without LGRs, the compressed index used has to be the one for level 0 grid. Therefore, a method compressedLevelZeroSize() has been added in the CartesianIndexMapper interface, for CpGrid and Polyhedral specializations.

### DIFF
--- a/opm/grid/CpGrid.hpp
+++ b/opm/grid/CpGrid.hpp
@@ -66,7 +66,7 @@ class NNC;
 
 namespace Dune
 {
-
+    template<typename Grid> class CartesianIndexMapper;
     class CpGrid;
 
     namespace cpgrid
@@ -237,6 +237,7 @@ namespace Dune
     class CpGrid
         : public GridDefaultImplementation<3, 3, double, CpGridFamily>
     {
+        friend class Dune::CartesianIndexMapper<CpGrid>;
         friend class cpgrid::CpGridData;
         friend class cpgrid::Entity<0>;
         friend class cpgrid::Entity<1>;
--- a/opm/grid/CpGrid.hpp
+++ b/opm/grid/CpGrid.hpp
@@ -66,7 +66,6 @@ class NNC;
 
 namespace Dune
 {
-    template<typename Grid> class CartesianIndexMapper;
     class CpGrid;
 
     namespace cpgrid
@@ -237,7 +236,6 @@ namespace Dune
     class CpGrid
         : public GridDefaultImplementation<3, 3, double, CpGridFamily>
     {
-        friend class Dune::CartesianIndexMapper<CpGrid>;
         friend class cpgrid::CpGridData;
         friend class cpgrid::Entity<0>;
         friend class cpgrid::Entity<1>;
@@ -398,6 +396,9 @@ namespace Dune
         /// those dealing with permeability fields from the input deck
         /// from whence the current CpGrid was constructed.
         const std::vector<int>& globalCell() const;
+
+        /// @brief Returns either data_ or distributed_data_(if non empty).
+        const std::vector<std::shared_ptr<Dune::cpgrid::CpGridData>>& chooseData() const;
 
         /// @brief
         ///    Extract Cartesian index triplet (i,j,k) of an active cell.
--- a/opm/grid/cpgrid/CartesianIndexMapper.hpp
+++ b/opm/grid/cpgrid/CartesianIndexMapper.hpp
@@ -49,6 +49,11 @@ namespace Dune
             return grid_.globalCell().size();
         }
 
+        int compressedLevelZeroSize() const
+        {
+            return (*(grid_.data_[0])).global_cell_.size();
+        }
+
         int cartesianIndex( const int compressedElementIndex ) const
         {
             assert(  compressedElementIndex >= 0 && compressedElementIndex < compressedSize() );
--- a/opm/grid/cpgrid/CartesianIndexMapper.hpp
+++ b/opm/grid/cpgrid/CartesianIndexMapper.hpp
@@ -51,7 +51,8 @@ namespace Dune
 
         int compressedLevelZeroSize() const
         {
-            return (*(grid_.data_[0])).global_cell_.size();
+            
+            return (*grid_.chooseData()[0]).size(0);
         }
 
         int cartesianIndex( const int compressedElementIndex ) const
--- a/opm/grid/cpgrid/CpGrid.cpp
+++ b/opm/grid/cpgrid/CpGrid.cpp
@@ -601,16 +601,21 @@ const std::array<int, 3>& CpGrid::logicalCartesianSize() const
     return current_view_data_ -> logical_cartesian_size_;
 }
 
+const std::vector<std::shared_ptr<Dune::cpgrid::CpGridData>>& CpGrid::chooseData() const
+{
+    if (current_view_data_ == this-> data_.back().get()){
+        return data_;
+    }
+    else{
+        return distributed_data_;
+    }
+}
+
 const std::vector<int>& CpGrid::globalCell() const
 {
     // Temporary. For a grid with LGRs, we set the globalCell() of the as the one for level 0.
     //            Goal: CartesianIndexMapper well-defined for CpGrid LeafView with LGRs.
-    if (current_view_data_ == this-> data_.back().get()){
-        return current_view_data_ -> global_cell_;
-    }
-    else{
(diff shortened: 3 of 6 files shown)


aritorto:
jenkins build this opm-simulators=5008 serial please

aritorto:
jenkins build this opm-simulators=5008 serial please

opm/grid/cpgrid/CartesianIndexMapper.hpp
```diff
@@ -49,6 +49,11 @@ namespace Dune
             return grid_.globalCell().size();
         }
 
+        int compressedLevelZeroSize() const
+        {
+            return (*(grid_.data_[0])).global_cell_.size();
```
blattms, line 54: The problem here is that we need to use the correct member based on whether or not distributed_data_ is populated. Like above it works only if there is no entry in distributed_data_.
We need select either data_ or distributed_data_ depending on what current_view_data_ points to. Maybe we already have a function for that? If not we should probably implement one as there should be quite a few other occasion where we needed to the same thing (e.g. in .CpGrid::levelIndexSet)

Maybe using `size(0)` on CpGridData would be more expressive? Then we can skip one friend below.

aritorto: Oh I see, there are indeed a few occasions where we select between data_ and distributed_data_ in CpGrid implementation. I'll check if there is already a method doing that, so we replace those lines with it. 

size(0) gives total amount of cells on the CpGridData
global_cell_.size() gives total amount of ACTIVE cells on the CpGridData?

I was hesitating about this. Is this correct: compressed index == active index == value stored in global_cell_ [certain cell] ?

blattms:
I was wondering a bit why the test is green even though we have parallel tests. The explanation is that the RelPermDiagnostics are done before load balancing.

opm/grid/cpgrid/CartesianIndexMapper.hpp
blattms: we  (or at least I) are probably not really clear when we talk about this.
Only in the representation of the input file (EclipseGrid) inactive cells are represented. When constructing CpGrid from that information (after maybe even deactivating more), we simply skip the inactive cells. Information about inactive cells is available implicitly via global_cell_.

Hence the total amount of cells on CpGridData is the total amount of active cells. I guess compressed and active index are synonyms.

In global_cell_ the global (or cartesian index is stored). Hence for a 1x1x3 cartesian grid with the center cell inactive it would have 2 entries `{0 2}` and no entry would be equal to 1. (I had to use a 1D example because I always forget the ordering :sweat_smile:)

aritorto: Just saw that CpGrid::globalCell() internally distinguishes between data_ and distributed_data_ 
[https://github.com/OPM/opm-grid/blob/afe008a35ba4fb1b153826f651cb621552d05ca2/opm/grid/cpgrid/CpGrid.cpp#L604](url)

aritorto:
jenkins build this opm-simulators=5008 serial please

opm/grid/CpGrid.hpp
```diff
@@ -399,6 +399,8 @@ namespace Dune
         /// from whence the current CpGrid was constructed.
         const std::vector<int>& globalCell() const;
 
+        std::vector<std::shared_ptr<Dune::cpgrid::CpGridData>> chooseData() const;
```
blattms, line 402: Would be cool, if we would prevent a copying a vector here.

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -609,7 +619,7 @@ const std::vector<int>& CpGrid::globalCell() const
         return current_view_data_ -> global_cell_;
     }
     else{
-        return this -> distributed_data_[0] ->global_cell_;
+        return this -> distributed_data_[0] -> global_cell_;
```
blattms, line 622: Should we use the new function here, too?

opm/grid/CpGrid.hpp
blattms: And some doxygen documentation would be highly appreciated. It might save others some time later.

opm/grid/cpgrid/CpGrid.cpp
aritorto: The problem is with the if-statement part, current_view_data_ is not what chooseData() "returns".
Option1: chooseData() "returns" current_view_data_ when distributed_data_.empty(), and distrubuted_data_ otherwise.
Option2: (the current one) chooseData() "returns" data_ when current_view_data_ is equal to data_.back(), distributed_data_ otherwise. 
Option3: Others ideas...

Here, we could used the new method, but checking the size of data_, or similar. Not sure if it's an improvement.

aritorto:
jenkins build this opm-simulators=5008 serial please

opm/grid/cpgrid/CpGrid.cpp
blattms: Option 3:
chooseData stays the same and we do
```
return chooseData().back()-> global_cell_;
```
Shouldn't this work as we always want to use the leaf grid view here?

aritorto: Does this mean that distributed_data_[0] == distributed_data_.back(), when current_view_data_ != this-> data_.back().get()? If yes, then we can do your suggested modification

aritorto: Done!

aritorto:
jenkins build this opm-simulators=5008 serial please

opm/grid/CpGrid.hpp
```diff
@@ -66,7 +66,7 @@ class NNC;
 
 namespace Dune
 {
-
+    template<typename Grid> class CartesianIndexMapper;
```
blattms, line 69: I think with we don't need this and the friend declaration anymore because of `chooseData`.

opm/grid/CpGrid.hpp
```diff
@@ -237,6 +237,7 @@ namespace Dune
     class CpGrid
         : public GridDefaultImplementation<3, 3, double, CpGridFamily>
     {
+        friend class Dune::CartesianIndexMapper<CpGrid>;
```
blattms, line 240: Maybe not needed, anymore?

opm/grid/cpgrid/CpGridData.hpp
```diff
@@ -141,7 +142,7 @@ template<class T, int i> struct Mover;
 class CpGridData
 {
     template<class T, int i> friend struct mover::Mover;
-
+    friend Dune::CartesianIndexMapper<CpGrid>;
```
blattms, line 145: I think we can get rid of this friend declaration, too, if we use `chooseData()[level]->size(0)` to get the size.

opm/grid/cpgrid/CartesianIndexMapper.hpp
```diff
@@ -49,6 +49,12 @@ namespace Dune
             return grid_.globalCell().size();
         }
 
+        int compressedLevelZeroSize() const
+        {
+            
+            return (*grid_.chooseData()[0]).global_cell_.size();
```
blattms, line 55: As we want to get the number codim-0 elements on that level, could we use the more expressive
```
(*grid_.chooseData()[0]).size(0);
```
Then we would also not need to access private members and could delete the friend declaration in CpGridData

blattms (commented):
Thanks for addressing all the previous comments and sorry for adding a few more.

I hope with these we get rid off some friend declarations. That would be awesome.

opm/grid/cpgrid/CartesianIndexMapper.hpp
aritorto: I think global_cell_ is related to active cells too. Probably, under our assumptions for supporting LGRs, they coincide now, but might not be the case for the general setting. 

CpGridData::size( /*codim*/ 0) returns cell_to_face_.size() (which, if I get it correctly, is the total amount of cells, no matter if they are active or inactive). 

global_cell_.size() should be total amount of active cells. 

Under our assumptions, all cells on level zero, on each Lgr, and  on the leaf grid view, are active. So we can do your suggestion now.

aritorto:
All the comments are addressed

aritorto:
jenkins build this opm-simulators=5008 serial please

blattms:
Perfect. thanks a lot

PR #696, opened 2023-11-27
Add cartesianCoordinateLevel()

aritorto:
For a cell on the leaf grid view, cartesianIndexMapper.cartesianCoordinate( cell idx, ...) computes the ijk index of the parent cell or the equivalent cell on level zero, for CpGrid with LGRs. 

Now, for every level, included level 0, a new method to compute the ijk index, given the index cell on a certain level and its level, is included. 

Such a method is needed in RelpermDiagnostics.cpp, to create the string of a cell index, from level zero. See also OPM/opm-simulators#5029

### DIFF
--- a/opm/grid/cpgrid/CpGrid.cpp
+++ b/opm/grid/cpgrid/CpGrid.cpp
@@ -1847,6 +1847,8 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
     (*data_[num_patches +1]).index_set_ = std::make_unique<cpgrid::IndexSet>(data_[num_patches+1]->size(0), data_[num_patches+1]->size(3));
     // Leaf local_id_set_
     (*data_[num_patches +1]).local_id_set_ = std::make_shared<const cpgrid::IdSet>(*data_[num_patches+1]);
+    // Leaf logical_cartesian_size_
+    (*data_[num_patches +1]).logical_cartesian_size_ =  (*data_[0]).logical_cartesian_size_;
 }
 
 
--- a/tests/cpgrid/lookupdataCpGrid_test.cpp
+++ b/tests/cpgrid/lookupdataCpGrid_test.cpp
@@ -144,6 +144,16 @@ void lookup_check(const Dune::CpGrid& grid)
         const auto cartIdx = cartMapper.cartesianIndex(elem.index());
         BOOST_CHECK(cartIdx == lookUpCartesianData.getFieldPropCartesianIdx(elem));
         BOOST_CHECK(cartIdx == lookUpCartesianData.getFieldPropCartesianIdx(elem.index()));
+        // Extra checks related to Cartesian Coordinate
+        std::array<int,3> ijk;
+        cartMapper.cartesianCoordinate(elem.index(), ijk); // this ijk corresponds to the parent/equivalent cell in level 0.
+        std::array<int,3> ijkLevel0;
+        cartMapper.cartesianCoordinateLevel(elem.getOrigin().index(), ijkLevel0, 0);
+        BOOST_CHECK(ijk == ijkLevel0);
+        // Throw for level < 0 or level > maxLevel()
+        std::array<int,3> ijkThrow;
+        BOOST_CHECK_THROW(cartMapper.cartesianCoordinateLevel(elem.index(), ijkThrow, -3), std::invalid_argument);
+        BOOST_CHECK_THROW(cartMapper.cartesianCoordinateLevel(elem.index(), ijkThrow, grid.maxLevel() + 1), std::invalid_argument);
         // Checks related to LGR field properties
         if (elem.level())
         {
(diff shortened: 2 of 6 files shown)


aritorto:
jenkins build this serial opm-simulators=5011 please

aritorto:
jenkins build this serial opm-simulators=5029 please

aritorto:
jenkins build this serial opm-simulators=5029 please

tests/cpgrid/lookupdataCpGrid_test.cpp
```diff
@@ -175,14 +192,14 @@ BOOST_AUTO_TEST_CASE(one_lgr_grid)
 {
     // Create a grid
     Dune::CpGrid grid;
-    const std::array<double, 3> cell_sizes = {1.0, 1.0, 1.0};
-    const std::array<int, 3> grid_dim = {4,3,3};
+    const std::array<double, 3>& cell_sizes = {1.0, 1.0, 1.0};
+    const std::array<int, 3>& grid_dim = {4,3,3};
```
blattms, line 196: I am not insisting on changing this back, but what is the reason behind changing this?

I think the compiler might now create a temporary to hold the array and the life time of it is prolonged until the constant reference goes out of scope. The temporary object is used via the reference below.

Previously an object was created on the stack and used directly below

aritorto: Thanks for the explanation. I thought I was preventing copying. It doesn't take long to go back to the old version. I'll do it. Thanks again!

blattms: Please note that this is a bit different from capturing a return value by const reference as e.g. in
```
const auto& featureInLGR = lookUpDataLGR(elem, fakeLgrFeatures[elem.level()-1]);
```

Even here I would be unsure whether this is worth it or not. That depends a bit on the implementation of the function (does it result in return value optimization (RVO) or not?) and on the type returned (is a copy expensive or not?)

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1847,6 +1847,8 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
     (*data_[num_patches +1]).index_set_ = std::make_unique<cpgrid::IndexSet>(data_[num_patches+1]->size(0), data_[num_patches+1]->size(3));
     // Leaf local_id_set_
     (*data_[num_patches +1]).local_id_set_ = std::make_shared<const cpgrid::IdSet>(*data_[num_patches+1]);
+    // Leaf logical_cartesian_size_
+    (*data_[num_patches +1]).logical_cartesian_size_ =  (*data_[0]).logical_cartesian_size_;
```
blattms, line 1851: Is this a bug fix?

tests/test_lookupdata_polyhedral.cpp
```diff
@@ -126,9 +127,19 @@ void lookup_check(const Dune::PolyhedralGrid<3,3>& grid)
...
+        // Extra checks related to Cartesian Coordinate
+        std::array<int,3> ijk;
+        cartMapper.cartesianCoordinate(idx, ijk);
+        std::array<int,3> ijkLevel;
+        cartMapper.cartesianCoordinateLevel(idx, ijkLevel, 0);
+        BOOST_CHECK(ijk == ijkLevel);
+        // Throw for level > 0 (Local grid refinement not supported for Polyhedral Grid)
+        std::array<int,3> ijkThrow;
+        BOOST_CHECK_THROW(cartMapper.cartesianCoordinateLevel(idx, ijkThrow, 4), std::invalid_argument);
+        BOOST_CHECK_THROW(cartMapper.cartesianCoordinateLevel(idx, ijkThrow, -3), std::invalid_argument);
```
blattms, line 142: Thanks a lot for the checks.

blattms (commented):
Thanks a lot. This looks good.

I am just wondering whether we really should do all those changes to const references.
Maybe we should at least omit them for the places where variables where initialized from initializer list. This might of course be personal taste and lack of knowledge on my side.

opm/grid/cpgrid/CpGrid.cpp
aritorto: Yes, indeed, this logocal_cartesian_size_ was missing. (Should it have been done in a separate PR? :( )

aritorto:
Unnecesary/wrong references removed

opm/grid/cpgrid/CpGrid.cpp
blattms: I was just curious,

blattms:
> Unnecesary/wrong references removed

There was nothing wrong. Not even sure whether my suggestion was a good one. This topic is highly complicated and the performance depends on many things. In the end one would have to measure. But here in the tests that is not worth it, anyway.

blattms:
jenkins build this opm-simulators=5029 please

PR #697, opened 2023-11-28
[fix] Remove last_patch_cells

aritorto:
Unnecessary and wrong insert on the vector containing all the cell indices to be refined has been deleted.


aritorto:

akva2:
please consider if you really need serial in all your triggers, we are rather pressed on ci resources these days.

aritorto:
Oh, I wasn't aware. I'll keep it mind next time. It's my default jenkins instruction. Which one would be more suitable?

blattms (approved):
Thanks a lot for finding and fixing this.

PR #698, opened 2023-11-28
[BugFix] On special cells, old lgr face idx stored instead of leaf face idx

aritorto:
On leaf grid view cells touching the boundary of an LGR, the data stored on cell_to_face_ was incorrect: The old face idx on the lgr was stored, instead of the leaf grid view face index.


aritorto:
jenkins build this please

blattms (commented):
Thanks this fix looks good.

Please improve the title. It should probably tell what is fixed. The old state should be described in the description.

E.g. "[Bugfix] Use correct face indices for faces on boundary of LGR"

blattms:
thanks, much better. :+1:

PR #699, opened 2023-11-30
Static invalid index member added in CpGrid and Entity

aritorto:
Replace the use of "-1" for invalid indices (of type int, e.g. level, index cell, etc) by a static member of type int, with value std::numeric_limits<int>::max().

### DIFF
--- a/opm/grid/cpgrid/CpGrid.cpp
+++ b/opm/grid/cpgrid/CpGrid.cpp
@@ -1473,9 +1473,9 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
     // Map to relate boundary patch faces with their children refined/new-born ones. {0,oldFaceIdx} -> {level,{newFaceIdx0, ...}}
     std::map<std::array<int,2>, std::tuple<int, std::vector<int>>> old_to_new_boundaryPatchFaces;
     //
-    // For level0, attach children to each parent cell. For no parents, entry {-1, {}} representing {no level, {no children}}
+    // For level0, attach children to each parent cell. For no parents, entry {invalidIdx, {}} representing {no level, {no children}}
     auto& l0_parent_to_children_cells = (*data_[0]).parent_to_children_cells_;
-    l0_parent_to_children_cells.resize(data_[0]-> size(0), std::make_tuple(-1, std::vector<int>()));
+    l0_parent_to_children_cells.resize(data_[0]-> size(0), std::make_tuple(invalidIdx, std::vector<int>()));
     //
     // Get patches corner and face indices. We instantiate them with level1 info and then insert other levels info.
     std::vector<int> all_patch_corners = (*data_[0]).getPatchCorners(startIJK_vec[0], endIJK_vec[0]);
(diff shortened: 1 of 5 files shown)


aritorto:
jenkins build this please

atgeirr:
I think we have used -1 for this purpose for a long time, and would prefer to keep it negative. I was a little surprised the jenkins test was succesful!

Could we use std::numeric_limits::min instead?

aritorto:
Sure, I'll change it to min then. Thanks!

aritorto:
jenkins build this please

aritorto:
jenkins build this please

atgeirr:
Another minor (maybe) point: in the original design of the CpGrid there are no "invalid" entries in the topology tables, so what makes it needed now?

aritorto:
From my side, I can say that I'm trying to debug the new code that builds a CpGrid with Lgrs, from an LgrCollection, and, in this process, I'm finding out weak-points (not necessary "errors") on the code in CpGrid, precisely, CpGrid::addLgrsUpdateLeaf(). 
One of these weak-points is the use of -1 (or "empty value" which by default ends up being 0 when the type is integer) to "label" entities that do not verify a property, such as has parent cell, has children, etc.  

Another one, is the order we store faces. In principle, it seems not to matter when storing faces of each grid (each lgr, or the leaf view). But it might matter when we store cell_of_face_[ a cell on the leaf/ on an lgr]. "Our order" follows a {K_face false, J_face false, I_face false, I_face true, J_face true, K_face true}. Dune's order seems to be {I_face false, I_face true, J_false, J_face true, K_face false, K_face true}. So, there might be a coming-soon PR about this too.

aritorto:
@atgeirr  This new invalid index is used, mainly, to be able to use containers such as vectors (with size total amount of cells in each lgr, or leaf grid view), instead of (oredered/unordered) maps. 

For example, for level 0, the member data level_to_leaf_cells_[ cell index in level 0 ] = cell index on the leaf grid view, only makes sense to store values for the cells of level 0 that have not been refined. Since this is a vector of type int, for those entries that we do not populate, we get the default value 0.

Similar example, but for the leaf grid view is the data member child_to_parent_cells_. Not every cell on the leaf is a "child", so not every cell on the leaf has a parent cell on level zero.

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1473,9 +1473,9 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
     // Map to relate boundary patch faces with their children refined/new-born ones. {0,oldFaceIdx} -> {level,{newFaceIdx0, ...}}
     std::map<std::array<int,2>, std::tuple<int, std::vector<int>>> old_to_new_boundaryPatchFaces;
     //
-    // For level0, attach children to each parent cell. For no parents, entry {-1, {}} representing {no level, {no children}}
+    // For level0, attach children to each parent cell. For no parents, entry {invalidIdx, {}} representing {no level, {no children}}
```
blattms, line 1476: On second thought:
The first item is the level, right. Maybe it is better to keep using -1 here?

PR #700, opened 2023-12-05
Check cuboid shape of cells to be refined added

aritorto:
Local grid refinement is supported, currently, for cuboid cells. A check has been added and tested. A test with a non-cuboid shape is provided.

### DIFF
--- a/tests/cpgrid/cuboidShape_test.cpp
+++ b/tests/cpgrid/cuboidShape_test.cpp
@@ -1,0 +1,122 @@
+//===========================================================================
+//
+// File: cuboidShape_test.cpp
+//
+// Created: Tue 05.12.2023 11:00:00
+//
+// Author(s): Antonella Ritorto   <antonella.ritorto@opm-op.com>
+//
+// $Date$
+//
+// $Revision$
+//
+//===========================================================================
+
+/*
+  Copyright 2023 Equinor ASA.
+
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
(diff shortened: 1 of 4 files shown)


tests/cpgrid/cuboidShape_test.cpp
```diff
@@ -0,0 +1,122 @@
...
+        GRID
+        COORD
+        0 0 0  0 0 1
+        1 0 0  1 0 0
+        0 1 0  0 1 1
+        1 1 0  1 1 0
+        /
+        ZCORN
+        4*0
+        4*1
```
blattms, line 114: Looks like a perfect cube to me. Shouldn't one of the top or bottom coordinates be different from 0 or 1?

aritorto: line 108 and line 110 have the same corner twice. That meant to be gluing corner  '1' (from bottom face) with corner '5' (from top face), and  corner  '3' (from bottom face) with corner '7' (from top face).

Then, how should that part of the string be to reflex this shape with 6 different corners?

Thanks for finding the issue!

blattms: ZCORN is the depth of each point on the lines defined by COORDS.

aritorto: Yes, thanks. I checked the manual. Still need to figure how to pass the 8 corners (with 2 repeated ones) out, so that I get the non-cuboid shape gluing corner '1' with '5' and '3' with '7'

aritorto:
jenkins build this please

blattms:
Great. Thanks a lot.

PR #701, opened 2023-12-06
Throw on all MPI ranks for unsupported PINCH ALL combination.

blattms:
PINCH ALL is still not supported and one might get
```
opm-grid/opm/grid/MinpvProcessor.cpp:299] Support for ALL as option 4
of PINCH is only supported in OPM flow if one of the pinched out cells
has 0 MULTZ or PERMZ.
```

Unfortunately, this was thrown only on rank 0 as the others do not process files. Those were waiting in the next collective communication causing OPM flow to deadlock. With this change the failure is communicated and the program aborted.

### DIFF
--- a/opm/grid/cpgrid/processEclipseFormat.cpp
+++ b/opm/grid/cpgrid/processEclipseFormat.cpp
@@ -127,6 +127,16 @@ namespace cpgrid
     {
         std::vector<std::size_t> removed_cells;
         if (ccobj_.rank() != 0 ) {
+            if (ecl_state) {
+                // Handle potential exception during MINPV processing
+                // Needed because later there is collective communication that will
+                // otherwise deadlock
+                int success = 1;
+                ccobj_.broadcast(&success, 1, 0);
+                if (success == 0) {
+                    throw std::runtime_error("Error during MINPV processing");
+                }
+            }
             // Store global grid only on rank 0
             return removed_cells;
         }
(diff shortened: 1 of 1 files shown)


blattms:
jenkins build this please

opm/grid/cpgrid/processEclipseFormat.cpp
```diff
@@ -127,6 +127,16 @@ namespace cpgrid
     {
         std::vector<std::size_t> removed_cells;
         if (ccobj_.rank() != 0 ) {
+            if(ecl_state) {
```
atgeirr, line 130: Nitpicks: space after `if`, also after `try` on line 166, and on line 194 around `catch` and before the `{`.

opm/grid/cpgrid/processEclipseFormat.cpp
```diff
@@ -127,6 +127,16 @@ namespace cpgrid
...
+                // Handle potential exception during MINPV processing
+                // Needed because later there is collective communication that will
+                // otherwise deadlock
+                int success = 1;
+                ccobj_.broadcast(&success, 1, 0);
+                if (success == 0) {
+                    throw std::runtime_error("Exception on master process during MINPV processing");
```
atgeirr, line 137: The term "master process" is not precise and also not so useful for users. Suggest simply "Error encountered during MINPV processing".

atgeirr (approved):
Good catch, should be merged! Only minor nits.

blattms:
jenkins build this please

blattms:
Self merge because of prior approval. Comments were addressed.

PR #702, opened 2023-12-14
[Bugfix] Invalid corners on (leafGridView)face_to_point_

aritorto:
On the leaf grid view of a CpGrid with LGRs, there are faces equivalent to the ones from level zero which have not been refined. In particular, when populating face_to_point_ of coarse faces that happen to touch an LGR on its boundary on (at least) one corner, get the refined corner. 

There was a bug assigning  -1 in those cases, instead of the actual correct leaf grid view corner index. This PR fixes it.

### DIFF
--- a/opm/grid/cpgrid/CpGrid.cpp
+++ b/opm/grid/cpgrid/CpGrid.cpp
@@ -1689,17 +1689,12 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
             // Add the amount of points to the count num_points.
             num_points += old_face_to_point.size();
             for (int corn = 0; corn < 4; ++corn) {
-                // Auxiliary bool to identify boundary patch corners.
-                bool is_there_allPatchBoundCorn = false;
-                for(const auto& [l0_oldCorner, level_newCorner] : old_to_new_boundaryPatchCorners){
-                    is_there_allPatchBoundCorn = is_there_allPatchBoundCorn || (corn == l0_oldCorner[1]);
-                    //true-> boundary patch corner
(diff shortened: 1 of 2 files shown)


aritorto:
jenkins build this please

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1689,17 +1689,12 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
...
-                        aux_face_to_point[leafFaceIdx].push_back(level_to_leaf_corners[level_newCorner[0]][level_newCorner[1]]);
-                        break; // Go to the next corner (on the boundary of a patch)
-                    }
+                if (level_to_leaf_corners[0][old_face_to_point[corn]] == -1) {
+                    const auto [lgr, lgrCornIdx] = old_to_new_boundaryPatchCorners[{0, old_face_to_point[corn]}];
+                    aux_face_to_point[leafFaceIdx].push_back(level_to_leaf_corners[lgr][lgrCornIdx]);
                 }
-                if (!is_there_allPatchBoundCorn) {// Corner not involded in any LGR.
+                else{// Corner not involded in any LGR.
+                    assert(level_to_leaf_corners[0][old_face_to_point[corn]] != -1);
```
blattms, line 1697: We should remove this assert as it has to be fulfilled because this is the else branch of `if (level_to_leaf_corners[0][old_face_to_point[corn]] == -1)`

aritorto: Yes! It's a left over of debugging, it shouldn't be there.

aritorto:
jenkins build this please

PR #703, opened 2024-01-24
Print amount LGRs and leaf grid view cells

aritorto:
Print amount LGRs and leaf grid view cells, for CpGrid with LGRs


aritorto:
jenkins build this please

blattms:
I think having this information is very good. Maybe we should use `Opm::OpmLog::info` similar to [CPgrid.cpp:415](https://github.com/OPM/opm-grid/blob/df398761767c7072927c84ac11c61ee3e4b7837c/opm/grid/cpgrid/CpGrid.cpp#L415) then this will also show up in the log files.

@goncalvesmachadoc @gdfldm What kind of messages are commercial simulator printing about LGRs? Maybe we should mimic that.

blattms:
Shoot. Too trigger-happy. Did we run jenkins?

aritorto:
We did run it once. Too late now? We did not run it after changing scout for OpmLog::info : /

atgeirr:
The post-builder will hopefully say that all is ok. I can check out the latest master for myself and check as well.

blattms:
I did check it myself, shortly after merging. Compiles on my machine at least.

PR #704, opened 2024-01-26
[refactor] Remove numElems as function argument

aritorto:
Unnecessary function argument numElements removed. Associated OPM/opm-common#3876 and OPM/opm-simulators#5125


blattms:
:sunglasses: Looks nice. Thanks a lot for the simplification.

blattms (approved):

PR #706, opened 2024-01-29
[refactor] Pore volume taking into account LGRs

aritorto:
For the field property "PORV" (pore volume), when the grid is a CpGrid with LGRs, the pore volume of a leaf grid view cell is compute as the pore volume of its parent cells, multiplied by the leaf grid view cell volume, divided by the parent cell volume. In this way, the pore volume of the parent cell coincides with the sum of the pore volume of its children (cells).

This is "hidden" in the field property assigners defined in the class LookUpData. These methods are used in opm-simulators/ebos, when refactor is needed to support LGRs.

### DIFF
--- a/opm/grid/LookUpData.hh
+++ b/opm/grid/LookUpData.hh
@@ -310,9 +310,30 @@ std::vector<double> Opm::LookUpData<Grid,GridView>::assignFieldPropsDoubleOnLeaf
     unsigned int numElements = gridView_.size(0);
     fieldPropOnLeaf.resize(numElements);
     const auto& fieldProp = fieldPropsManager.get_double(propString);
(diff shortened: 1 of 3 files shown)


aritorto:
jenkins build this please

blattms:
Looks good.
The question is whether we should add a small test, e.g. checking that the sum of the pore volume of all cells of an LGR is near the one of the parent cell.

aritorto:
Thanks, yes, it might make sense to add a test. I'll convert the PR into a draft until I come up with one

aritorto:
jenkins build this please

aritorto:
jenkins build this please

blattms:
Thanks a lot. This is a great addition.

PR #707, opened 2024-01-31
[bugfix] Fix support for PINCH(4) ALL for zero pinched out permeability

blattms:
Fixes embarrasing bug that caused a throw whenever option 4 of PINCH was ALL. Now we support the corner case where one the cells in between has a zero permeability (or multz) set.

Full support for ALL is still lacking.


blattms:
jenkins build this please

akva2 (approved):
Yeah, we don't want to check the cell itself here.

PR #710, opened 2024-02-19
Fix integer (`int`) overflow in `preprocess.c` and `uniquepoints.c`

kjetilly:
Running large grids (134M cells, 512^3 logical Cartesian) causes segmentation faults from the function `finduniquepoints` and an error from `checkmemory`. The issue stems from `int` being used for variables that can overflow beyond `MAX_INT` or that are later being used in expressions that will overflow `MAX_INT`.

This PR makes two changes:

1) Change the type of `nc` in `finduniquepoints` to `size_t`

2) Change the type of `r, m, n` in `checkmemory` to `size_t`.

Note that the general usage of `int` in `preprecoss.c` and its siblings will have limitations on the number of cells, but these limitations are slightly larger than 134 M cells. Longer term solution would be to change to `size_t` or `long long`.

### DIFF
--- a/opm/grid/cpgpreprocess/preprocess.c
+++ b/opm/grid/cpgpreprocess/preprocess.c
@@ -172,7 +172,7 @@ checkmemory(int nz, struct processed_grid *out, int **intersections)
         n += MAX(n / 2, 12 * r);
     }
 
-    ok = m == out->m;
+    ok = m == ((size_t)(out->m));
     if (! ok) {
         void *p1, *p2, *p3, *p4;
 
@@ -191,7 +191,7 @@ checkmemory(int nz, struct processed_grid *out, int **intersections)
         if (ok) { out->m = m; }
     }
 
-    if (ok && (n != out->n)) {
+    if (ok && (n != ((size_t)(out->n)))) {
         void *p1;
 
         p1 = realloc(out->face_nodes, n * sizeof *out->face_nodes);
--- a/opm/grid/cpgpreprocess/preprocess.c
+++ b/opm/grid/cpgpreprocess/preprocess.c
@@ -154,7 +154,8 @@ compute_cell_index(const int dims[3], int i, int j,
 static int
 checkmemory(int nz, struct processed_grid *out, int **intersections)
 {
-    int r, m, n, ok;
+    size_t r, m, n;
+    int ok;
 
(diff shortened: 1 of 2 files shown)


kjetilly:
Jenkins build this please.

blattms:
You might be opening a can of worms here. I have been bothered by the use of 32bit ints in opm-flow when it comes to cells, etc. for quite some time. I just hoped that it would never become a problem.

Just be aware that even in opm-simulators we rely on cell indices being representable in a 32 bit integer. One place that I am aware of is EclTransmissibilty::trans_ where we store the face transmissibility by encoding the indices of the two neighboring cells in  a 64bit integer. This will be a problem if the number of active is not representable by a 32bit integer.

Would be cool if you would add a check whether the number of active cells is below the threshold.

atgeirr:
The problem size here in number of cells is far below MAX_INT, but we still have problems because i.e. vertices can be up to 8 times as many. So this just addresses a problem in the initial grid processing, and therefore does not really open a can of worms, it just makes it possible to push a bit further towards MAX_INT cells, possibly exposing further problems downstream. Allowing 2e9 cell runs would indeed require a much more comprehensive refactoring.

atgeirr (approved):

atgeirr:
jenkins build this please

blattms:
Seems like there are just two warnings to fixed and this can go in, then

bska:
jenkins build this please

opm/grid/cpgpreprocess/preprocess.c
```diff
@@ -190,7 +191,7 @@ checkmemory(int nz, struct processed_grid *out, int **intersections)
         if (ok) { out->m = m; }
     }
 
-    if (ok && (n != out->n)) {
+    if (ok && (n != ((size_t)(out->n))) {
```
bska, line 194: Looks like there's a right parenthesis (`)`) missing here.

bska:
jenkins build this please

opm/grid/cpgpreprocess/preprocess.c
kjetilly: I fixed it now, but had to do a double take, so the current build that you triggered might not build against the correct version.

bska (approved):

PR #711, opened 2024-02-19
Get coarse face from a refined face on the boundary of an LGR

aritorto:
To compute transmissibility, a reduction-argument is used based on certain symmetry between the 'usual' 6 faces of a cell. When the grid is a CpGrid with LGRs, (refined) faces with one neighboring coarse cell and one neighboring refined cell can be found. In these cases, the reduction-argument based on 'symmetry' may lead to wrong transmissibility values.
 
An alternative approach to compute transmissibilities of faces located on LGR boundaries, the 'parent coarse / original' face from level zero could be tracked and used. This PR  adds (and tests) the method getParentFace.

### DIFF
--- a/opm/grid/LookUpData.hh
+++ b/opm/grid/LookUpData.hh
@@ -35,10 +35,12 @@
 #define OPM_LOOKUPDATA_HH
 
 #include <dune/grid/common/mcmgmapper.hh>
+#include <dune/grid/common/intersection.hh>
 
 #include <opm/input/eclipse/EclipseState/Grid/FieldPropsManager.hpp>
 #include <opm/grid/cpgrid/Entity.hpp>
 
+#include <algorithm>
 #include <functional>
 #include <string>
 #include <type_traits>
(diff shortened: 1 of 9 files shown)


aritorto:
jenkins build this please

aritorto:
jenkins build this please

aritorto:
jenkins build this please

aritorto:
jenkins build this please

opm/grid/LookUpData.hh
```diff
@@ -431,6 +464,49 @@ Opm::LookUpData<Grid,GridView>::getFieldPropIdx(const int& elemIdx) const
...
+
+template<typename Grid, typename GridView>
+template<typename GridType>
+typename std::enable_if_t<!std::is_same_v<GridType,Dune::CpGrid>,bool>
+Opm::LookUpData<Grid,GridView>::isOnLgrBoundary(const int& faceIdx [[maybe_unused]]) const
+{
+    static_assert(std::is_same_v<Grid,GridType>);
+    // Check there are no LGRs. LGRs (level>0) only supported for CpGrid.
+    assert(gridView_.grid().maxLevel() == 0);
+    return gridView_.grid().maxLevel()>0;
```
blattms, line 496: If we merge this we need a working implementation. This will just run into assert for any refined grid other than CpGrid.
Wouldn't it be possible to use the intersection instead of an index? Then we can simply check the `!intersection.neighbor() || intersection.inside().level() != intersection.outside.level()`
Would work for any DUNE grid.

opm/grid/LookUpData.hh
```diff
@@ -431,6 +464,49 @@ Opm::LookUpData<Grid,GridView>::getFieldPropIdx(const int& elemIdx) const
...
 
+template<typename Grid, typename GridView>
+template<typename GridType>
+typename std::enable_if_t<!std::is_same_v<GridType,Dune::CpGrid>,int>
+Opm::LookUpData<Grid,GridView>::getParentFaceIdxFromLgrBoundaryFace(const int& faceIdx) const
+{
+    static_assert(std::is_same_v<Grid,GridType>);
+    // Check there are no LGRs. LGRs (level>0) only supported for CpGrid.
+    assert(gridView_.grid().maxLevel() == 0);
+    return faceIdx;
```
blattms, line 475: I fear there is no working implementation for non-CpGrids here. At least no easy or cheap one. That is  a pity.

I think just returning an index is really dangerous. At least we should throw an exception with a meaningful error message here.

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1854,6 +1854,66 @@ const std::map<std::string,int>& CpGrid::getLgrNameToLevel() const{
...
+    }
+    // - Discard the case when the face belongs to the interior of an LGR,
+    // or both cells are equivalent to their parent cells from level zero.
+    const auto& cell1 = Dune::cpgrid::Entity<0>(*current_view_data_,faceToCell[0].index(), true);
+    const auto& cell2 = Dune::cpgrid::Entity<0>(*current_view_data_,faceToCell[1].index(), true);
+    if ( cell1.level() == cell2.level() ) {
+        OPM_THROW(std::invalid_argument, "Face is not on the boundary of an LGR.");
+    }
+
+    // Identufy the coarse and the refined neighboring cell
```
blattms, line 1887: Indentify

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1854,6 +1854,66 @@ const std::map<std::string,int>& CpGrid::getLgrNameToLevel() const{
...
+        assert(cell1.level() > 0);
+        return cell1.level();
+    }
+    else {
+        const auto& cell2 = Dune::cpgrid::Entity<0>(*current_view_data_,faceToCell[1].index(), true);
+        return ( (cell1.level() == 0) && (cell2.level() > 0) ) || ( (cell1.level() > 0) && (cell2.level() == 0) );
+    }
+}
+
+const cpgrid::EntityRep<1> CpGrid::getParentFaceFromLgrBoundaryFace(int faceOnBoundaryLgrIdx) const
```
blattms, line 1871: I need to think about this code. It uses a lot of implementation details. Hopefully we can do something similar using the DUNE grid interface.

blattms (commented):
This is a rather premature review, sorry. I will need to think about this a bit.

opm/grid/LookUpData.hh
aritorto: That could work. I didn't notice. In principle, both levels could be different and not zero, but with the assumption of disjoint LGRs, then at least one of them should be zero

aritorto: On the other hand, we could just return false, and remove the unnecessary assert

opm/grid/LookUpData.hh
aritorto: I see, the idea behind was to have such a method, "getParentFAce...", defined for all type of grids. Then, for non CpGrid grids, would be better to throw instead of actually returning the same face index (or face)

aritorto:
jenkins build this opm-simulators=5218 please

aritorto:
jenkins build this opm-simulators=5218 please

aritorto:
Replaced/improved by OPM/opm-grid#716

PR #712, opened 2024-02-21
Get level idx added to lookup cell indices

aritorto:
For lookup indices, a new method has been added. Given a cell index from the leaf grid view:
- For general grids, returns the same index. 
- For CpGrid with LGRs, return the index on the level when the cell has a father, or the index in level zero otherwise. 

This method will be used in a coming PR in opm-simulators, to skip repeated computation of transmissibilities.


aritorto:
jenkins build this please

aritorto:
Unnecessary

PR #718, opened 2024-04-01
[fix] Fixes out-of-bounds access during refinement

blattms:
This fixes fallout from #715.


blattms:
jenkins build this please

akva2 (approved):
It is a bit hard to follow all details on the phone but assert is correct now and if it does not trigger this is certainly an improvement.

PR #719, opened 2024-04-01
avoid use of invalidated iterator

akva2:
see https://ci.opm-project.org/job/opm-grid-static-analysis/225/testReport/junit/(root)/debug_iterator/distribution_test_parallel/

Map is resized after newEnd has been calculated, thus the  iterator has been invalidated.


blattms:
That is certainly safer. 
I think in this case the old way should have worked as expected as this is a std::vector and reducing the size should not result reallocation.

PR #723, opened 2024-05-23
Improve refinement and partially support adaptivity for CpGrid

aritorto:
This PR 
- does not affect the update of the reference manual. 
- replaces OPM/opm-grid#721. 
- improves the refinement of CpGrid as well as partially support adaptivity, for general corner-point grids.
Marking cells with no neighboring connections for refinement is not supported yet.

Cells can be marked for refinement, without any assumption on the shape of the total cells to be refined. Previously, only a block of cells.

The adapted grid is the mix of coarse cells (not involved in refinement) and the refined entities.

### DIFF
--- a/opm/grid/CpGrid.hpp
+++ b/opm/grid/CpGrid.hpp
@@ -44,6 +44,7 @@
 
 #include <dune/grid/common/grid.hh>
 #include <opm/grid/cpgrid/CpGridDataTraits.hpp>
+#include <opm/grid/cpgrid/DefaultGeometryPolicy.hpp>
 #include <opm/grid/cpgrid/OrientedEntityTable.hpp>
 #include <opm/grid/cpgpreprocess/preprocess.h>
 #include <opm/grid/utility/platform_dependent/reenable_warnings.h> //  Not really needed it seems, but alas.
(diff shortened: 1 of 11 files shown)


opm/grid/cpgrid/CpGridData.cpp
```diff
@@ -2037,6 +2069,76 @@ bool CpGridData::patchesShareFace(const std::vector<std::array<int,3>>& startIJK
...
+int CpGridData::sharedFaceTag(const std::vector<std::array<int,3>>& startIJK_2Patches, const std::vector<std::array<int,3>>& endIJK_2Patches) const
+{
+    assert(startIJK_2Patches.size() == 2);
+    assert(endIJK_2Patches.size() == 2);
+
+    int faceTag = -1; // 0 represents I_FACE, 1 J_FACE, and 2 K_FACE. Use -1 for no sharing face case.
+     
+    if (patchesShareFace(startIJK_2Patches, endIJK_2Patches)) {
+        
+        const auto& detectSharing = [](std::vector<int> faceIdxs, std::vector<int> otherFaceIdxs){
```
blattms, line 2081: Please use const references for the function parameters to prevent copying.

opm/grid/cpgrid/CpGridData.cpp
```diff
@@ -2037,6 +2069,76 @@ bool CpGridData::patchesShareFace(const std::vector<std::array<int,3>>& startIJK
...
+        if (endIJK_2Patches[1][1] == startIJK_2Patches[1][1]) {
+            isShared = isShared || detectSharing(jFalse, jTrueOther);
+            if (isShared) {
+                faceTag = 1;
+            }  
+        }
+        if (startIJK_2Patches[1][2] == endIJK_2Patches[0][2]) {
+            isShared = isShared || detectSharing(kTrue, kFalseOther);
+            if (isShared) {
+                faceTag = 2;
```
blattms, line 2128: If there is no refinement, yet. then for CpGrid, the faces in k-direction should always coincide if the ijk indices match. Hence the inner check here might be superfluous.

opm/grid/cpgrid/CpGridData.cpp
```diff
@@ -2037,6 +2069,76 @@ bool CpGridData::patchesShareFace(const std::vector<std::array<int,3>>& startIJK
...
+            }  
+        }
+        if (endIJK_2Patches[1][2] == startIJK_2Patches[0][2]) {
+            isShared = isShared || detectSharing(kFalse, kTrueOther);
+            if (isShared) {
+                faceTag = 2;
+            }  
+        }
```
blattms, line 2136: Is this really needed? Shouldn't detectSharing be symmetric in the sense that `detectSharing(a, b) ==detectSharing(b, a)` holds?

This might also holds for every second if statement above

opm/grid/cpgrid/CpGridData.cpp
```diff
@@ -2177,15 +2279,10 @@ CpGridData::refineSingleCell(const std::array<int,3>& cells_per_dim, const int&
...
-    // Get ijk from parent cell (Cartesian index). Needed to compute width, length, and height of the parent cell.
-    std::array<int,3> ijk;
-    getIJK(parent_idx, ijk);
-    // Get dx,dy,dz for each cell of the patch to be refined
-    const auto& [widthX, lengthY, heightZ] = getWidthsLengthsHeights(ijk, {ijk[0]+1, ijk[1]+1, ijk[2]+1});
     // Refine parent cell
     parent_cell.refineCellifiedPatch(cells_per_dim, refined_geometries, refined_cell_to_point, refined_cell_to_face,
                                      refined_face_to_point, refined_face_to_cell, refined_face_tags, refined_face_normals,
-                                     {1,1,1}, widthX, lengthY, heightZ);
+                                     {1,1,1}, /*widthX, lengthY, heightZ*/ {1.}, {1.}, {1.});
```
blattms, line 2285: I just wonder whether the last three parameter could be computed inside `refineCellifiedPatch`, but that is not urgent,

opm/grid/cpgrid/CpGridData.cpp
```diff
@@ -2509,6 +2606,65 @@ CpGridData::refinePatch(const std::array<int,3>& cells_per_dim, const std::array
...
 }
 
+bool CpGridData::mark(int refCount, const cpgrid::Entity<0>& element)
+{
+    if (refCount == -1) {
+        OPM_THROW(std::logic_error, "Coarsening is not supported yet.");
+    }
+    // Check the cell to be marked for refinement has no NNC (no neighbouring connections). Throw otherwise.
+    if (hasNNCs({element.index()}) && (refCount == 1)) {
+        OPM_THROW(std::logic_error, "NNC face on a cell containing LGR is not supported yet.");
```
blattms, line 2616: Maybe use message "Refinement of cells with face representing an NNC is not supported, yet"?

opm/grid/cpgrid/CpGridData.cpp
```diff
@@ -2509,6 +2606,65 @@ CpGridData::refinePatch(const std::array<int,3>& cells_per_dim, const std::array
...
+}
+
+int CpGridData::getMark(const cpgrid::Entity<0>& element) const
+{
+    if (mark_.empty()) {
+        OPM_THROW(std::logic_error, "No element has been marked.");
+    }
+    else {
+        return mark_[element.index()]; // 1 refinement, 0 doing nothing, -1 coarsening (not supported yet)
+    }
```
blattms, line 2633: An alternative would be to return 0 if `mark_` is empty:
```
return mark_.empty() ? 0 : mark_[element.index()];
```

opm/grid/cpgrid/CpGridData.cpp
```diff
@@ -2509,6 +2606,65 @@ CpGridData::refinePatch(const std::array<int,3>& cells_per_dim, const std::array
...
+        bool refinementOrCoarsening = false;
+        for (int elemIdx = 0; elemIdx <  this-> size(0); ++elemIdx) {
+            const auto& element = Dune::cpgrid::Entity<0>(*this, elemIdx, true);
+            const auto& elemMark = getMark(element);  // 1 (to be refined), 0 (do nothing), -1 (to be coarsened - not supported yet)
+            refinementOrCoarsening = refinementOrCoarsening || (elemMark != 0);
+            if (refinementOrCoarsening) {
+                break;
+            }
+        }
+        return refinementOrCoarsening;
```
blattms, line 2654: Might this accomplish the same?
```suggestion
        for (int elemIdx = 0; elemIdx <  this-> size(0); ++elemIdx) {
            const auto& element = Dune::cpgrid::Entity<0>(*this, elemIdx, true);
            if (getMark(element) != 0)  // 1 (to be refined), 0 (do nothing), -1 (to be coarsened - not supported yet)
                return true;
        }
        return false;
```

opm/grid/cpgrid/CpGridData.cpp
```diff
@@ -2509,6 +2606,65 @@ CpGridData::refinePatch(const std::array<int,3>& cells_per_dim, const std::array
...
+}
+
+bool CpGridData::adapt()
+{
+    return preAdapt();
+}
+
+void CpGridData::postAdapt()
+{
+    mark_.resize(this->size(0), 0);
```
blattms, line 2665:
```suggestion
    mark_.clear();
    mark_shrink_to_fit();
```
First line is equivalent to yours but more expressive. Second one might actually free the memory (without it the underlying C-array will have the same size as before (`mark_.capacity() > mark.size()` will hold)

blattms (commented):
Nice work!
I will try to do this review iteratively. This is the review of CpGridData.cpp

opm/grid/cpgrid/Entity.hpp
```diff
@@ -550,9 +519,11 @@ Dune::cpgrid::Geometry<3,3> Dune::cpgrid::Entity<codim>::geometryInFather() cons
         return Dune::cpgrid::Geometry<3,3>(center_in_father_reference_elem, volume_in_father_reference_elem,
                                            in_father_reference_elem_corners, in_father_reference_elem_corner_indices_.data());
     }
+    else {
+        OPM_THROW(std::logic_error, "Entity has no father. Entry should be -1");
+    }
```
blattms, line 524: Mmh. Please explain what the second sentence is telling us.

blattms (commented):
Next batch of comments. Still two files to comment on.

opm/grid/CpGrid.hpp
```diff
@@ -555,58 +556,590 @@ namespace Dune
...
+                                           std::map<std::array<int,2>,int> elemLgrAndElemLgrCorner_to_adaptedCorner,
+                                           std::map<std::array<int,2>, std::array<int,2>> vanishedRefinedCorner_to_itsLastAppearance,
+                                           const std::vector<std::shared_ptr<Dune::cpgrid::CpGridData>>& markedElem_to_itsLgr,
+                                           const std::vector<int>& assignRefinedLevel,
+                                           std::map<std::array<int,2>,int> markedElemAndEquivRefinedCorn_to_corner,
+                                           const std::vector<std::vector<std::array<int,2>>>& cornerInMarkedElemWithEquivRefinedCorner,
+                                           const std::vector<std::array<int,3>>& cells_per_dim_vec,
+                                           int preAdaptMaxLevel);
+
+        /// @brief Get the ijk index of a refined corner, given its corner index of a single-cell-refinement.
```
blattms, line 1028: What ijk index is this? Is that the index of the refined cell with respect to the parent cell

opm/grid/CpGrid.hpp
```diff
@@ -555,58 +556,590 @@ namespace Dune
...
+        /// @brief Determine if a refined face is located in the interior of the single-cell-refinement.
+        ///
+        /// @param [in] cells_per_dim:    Total children cells in each direction (x-,-, and z-direction) of the single-cell-refinement.
+        /// @param [in] faceIdxInLgr:     Face index in the single-cell-refinement.
+        /// @param [in] elemLgr_ptr:      Pointer to the single-cell-refinement grid.
+        bool isRefinedFaceInInteriorLgr(const std::array<int,3>& cells_per_dim, int faceIdxInLgr,
+                                        const std::shared_ptr<cpgrid::CpGridData>& elemLgr_ptr) const;
+
+        /// @brief Determine if a refined corner is located on the boundary of the single-cell-refinement, and does not coincide with
+        ///        a preApradt-existing corner.
```
blattms, line 1057:
```suggestion
        ///        a preAdapt-existing corner.
```

opm/grid/CpGrid.hpp
```diff
@@ -555,58 +556,590 @@ namespace Dune
...
+        /// @param [in] cells_per_dim:    Total children cells in each direction (x-,-, and z-direction) of the single-cell-refinement.
+        /// @param [in] cornerIdxInLgr:   Corner index in the single-cell-refinement.
+        bool isRefinedNewBornCornerOnLgrBoundary(const std::array<int,3>& cells_per_dim, int cornerIdxInLgr) const;
+
+        /// @brief Determine if a new refined corner is located on an edge of the parent cell. In particular, it's on the boundary of
+        ///        the single-cell-refinement, and does not coincide with  a preApradt-existing corner.
+        ///
+        /// @param [in] cells_per_dim:    Total children cells in each direction (x-,-, and z-direction) of the single-cell-refinement.
+        /// @param [in] cornerIdxInLgr:   Corner index in the single-cell-refinement.
+        bool newRefinedCornerLaysOnEdge(const std::array<int,3>& cells_per_dim, int cornerIdxInLgr) const;
```
blattms, line 1068:
```suggestion
        bool newRefinedCornerLiesOnEdge(const std::array<int,3>& cells_per_dim, int cornerIdxInLgr) const;
```

opm/grid/CpGrid.hpp
```diff
@@ -555,58 +556,590 @@ namespace Dune
...
+
+        /// @brief Determine if a refined corner is located on the boundary of the single-cell-refinement, and does not coincide with
+        ///        a preApradt-existing corner.
+        ///
+        /// @param [in] cells_per_dim:    Total children cells in each direction (x-,-, and z-direction) of the single-cell-refinement.
+        /// @param [in] cornerIdxInLgr:   Corner index in the single-cell-refinement.
+        bool isRefinedNewBornCornerOnLgrBoundary(const std::array<int,3>& cells_per_dim, int cornerIdxInLgr) const;
+
+        /// @brief Determine if a new refined corner is located on an edge of the parent cell. In particular, it's on the boundary of
+        ///        the single-cell-refinement, and does not coincide with  a preApradt-existing corner.
```
blattms, line 1064: Thers seems to a typo in the second last word.

opm/grid/CpGrid.hpp
```diff
@@ -555,58 +556,590 @@ namespace Dune
...
+        ///
+        /// @param [in] cells_per_dim:    Total children cells in each direction (x-,-, and z-direction) of the single-cell-refinement.
```
blattms, line 1066:
```suggestion
        /// @param [in] cells_per_dim:    Total children cells in each direction (x-, y-, and z-direction) of the single-cell-refinement.
```

blattms (commented):
One file to go.

opm/grid/cpgrid/CpGridData.cpp
aritorto: The endIJK of a patch is the last corner of a cell with ijk index {endIijk[0]-1, endijk[1]-1, endijk[2]-1}. Therefore, when the startIJK_2Patches[1][2] == endIJK_2Patches[0][2], this means that there is a plane z=const where the top of  the second patch lays on" and  "the bottom of the first patch" are contained. But in principle, these top-patch2 and bottom-patch1 might not intersect at all, or not...

aritorto: This check is used in addLgrsAndUpdateLeafGrid which assumes the starting grid to be the GLOBAL grid (level zero), and to be Cartesian.  
This check is not used in adapt() refinement (which does not assume the starting grid to be Cartesian).

opm/grid/cpgrid/CpGridData.cpp
aritorto: As you mentioned, detectSharing(a,b) == detectSharing(b,a). Here we have 12 (potentially) different sets of face indices 
- 6 for patch_1 (all the faces laying on the left, right, front, back, bottom, and top boundary of the patch_1) 
and 
- other 6 for patch_2 (all the faces laying on the left, right, front, back, bottom, and top boundary of patch_2).
We have 6 if-statements since we check if patch1 is on the left, right, front, back, bottom, top of patch2  (and, for symmetry, we do not check the same 6 cases for patch2).  

Am I still repeating cases and I do not see it?  In any case, I'll add this description as a comment in the body function. Let me know if I still messed it up with extra cases/checks!

(I found a typo there too, so thanks for pointing at this piece of code!)

opm/grid/cpgrid/CpGridData.cpp
aritorto: Oh, that reminds me that I wanted to change the name of that method since now we simple refine one cell. Maybe we postpone that since we might need in the future this more general version with width, length, and height as additional function arguments.

opm/grid/cpgrid/CpGridData.cpp
```diff
@@ -1687,6 +1687,38 @@ void CpGridData::distributeGlobalGrid(CpGrid& grid,
 #endif
 }
 
+std::array<Dune::FieldVector<double,3>,8> CpGridData::getReferenceRefinedCorners(int idxInParentCell, const std::array<int,3>& cells_per_dim) const
+{
+    // Refined cells in parent cell: k*cells_per_dim[0]*cells_per_dim[1] + j*cells_per_dim[0] + i
+    std::array<int,3> ijk = {0,0,0};
+    ijk[0] = idxInParentCell % cells_per_dim[0];
+    idxInParentCell -= ijk[0]; // k*cells_per_dim[0]*cells_per_dim[1] + j*cells_per_dim[0]
```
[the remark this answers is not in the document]
aritorto: Thanks! I skip this one for now, might come back once I'm done with the other comments

opm/grid/cpgrid/CpGridData.cpp
aritorto: Nice simplification!

opm/grid/cpgrid/CpGridData.cpp
aritorto: Indeed, I tried to commit this suggestion and an error occurred so I'm incorporating this "by hand"

blattms (commented):
Here is the next portion of my review.
Last file (CpGrid.cpp) is the largest to review. This is the first part. I am at defineChildToParentAndIdxInParentCell

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1471,6 +1471,490 @@ template cpgrid::Entity<0> createEntity(const CpGrid&, int, bool);
...
+}
+
+bool CpGrid::adapt()
+{
+    const std::vector<std::array<int,3>>& cells_per_dim_vec = {{3,3,3}}; // Arbitrary chosen values.
+    std::vector<int> assignRefinedLevel(current_view_data_-> size(0));
+    const auto& preAdaptMaxLevel = this ->maxLevel();
+    for (int elemIdx = 0; elemIdx < static_cast<int>(current_view_data_->size(0)); ++elemIdx) {
+        const auto& element = cpgrid::Entity<0>(*current_view_data_, elemIdx, true);
+        assignRefinedLevel[elemIdx] = (this->getMark(element) == 1) ? (preAdaptMaxLevel +1) : 0;
```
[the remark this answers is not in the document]
blattms: Clearer now. It is the level where the kids will be.

opm/grid/cpgrid/CpGridData.cpp
aritorto: I see your point. However, I might have misinterpreted the preAdapt, adapt, postAdapt. I try to illustrate y interpretation with an example. 
- Start with a grid_0 that has 10 cells. 
- Mark 1 element for refinement into 2x2x2 refined cells.  This marking process happens in grid_0 so the vector mark_ (for grid_0) has size 10.
- Call preAdapt which returns true since 1 element has been marked for refinement.
- Call adapt. This creates refined_level_1_grid with the 2x2x2 = 8 new refined cells, and the leaf_grid with the 9 coarse cells equivalent to the non-marked elements from grid_0, together with the new 8 refined cells. So the leaf_grid has 17 cells in total.
- Call postAdapt which, in my interpretation, means leaf_grid.postAdapt() where leaf_grid.mark_ is created with size 17 cells and all marks equal to 0, that might be modified in a future refine-process. 

From your comment, I get the feeling that my interpretation is wrong, and we should always mark element on the same "starting grid", in this example, grid_0. Then, mark_.clear() would clean the marks but the capacity will be 10 (total amount of cells in grid_0) and not 17 (total amount of cells on leaf_grid) This means in particular that marking cells on the leaf_grid with a next adapt() call won't work when you mark elements with indices greater than 9. 

Is my interpretation wrong? Shall we always mark element "from a fix starting grid"?

opm/grid/cpgrid/Entity.hpp
```diff
@@ -439,14 +433,35 @@ int Entity<codim>::level() const
...
+bool Entity<codim>::isNew() const
+{
+    // WIP
+    // For an Entity "to be new" means that
+    // - it has a father
+    // - it has been created in the last call of adapt()
+    // To determine that, we track the level of the Entity and
+    // check if it was marked for refinement in the previos state
+    // of current_view_data_
+    return (isLeaf() && hasFather());
```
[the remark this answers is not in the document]
aritorto: Actually the WIP is accidentally still there. From your comment, I'm assuming I should keep it. Maybe this is also related to my potentially wrong interpretation of preAdapt, adapt, postAdapt...

opm/grid/cpgrid/Entity.hpp
```diff
@@ -439,14 +433,35 @@ int Entity<codim>::level() const
...
+}
+
+
+template<int codim>
+bool Entity<codim>::mightVanish() const
+{
+    const auto refinementMark = pgrid_ -> getMark(*this);
+    return (refinementMark == 1);
```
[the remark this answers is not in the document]
aritorto: This is probably again related to my interpretation of how the marking-elements process goes. I assume that when your stating grid is certain_grid with N cells, you marked some element on that grid, let's sat n1< ... < nr < N. Those marked element get mark equal to 1 when they will vanished and replaced by their refined children cells. Is this correct? When the mark is 0 or -1, it means "do nothing" or "coarsening (not supported yet, and in this case, this cell and "all its siblings cells" will vanish)".

opm/grid/cpgrid/Entity.hpp
aritorto: Unfortunately, it tell us implementation details that no one should care about. I'll remove that, thanks!

opm/grid/cpgrid/CpGridData.cpp
blattms: I am sorry. I misread your code. I thought you would resize mark_ to have size zero, here. (See also my other coment above concerning mark_).

What I am wondering is whether we should have an additional array on CpGridData that always has a positive size. (mark_.size() > 0). we have had occasions where we needed to save as much memory as possible.

Maybe we could use an alterantive approach in the long run:
- Initially `mark_` has size zero
- If you call `mark()` and `mark_` has size zero, then we resize it before marking.
- In postAdapt we then do as suggested above (clear and shrink_to_fit) to reduce the memory consumption again.

I am not 100% sure what you mean by starting_grid. But you actually might have read my mind here. I am really wondering where we should store the marks. If we store them on the leaf, then we have always a very big array. An alternative is to store it only on  the level (grid). Those might have less elements than the leaf grid and we save space for recursive refinement. But let's postpone that for now.

aritorto: I see your point. So far, we do the first two steps that you describe inside mark( int, entity) definition:
 if (mark_.empty()) {
        mark_.resize(this->size(0));
    }
    mark_[element.index()] = refCount;
    return (mark_[element.index()] == refCount);

With starting_grid I mean the grid where you select and mark elements to be refine. For example, level 0 grid (GLOBAL grid). 

The code also supports (up to further debugging) that:
- You start with a level 0 grid (unique existing grid so far)
-  You mark elements for refinement
- Calling preAdapt, adapt, and postAdapt create your refined-level-1-grid and the leaf-grid-view (mixed grid with coarse and refined cells).
- If you want, now you can mark element **on the leaf-grid-view** to be refined in a second call of adapt(). You select some leaf-elements and mark them for refinement.
- Call a second time preAdapt, adapt, postAdapt. This
     - creates refined-level-2-grid 
     -  deletes "the old leaf-grid-view" where you marked (in the second marking process) elements
     - creates a new and unique leaf-grid_view taking into account level 0 (starting_grid) elements not involved in any refinement, refined cells from refined-level-1grid, and refined-level-2-grid. 

In the end, you have:
- level 0 grid (what I would call "starting_grid_1" before invoking adapt() for first time)
- level 1 refined grid (created on the first call of adapt())
- level 2 refined grid (created on the second call of adapt() )
- leaf_grid_view_2 (created on the second call of adapt(), with "starting_grid_2" the "old leaf_grid_view" created from the frist call of adapt() that now does not exist anymore, where you selected and marked elements a second time).

I hope this helps!

tests/cpgrid/grid_lgr_test.cpp
```diff
@@ -151,42 +153,60 @@ void refinePatch_and_check(Dune::CpGrid& coarse_grid,
                     BOOST_CHECK( entity.isLeaf() == true);
                     // If it == endIt, then entity.isLeaf() true (when dristibuted_data_ is empty)
                     BOOST_CHECK( it == endIt);
+                    // Mark with 0 (not involved in refinement)
+                    coarse_grid.mark(0, entity);
```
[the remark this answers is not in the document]
aritorto: Opps! That shouldn't be there. It was my very first attempt to invoke mark(). I'll delete it! Nice catch!

tests/cpgrid/grid_lgr_test.cpp
```diff
@@ -151,42 +153,60 @@ void refinePatch_and_check(Dune::CpGrid& coarse_grid,
...
-                        BOOST_CHECK(it ->hasFather() == true);
-                        BOOST_CHECK(it ->level() == lgr);
-                        referenceElemOneParent_volume += it-> geometryInFather().volume();
-                        for (int c = 0; c < 3; ++c)
-                        {
-                            referenceElem_entity_center[c] += (it-> geometryInFather().center())[c];
-                        }
-                        // std::cout << it->index() << '\n';
-                    }
+                       BOOST_CHECK_EQUAL( it == endIt, false);
```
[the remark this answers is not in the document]
aritorto: Indeed!

tests/cpgrid/grid_lgr_test.cpp
```diff
@@ -592,9 +612,7 @@ BOOST_AUTO_TEST_CASE(pathces_share_face)
     const std::vector<std::array<int,3>> startIJK_vec = {{0,0,0}, {2,0,0}, {3,2,2}};
     const std::vector<std::array<int,3>> endIJK_vec = {{2,1,1}, {3,1,1}, {4,3,3}};
     const std::vector<std::string> lgr_name_vec = {"LGR1", "LGR2", "LGR3"};
-    BOOST_CHECK_THROW(coarse_grid.addLgrsUpdateLeafView(cells_per_dim_vec, startIJK_vec, endIJK_vec, lgr_name_vec), std::logic_error);
-    BOOST_CHECK_EQUAL(coarse_grid.chooseData()[0]->patchesShareFace(startIJK_vec, endIJK_vec), true);
```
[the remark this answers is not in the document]
aritorto: They are gone since now we do support blocks of cells to be refined sharing faces on their boundaries :)
Screenshot from 2024-06-07 16-47-12

tests/cpgrid/grid_lgr_test.cpp
```diff
@@ -604,15 +622,54 @@ BOOST_AUTO_TEST_CASE(pathces_share_faceB)
...
-    std::cout << "Patches shared at least one face." << "\n";
+    const std::vector<std::array<int,3>> cells_per_dim_vec = {{4,2,2}, {3,2,2}}; 
+    const std::vector<std::array<int,3>> startIJK_vec = {{0,0,0}, {1,0,0}}; 
+    const std::vector<std::array<int,3>> endIJK_vec = {{1,1,1}, {2,1,1}}; 
+    const std::vector<std::string> lgr_name_vec = {"LGR1", "LGR2"}; 
+    refinePatch_and_check(coarse_grid, cells_per_dim_vec, startIJK_vec, endIJK_vec, lgr_name_vec);
+}
+
+BOOST_AUTO_TEST_CASE(pathces_share_Iface_with_diff_cells_per_dim)
```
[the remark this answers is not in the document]
aritorto: Thanks!

tests/cpgrid/adapt_cpgrid_test.cpp
```diff
@@ -0,0 +1,563 @@
...
+                    bool closedCenter =  (std::abs(element.geometry().center()[0] - equiv_element_iter->geometry().center()[0]) < 1e-12) &&
+                        (std::abs(element.geometry().center()[1] - equiv_element_iter->geometry().center()[1]) < 1e-12) &&
+                        (std::abs(element.geometry().center()[2] - equiv_element_iter->geometry().center()[2])< 1e-12);
+
+                    while (!closedCenter) {
+                        ++equiv_element_iter;
+                        closedCenter = (std::abs(element.geometry().center()[0] - equiv_element_iter->geometry().center()[0]) < 1e-12) &&
+                            (std::abs(element.geometry().center()[1] - equiv_element_iter->geometry().center()[1]) < 1e-12) &&
+                            (std::abs(element.geometry().center()[2] - equiv_element_iter->geometry().center()[2])< 1e-12);
+                    }
```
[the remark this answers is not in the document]
aritorto: I'm adding the check that the iterator is not the end one. With the lambda-expression suggested, I'm having some issues. I postpone that change for now. Thanks!

tests/cpgrid/adapt_cpgrid_test.cpp
```diff
@@ -0,0 +1,562 @@
...
+                }
+                BOOST_CHECK( std::get<0>(data[startingGridIdx]-> parent_to_children_cells_[level_cellIdx[1]]) == -1);
+                BOOST_CHECK( std::get<1>(data[startingGridIdx]->parent_to_children_cells_[level_cellIdx[1]]).empty());
+                // Get index of the cell in level 0
+                const auto& entityOldIdx =   adapted_leaf.leaf_to_level_cells_[element.index()][1];
+                BOOST_CHECK( element.getOrigin().index() == entityOldIdx);
+                BOOST_CHECK( element.getOrigin().level() == 0);
+                BOOST_CHECK( element.isNew() == false);
+            }
+            BOOST_CHECK( element.mightVanish() == false); // marks get rewrtitten and set to 0 via postAdapt call
```
aritorto, line 291: @blattms: mightVanish() is tested here, for example

tests/cpgrid/adapt_cpgrid_test.cpp
```diff
@@ -0,0 +1,562 @@
...
+            auto it = element.hbegin(coarse_grid.maxLevel()); // With element.level(), fails
+            auto endIt = element.hend(coarse_grid.maxLevel());
+            const auto& [lgr, childrenList] = (*data[startingGridIdx]).parent_to_children_cells_[element.index()];
+            if (std::find(markedCells.begin(), markedCells.end(), element.index()) == markedCells.end()){
+                BOOST_CHECK_EQUAL(lgr, -1);
+                BOOST_CHECK(childrenList.empty());
+                BOOST_CHECK( element.isLeaf() == true);
+                // If it == endIt, then entity.isLeaf() true (when dristibuted_data_ is empty)
+                BOOST_CHECK( it == endIt);
+                BOOST_CHECK( element.mightVanish() == false);
```
aritorto, line 313: @blattms mightVanish() tested here too

tests/cpgrid/adapt_cpgrid_test.cpp
```diff
@@ -0,0 +1,562 @@
...
+
+    std::vector<int> assignRefinedLevel(coarse_grid.chooseData()[startingGridIdx]->size(0));
+
+    for (const auto& elemIdx : markedCells)
+    {
+        const auto& elem =  Dune::cpgrid::Entity<0>(*(coarse_grid.chooseData()[startingGridIdx]), elemIdx, true);
+        coarse_grid.mark(1, elem);
+        assignRefinedLevel[elemIdx] = coarse_grid.maxLevel() + 1;
+        BOOST_CHECK( coarse_grid.getMark(elem) == 1);
+        BOOST_CHECK( elem.mightVanish() == true);
```
aritorto, line 98: mightVanish() check

opm/grid/cpgrid/Entity.hpp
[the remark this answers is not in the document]
aritorto: I added a few comments in some of the places where mightVanish() get tested, in adapted_cpgrid_test.cpp

opm/grid/CpGrid.hpp
aritorto: If we refine a single cell with index "cell_idx (or "elemLgr")" on the "starting grid" (preAdapt grid, grid where we select and mark elements fore refinement), with cells_per_dim = { nx, ny, nz}, then this single-cell-auxiliary-refinement  has(with auxiliary I mean that in the end, the grid does not store this info, it's only an auxiliary computation):
- (nx +1)*(ny +1)*(nz +1) (auxiliary) refined corners,
- ((nx+1)*ny*nz) + (nx*(ny+1)*nz) + (nx*ny*(nz+1)) (auxiliary) refined faces, and
- nx*ny*nz (auxiliary) refined cells. 

The function argument cornerIdxInLgr is then an integer between 0 and  (nx +1)*(ny +1)*(nz +1) .
The return ijk value is computed taking into account how refined corners are stored in Geometry::refineCellifiedPatch [the name of this method does not fit so well now that we are actually refining single cells]. The order of the created corners in the refine method in Geometry.hpp is the followed one:
(j*(nx+1)*(nz+1)) + (i*(nz+1)) + k.

I hope this helps! I try to improve the description in the documentation.

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1471,6 +1471,490 @@ template cpgrid::Entity<0> createEntity(const CpGrid&, int, bool);
...
+    }
+    if(data_.size()>1) {
+        // Get the level where the element was born and its index in that level.
+        const auto& elemLevel =  element.level();
+        const auto& elemInLevel = (elemLevel == 0) ? element.getOrigin() : element.getLevelElem();
+        // Mark element in its level
+        data_[elemLevel] -> mark(refCount, elemInLevel);
+    }
+    // Mark element (also) in current_view_data_
+    return current_view_data_-> mark(refCount, element);
```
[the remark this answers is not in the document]
aritorto: I skip this comment until we meet and define how to design the adapt process ( in which grid we are allowed to mark elements, etc)

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1471,6 +1471,490 @@ template cpgrid::Entity<0> createEntity(const CpGrid&, int, bool);
...
+    bool isPreAdapted = false;
+    for (const auto& preAdaptGrid : data_) {
+        isPreAdapted = isPreAdapted || (preAdaptGrid -> preAdapt());
+    }
+    return isPreAdapted; //current_view_data_-> preAdapt();
+}
+
+bool CpGrid::adapt()
+{
+    const std::vector<std::array<int,3>>& cells_per_dim_vec = {{3,3,3}}; // Arbitrary chosen values.
```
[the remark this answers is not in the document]
aritorto: Let's trust in your intuition then!

opm/grid/cpgrid/CpGrid.cpp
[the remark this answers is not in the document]
aritorto: Great! I'll add this in the list of things to be discussed, just in case

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1516,610 +2031,1764 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
...
+            const auto& shiftedLevel = markedElemLevel - preAdaptMaxLevel-1;
+            // Build auxiliary LGR for the refinement of this element
+            const auto& [elemLgr_ptr,
+                         parentCorners_to_equivalentRefinedCorners,
+                         parentFace_to_itsRefinedFaces,
+                         parentCell_to_itsRefinedCells,
+                         refinedFace_to_itsParentFace,
+                         refinedCell_to_itsParentCell]
+                = current_view_data_-> refineSingleCell(cells_per_dim_vec[shiftedLevel], elemIdx);
+            markedElem_to_itsLgr[ elemIdx ] = elemLgr_ptr;
```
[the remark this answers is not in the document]
aritorto: If you're refereeing to markedElem_to_itsLgr [elemIdx] = elemLgr_ptr, I would like to mention a few things. Recall that now the strategy for refinement is to refine each single cell independently of the rest of the marked elements. In this way, each marked element ( "markedElem" ) get its own tiny-auxiliary-lgr ( "its Lgr"). So here, we have a container, markedElem_to_itsLgr with size total amount of cells of the "starting grid" (grid where the selection and marking of element have been done). All the entries are empty except for the "elemIdx" of marked elements, where we store a pointer to the single-cell-refinement of the cell with index "elemIdx" . 

Once we refine all the marked elements, then we store entities very carefully (without repetition) in each refined-level-grid and , later on, on the leaf-grid-view. Here is where assignedrEfinedLevel plays a role. For example, 
- Marked elements with indices 5, 9, 11, 12, 13.
- We want the refinement of cells 5, 9, and 11, to belong to the same refined-level-grid, let's say, refined-level1-grid, and cells 12 and 13 , refined-level2-grid. Then, assignRefinedLevel[x] = 1, for x=5,9,11, and ssignRefinedLevel[x] = 2, for x=12,13. For the non-marked elements, it's equal to 0. 


I hope this helps!

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1516,610 +2031,1764 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
...
+
+            const auto& childrenCount = cells_per_dim_vec[shiftedLevel][0]*cells_per_dim_vec[shiftedLevel][1]*cells_per_dim_vec[shiftedLevel][2];
+            std::vector<int> refinedChildrenList(childrenCount);
+
+            for (int refinedCell = 0; refinedCell < childrenCount; ++refinedCell) {
+
+                elemLgrAndElemLgrCell_to_adaptedCell[{elemIdx, refinedCell}] = cell_count;
+                adaptedCell_to_elemLgrAndElemLgrCell[cell_count] = {elemIdx, refinedCell};
```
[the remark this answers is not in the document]
aritorto: In the next line we do cell_count+=1, so every iteration of the loop gets a new value of cell_count, therefore, no rewriting risk. 

The role of these containers elemLgrAndElemLgrCell_to_..... and ...._to_elemLgrAndElemLgrCell (some for faces, and corenrs), is to relate refined cells (faces/corners) from each individual single-cell-refinement (refinement of the cell with index "elemLgr", which has refined cells with indeces "elemLgrCell"- yes, I know- I couldn't come up with a better name :/ )  with actual refined-level-grids and the leaf-grid-view (built and stored in future code).

So, somehow we have floating around the code tiny auxiliary refinements that are purely implementation details and are not stored in the CpGrid object at all. Those tiny auxiliary refinements are the single-cell-refinements, one per marked element. 

I hope this helps. Any clearer variable name is also welcome. Maybe elemIdxAndRefinedCellInElemIdxRefiement_to_..... (leaf grid view or refined level grid)?

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1516,610 +2031,1764 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
...
+
+                elemLgrAndElemLgrCell_to_refinedLevelAndRefinedCell[{elemIdx, refinedCell}] = { markedElemLevel, refined_cell_count_vec[shiftedLevel]};
+                refinedLevelAndRefinedCell_to_elemLgrAndElemLgrCell[{markedElemLevel, refined_cell_count_vec[shiftedLevel]}] = {elemIdx, refinedCell};
+                refinedChildrenList[refinedCell] = refined_cell_count_vec[shiftedLevel];
+                refined_cell_count_vec[shiftedLevel] +=1;
+
+            }
+            refined_parent_to_children_cells_vec[elemLevel][elemLevelIdx] = std::make_pair( markedElemLevel, refinedChildrenList);
+            for (const auto& [markedCorner, lgrEquivCorner] : parentCorners_to_equivalentRefinedCorners) {
+                cornerInMarkedElemWithEquivRefinedCorner[markedCorner].push_back({elemIdx, lgrEquivCorner});
```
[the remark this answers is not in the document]
aritorto: Yes! Just a tiny fix: from 0 to (cells_per_dim_vec[shiftedLevel][0] +1) * (cells_per_dim_vec[shiftedLevel][1] +1 ) * (cells_per_dim_vec[shiftedLevel][2]+1)

tests/cpgrid/adapt_cpgrid_test.cpp
```diff
@@ -0,0 +1,563 @@
...
+                for(const auto& coord: point.center())
+                    BOOST_TEST(std::isfinite(coord));
+
+            }
+            for(const auto& cell: adapted_leaf.geomVector<3>()) {
+                auto equiv_cell_iter = blockRefinement_leaf.geomVector<3>().begin();
+                while (cell.center() != equiv_cell_iter->center()) {
+                    ++equiv_cell_iter;
+                }
+                CHECK_COORDINATES(cell.center(), equiv_cell_iter->center());
```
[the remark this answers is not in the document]
aritorto: I had some trouble with that lambda, but I did add the check of the iterator not being at the end.

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1516,610 +2031,1764 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
...
-            leaf_corners[level_to_leaf_corners[l][corner]]
-                = level_data.geometry_.geomVector(std::integral_constant<int,3>()) -> get(corner);
+    // -- Refined to adapted cells --
+    for (int shiftedLevel = 0; shiftedLevel < static_cast<int>(refined_cell_count_vec.size()); ++shiftedLevel){
+        refined_level_to_leaf_cells_vec[shiftedLevel].resize(refined_cell_count_vec[shiftedLevel]);
+        for (int cell = 0; cell < refined_cell_count_vec[shiftedLevel]; ++cell) {
+            const auto& elemLgr = refinedLevelAndRefinedCell_to_elemLgrAndElemLgrCell[{shiftedLevel + auxiliaryIdx, cell}][0];
+            const auto& elemLgrCell = refinedLevelAndRefinedCell_to_elemLgrAndElemLgrCell[{shiftedLevel + auxiliaryIdx, cell}][1];
+            const auto& adaptedCellIdx = elemLgrAndElemLgrCell_to_adaptedCell[{elemLgr, elemLgrCell}];
+            refined_level_to_leaf_cells_vec[shiftedLevel][cell] = adaptedCellIdx;
```
blattms, line 2261: Would this simpler statement work?
```suggestion
            const auto& elemLgrCellPair = refinedLevelAndRefinedCell_to_elemLgrAndElemLgrCell[{shiftedLevel + auxiliaryIdx, cell}];
            const auto& adaptedCellIdx = elemLgrAndElemLgrCell_to_adaptedCell[elemLgrCellPair];
            refined_level_to_leaf_cells_vec[shiftedLevel][cell] = adaptedCellIdx;
```
```

refined_level_to_leaf_cells_vec[shiftedLevel][cell] = elemLgrAndElemLgrCell_to_adaptedCell[refinedLevelAndRefinedCell_to_elemLgrAndElemLgrCell[elemLgrCellPair]

**Review by blattms — COMMENTED** [review]
Next batch of comments.

I am still in CpGrid.cpp. Now function `CpGrid::identifyRefinedCornersPerLevel`

File: opm/grid/cpgrid/CpGridData.cpp (hunk 4 above)
Comment (blattms) [reply opm/grid/cpgrid/CpGridData.cpp:2311]: Can't the width, height, length be computed from the triplet `{a, b, c}` before? `width=1./a`, etc.

Comment (aritorto) [reply opm/grid/cpgrid/CpGridData.cpp:2311]: I would say we do not need any height, width, or length vectors here, since we are refining single cells, so we can just use {1., 1., 1.}. 

Heights, widths, and lengths were introduced due to the old-refinement-approach, refining a block of cells treating it as a huge single cell. The role of heights, widths, and lengths was to make sure that we have the refined children cells of each parent cell occupying the volume of their parent cell. 

When we set {1., 1., 1.}, and, let's say we want {nx, ny, nz} children cells in each direction, then the volume of each children - in the unit cube - is (1./nx)*(1./ny) *(1./nz). This is perfect for refinement of a single-cell. ( Yet not perfect for a block of at least two cells with different heights, widths, and/or lengths).

File: opm/grid/cpgrid/CpGrid.cpp [hunk 34]
```diff
@@ -1516,610 +2031,1764 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
...
+            } // end-cell_to_face
+            // Refined cell to face.
+            refined_cell_to_face_vec[shiftedLevel].appendRow(aux_refined_cell_to_face.begin(), aux_refined_cell_to_face.end());
+            // Get the cell geometry.
+            const auto& elemLgrGeom =  (*(elemLgrData->geometry_.geomVector(std::integral_constant<int,0>())))[elemLgrCellEntity];
+
+            // Create a pointer to the first element of "refined_cell_to_point" (required as the fourth argement to construct a Geometry<3,3> type object).
+            int* indices_storage_ptr = refined_cell_to_point_vec[shiftedLevel][cell].data();
+            refined_cells_vec[shiftedLevel][cell] = cpgrid::Geometry<3,3>(elemLgrGeom.center(), elemLgrGeom.volume(),
+                                                                          allLevelCorners, indices_storage_ptr);
```
blattms, line 3222: To me it seems that refined_cell_to_point_vec is a temporary container. It's entries are later copied to An instance of CpGridData.cell_to_point and the container will be deleted. Doesn't this mean that indices_storage_ptr will become a dangling pointer?

blattms: Maybe we should/can later just reset this pointer in all the Geometries?
Or use the members on CpGridData for populating them directly?

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1471,6 +1471,490 @@ template cpgrid::Entity<0> createEntity(const CpGrid&, int, bool);
...
+
+        // Further Refined grid Attributes
+        //
+        // Populate some attributes of the level LGR
+        (*data_[refinedLevelGridIdx]).level_data_ptr_ = &(this -> data_);
+        (*data_[refinedLevelGridIdx]).level_ = refinedLevelGridIdx;
+        this -> lgr_names_[lgr_name_vec[level]] = refinedLevelGridIdx; // {"name_lgr", level}
+        (*data_[refinedLevelGridIdx]).child_to_parent_cells_ = refined_child_to_parent_cells_vec[level];
+        (*data_[refinedLevelGridIdx]).cell_to_idxInParentCell_ = refined_cell_to_idxInParentCell_vec[level];
+        (*data_[refinedLevelGridIdx]).level_to_leaf_cells_ =  refined_level_to_leaf_cells_vec[level];
```
blattms, line 1867: I have the feeling this here leads to dangling pointers in the Geometry instances that you have populated in all those helper functions above. The pointers in there will still point to the temporary containers refined*, but those containers will be destroyed eventually.

Would it be possible to already create std::shared_ptr<CpGridData> for all the levels and the new leaf before caling those functions. THen we can pass references to the containers in there to all these functions to populate them directly?

Here is a rough sketch of my idea:
```
std::shared_ptr<CpGridData> level3_data = {};
std::vector<std::reference_wrapper<Dune::cpgrid::EntityVariableBase<cpgrid::Geometry<0,3>>>> refined_corners_vec(levels);
refined_corners_vec[3] = *(refinedLevel_geometries.geomVector(std::integral_constant<int,3>()));
...
// populate all the containes ...
....
```

The alternative would be to correct all the pointers in the Geometry instances after copying to CpGridData members here. I think both is OK. Just keep in mind that the copies also cost time and memory.

blattms (commented):
Next batch.

I still want to check the functions in line 3440 and below (CpGrid::isRefinedCornerInInteriorLgr, etc.).

opm/grid/cpgrid/CpGrid.cpp
blattms: I think there is an easier and more elegant solution:

Use std::swap or the swap member function instead of assignment. 
```
(*data_[refinedLevelGridIdx]).child_to_parent_cells_.swap(refined_child_to_parent_cells_vec[level]);
...
```
That is supposed to simply exchange the underlying pointers, which is exactly what we want here.  For this we might need implement swap member functions for `SparseTable`, that calls swap on all members, and `OrientedEntityTable` would simply hve `using super_t::swap;`.

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1516,610 +2031,1764 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
...
             }
+        } // end-if-elemMark==1
+    } // end-elem-for-loop
+}
+
+void CpGrid::defineChildToParentAndIdxInParentCell(std::vector<std::vector<std::array<int,2>>>& refined_child_to_parent_cells_vec,
+                                                   std::vector<std::vector<int>>& refined_cell_to_idxInParentCell_vec,
+                                                   std::vector<std::array<int,2>>& adapted_child_to_parent_cells,
+                                                   std::vector<int>& adapted_cell_to_idxInParentCell,
+                                                   std::map<std::array<int,2>,std::array<int,2>> refinedLevelAndRefinedCell_to_elemLgrAndElemLgrCell,
```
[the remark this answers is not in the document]
aritorto: They do not. I'll change then void to tuple return type. Thanks!

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1516,610 +2031,1764 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
...
+    // Rewrite only the entries of adapted cells that have a parent cell
+    for (int cell = 0; cell < cell_count; ++cell) {
+        const auto& elemLgr = adaptedCell_to_elemLgrAndElemLgrCell[cell][0];
+        // (elemLgr == -1) represents that the adapted cell is equivalent to a cell from the starting grid (current_view_data_)
+        if (elemLgr != -1) {
+            // Search for the level where the parent cell was born, and its index in that level grid.
+            // Notice that elemLgr is a cell index in current_view_data_ (the starting grid where elements got marked for (further) refinement).
+            const auto& element = Dune::cpgrid::Entity<0>(*current_view_data_, elemLgr, true); // elemLgr == parent cell index in starting grid.
+            const auto& elemLevel = element.level(); // parent cell level (where the parent cell was born, might not coincide with the starting grid).
+            const auto& elemLevelIdx = (elemLevel == 0) ? element.getOrigin().index() : element.getLevelElem().index(); // parent cell index in the level it was born.
```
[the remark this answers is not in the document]
aritorto: I see your point. Unfortunately, getOrigin() returns 
- the parent entity if hasFather(), or,
- for cells on the leaf-grid-view, without father, returns the equivalent entity on the level that entity was born [To be fixed now: current code only considers level 0 as possible grids. I'll change this!], otherwise, 
- returns itself. g
getLevelElem() instead, it's supposed to be used only for leaf-grid-view cells, and it returns the equivalent refined or coarse cell on the level-grid where it was born. 

So, I would say you're right and we can simplify that piece of code, however, getOrigin() needs to consider all possible levels. I'll fix that. Great suggestion!

opm/grid/cpgrid/Entity.hpp
aritorto, line 1: Commit message is wrong. The changes were done in Entity::grtOrigin()

opm/grid/cpgrid/CpGrid.cpp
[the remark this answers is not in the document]
aritorto: I did simplify a bit this piece of code, however, I couldn't get rid of some if-statements. The main reason behind this is that we do not want always a father of an equivalent level element. For example, all the coarse cells not involved in any refinement (neither pre-adapt refined-grids nor new-refined-grids created in the current call of adapt) will get a child_to_parent_ value. They are "by default" {-1,-1}.

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1516,610 +2031,1764 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
...
+            // Populate child_to_parent_ relationships from the previous (existing grids before calling adapt).
+            assert(elemLgr == -1);
+            const auto& elemIdx = adaptedCell_to_elemLgrAndElemLgrCell[cell][1];
+            // Search for the level where the parent cell was born, and its index in that level grid.
+            // Notice that elemLgr is a cell index in current_view_data_ (the starting grid where elements got marked for (further) refinement).
+            const auto& element = Dune::cpgrid::Entity<0>(*current_view_data_, elemIdx, true); // elemIdx of equivalent cell in the starting grid.
+            if(element.hasFather()) {
+                adapted_child_to_parent_cells[cell] = {element.father().level(), element.father().index()};
+                const auto& elemLevel = element.level(); // parent cell level (level where the parent cell was born, might not coincide with the starting grid).
+                const auto& elemLevelIdx = (elemLevel == 0) ? element.getOrigin().index() : element.getLevelElem().index(); // parent cell index in the level it was born.
```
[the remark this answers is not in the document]
aritorto: See above for a brief description of what getOrigin() and getelvelElem() do. Now, there is Entity::getEquivLevelElem() which does not throw, returns the equivalent entity in the level it was born, for all levels, including 0. 

(Maybe it's worth it to say that getLevelElem() assumes that the grid is the leaf-grid-view (it throws otherwise, that means it throws when the grid associated with the current entity is any level grid, including level 0).

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1516,610 +2031,1764 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
...
+        const auto& level =  shiftedLevel + startingGridIdx +1;
+        // Every refined cell has a parent cell
+        for (int cell = 0; cell < refined_cell_count_vec[shiftedLevel]; ++cell) {
+            const auto& elemLgr = refinedLevelAndRefinedCell_to_elemLgrAndElemLgrCell[{level, cell}][0];
+            // (elemLgr == -1) means that the adapted cell is equivalent to a cell from the starting grid (current_view_data_)
+            assert(elemLgr != -1);
+            // Search for the level where the parent cell was born, and its index in that level grid.
+            // Notice that elemLgr is a cell index in current_view_data_ (the starting grid where elements got marked for (further) refinement).
+            const auto& element = Dune::cpgrid::Entity<0>(*current_view_data_, elemLgr, true);  // elemLgr == parent cell index in starting grid.
+            const auto& elemLevel = element.level(); // parent cell level (level where the parent cell was born, might not coincide with the starting grid).
```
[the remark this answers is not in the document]
aritorto: In the current version, the code it's a bit simpler. In any case, elemLgr represent the element index of a cell that got marked to be refined, in the "starting grid", current_view_data_. On the other hand, elemLevel here represents the level where the marked element with index "elemLgr" was born. This might belevel 0, or any preAdapt refined-level-grid. If current_view_data_ coincides with level zero grid, then elemLevel is equal to 0. If current_view_data_ represents a mixed grid with coarse and refined cells, with "grid.maxLevel() == 3", then elemLevel could be 0,1,2, or 3, depending on where it was born.

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1516,610 +2031,1764 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
...
+                                                  std::vector<std::array<int,2>>& leaf_to_level_cells,
+                                                  std::map<std::array<int,2>,std::array<int,2>> elemLgrAndElemLgrCell_to_refinedLevelAndRefinedCell,
+                                                  std::map<std::array<int,2>,std::array<int,2>> refinedLevelAndRefinedCell_to_elemLgrAndElemLgrCell,
+                                                  const std::vector<int> refined_cell_count_vec,
+                                                  std::map<std::array<int,2>,int> elemLgrAndElemLgrCell_to_adaptedCell,
+                                                  std::unordered_map<int,std::array<int,2>> adaptedCell_to_elemLgrAndElemLgrCell,
+                                                  const int cell_count)
+{
+    const int& preAdaptMaxLevel = this->maxLevel();
+    const int& auxiliaryIdx = (preAdaptMaxLevel>0) ? (preAdaptMaxLevel +1) : 1;
```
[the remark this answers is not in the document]
aritorto: Totally right. I guess I was a bit confused. I'll correct that. Thanks!

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1516,610 +2031,1764 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
...
+    // -- Adapted to {level, cell index in that level}  --
+    leaf_to_level_cells.resize(cell_count);
+    for (int cell = 0; cell < cell_count; ++cell) {
+        const auto& elemLgr = adaptedCell_to_elemLgrAndElemLgrCell[cell][0];
+        const auto& elemLgrCell = adaptedCell_to_elemLgrAndElemLgrCell[cell][1];
```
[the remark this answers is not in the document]
aritorto: It's an array. I forgot about these structured bindings. I'll keep them in mind now, thanks!

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1516,610 +2031,1764 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
...
+        // elemLgr == -1 means that this adapted cell is equivalent to a cell from the starting grid. So we need to find out the level where that equivalent
+        // cell was born, as well as its cell index in that level.
+        if (elemLgr == -1) {
+            const auto& element = Dune::cpgrid::Entity<0>(*current_view_data_, elemLgrCell, true);
+            const auto& levelCellIdx = (element.level() == 0) ? element.getOrigin().index() : element.getLevelElem().index();
```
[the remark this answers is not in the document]
aritorto: Probably with the previous comments it's not necessary to describe these methods here too : )

opm/grid/cpgrid/CpGrid.cpp
aritorto: I thought the other version was slightly more readers-friendly. This line does simplify the code though

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1516,610 +2031,1764 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
...
+                adaptedCorn =  elemLgrAndElemLgrCorner_to_adaptedCorner[{elemLgr, elemLgrCorn}];
+            }
+            else{
+                // Corner might have vanished - Search its equivalent lgr-corner in that case -
+                // last lgr where the corner appears -
+                int lastAppearanceLgr = 0; // It'll get rewritten
+                int lastAppearanceLgrEquivCorner = 0; // It'll get rewritten
+                if ((elemLgr ==-1) && (!cornerInMarkedElemWithEquivRefinedCorner[elemLgrCorn].empty())) {
+                    lastAppearanceLgr = cornerInMarkedElemWithEquivRefinedCorner[elemLgrCorn].back()[0];
+                    lastAppearanceLgrEquivCorner = cornerInMarkedElemWithEquivRefinedCorner[elemLgrCorn].back()[1];
```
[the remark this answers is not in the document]
aritorto: It seems that for array (this is the case here const std::array<int, 2>) std::tie does not fit (it fits for std::tuple,std::tuple<int&, int&>)

opm/grid/cpgrid/CpGrid.cpp
aritorto: Great catch! Many thanks. Now it fixed. most of the attributes now get swapped.  swap added in EntitityVaraibleBase and OrientedEntityTable.

opm/grid/cpgrid/CpGrid.cpp
aritorto: Indeed, thanks for catching it! It's fixed using your suggestion f implementing swap. See last commits :)

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1516,610 +2031,1764 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
...
+                        // To locate vanished corners, we need a while-loop, since {elemLgr, elemLgrcorner} leads to
+                        // {neighboringElemLgr, neighboringElemLgrCornerIdx}, which might have also vanished.
+                        // Then, use the lastest appearance of the current corner, meaning, the first (and unique one - by construction) that
+                        // gives elemLgrAndElemLgrCorner_to_adaptedCorner.count({lastAppearanceElemLgr, lastAppearanceElemLgCorner}) == 1).
+                        lastAppearanceLgr = vanishedRefinedCorner_to_itsLastAppearance[{elemLgr, elemLgrCorn}][0];
+                        // this corner lays on the area occupied by a coarse face that got refined and belonged to two marked elements.
+                        // Get the index of this corner with respect to the greatest marked element index.
+                        lastAppearanceLgrEquivCorner = vanishedRefinedCorner_to_itsLastAppearance[{elemLgr, elemLgrCorn}][1];
+                        while (elemLgrAndElemLgrCorner_to_adaptedCorner.count({lastAppearanceLgr, lastAppearanceLgrEquivCorner}) == 0) {
```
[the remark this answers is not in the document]
aritorto: There is some improvement now in this piece of code (and places with similar code), but still using count(). I'll try to replace it by find soon

aritorto:
@blattms many thanks for your valuable feedback! I believe I handled all/most of your comments (being the main: (1) change count-argument with find-approach, and (2) solve the dangling pointers issue via swap). 

I would say this PR is ready for a "final review"

aritorto:
jenkins build this please

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -2969,14 +2969,13 @@ void CpGrid::populateLeafGridCells(Dune::cpgrid::EntityVariableBase<cpgrid::Geom
             const auto& preAdaptFace = face.index();
             int adaptedFace = 0; // It will be rewritten.
             // Face is stored in adapted_faces
-            if ( elemLgrAndElemLgrFace_to_adaptedFace.count({elemLgr, preAdaptFace}) == 1) {
+            if ( elemLgrAndElemLgrFace_to_adaptedFace.find({elemLgr, preAdaptFace}) != elemLgrAndElemLgrFace_to_adaptedFace.end()) {
                 adaptedFace = elemLgrAndElemLgrFace_to_adaptedFace[{elemLgr, preAdaptFace}];
```
blattms, line 2973: Please check all similar occurrences (previously `...count()==1`) again. I suggested to used find instead of count to get rid off multiple lookups.
But your change here still has two identical lookups/searches (one in line 2972 (`find`) and one in line 2973 (`operator[]`).

My suggestd change here would be:
```
if ( auto candidate = elemLgrAndElemLgrFace_to_adaptedFace.find({elemLgr, preAdaptFace}); candidate != elemLgrAndElemLgrFace_to_adaptedFace.end()) {
  adaptedFace = *candidate;
}
```
Note that there is only one lookup. In the if-clause we simply dereference the iterator, which is fast.

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -3060,11 +3059,11 @@ void CpGrid::populateRefinedCells(std::vector<Dune::cpgrid::EntityVariableBase<c
             for (int corn = 0; corn < 8; ++corn) {
                 int refinedCorn;
                 const auto& preAdaptCorn = preAdapt_cell_to_point[corn];
-                if ( elemLgrAndElemLgrCorner_to_refinedLevelAndRefinedCorner.count({elemLgr, preAdaptCorn}) == 0 ) {
+                if ( elemLgrAndElemLgrCorner_to_refinedLevelAndRefinedCorner.find({elemLgr, preAdaptCorn}) == elemLgrAndElemLgrCorner_to_refinedLevelAndRefinedCorner.end() ) {
                     // Corner might have vanished - Search its equivalent lgr-corner in that case -
                     // last lgr where the corner appears -
                     std::array<int,2> lastAppearanceLgr_lgrCorner = {0, 0}; // It'll be rewritten
-                    if(markedElemAndEquivRefinedCorn_to_corner.count({elemLgr, preAdaptCorn}) == 1) {
+                    if(markedElemAndEquivRefinedCorn_to_corner.find({elemLgr, preAdaptCorn}) != markedElemAndEquivRefinedCorn_to_corner.end()) {
```
blattms, line 3066: Please store iterator returned by find and use that to get the value. see other comment.

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -2739,23 +2736,24 @@ void CpGrid::populateLeafGridFaces(Dune::cpgrid::EntityVariableBase<cpgrid::Geom
...
+        const auto& preAdapt_face_to_point = grid_or_elemLgr_data.face_to_point_[elemLgrFace];
         // Add the amount of points to the count num_points.
         num_points += preAdapt_face_to_point.size();
 
         for (int corn = 0; corn < static_cast<int>(preAdapt_face_to_point.size()); ++corn) {
             int adaptedCorn = 0; // It'll get rewritten.
             const auto& elemLgrCorn = preAdapt_face_to_point[corn];
             // Corner is stored in adapted_corners
-            if ( elemLgrAndElemLgrCorner_to_adaptedCorner.count({elemLgr, elemLgrCorn}) == 1)  {
+            if ( elemLgrAndElemLgrCorner_to_adaptedCorner.find({elemLgr, elemLgrCorn}) !=  elemLgrAndElemLgrCorner_to_adaptedCorner.end()) {
```
blattms, line 2756: Please store iterator returned by find and use that to get the value. see other comment.

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -2958,7 +2958,7 @@ void CpGrid::populateLeafGridCells(Dune::cpgrid::EntityVariableBase<cpgrid::Geom
             }
             // Corner is stored in adapted_corners
             else {
-                assert(  elemLgrAndElemLgrCorner_to_adaptedCorner.count({elemLgr, preAdaptCorn}) == 1);
+                assert(  elemLgrAndElemLgrCorner_to_adaptedCorner.find({elemLgr, preAdaptCorn}) != elemLgrAndElemLgrCorner_to_adaptedCorner.end());
```
blattms, line 2961: Please store iterator returned by find and use that to get the value below. see other comment.

opm/grid/cpgrid/EntityRep.hpp
```diff
@@ -234,6 +234,7 @@ namespace Dune
             using V::data;
             using V::operator[];
             using V::resize;
+            using V::swap;
```
blattms, line 237: Doesn't really hurt, but is not needed.

opm/grid/cpgrid/CpGrid.cpp
aritorto: Opps! Sorry, it seems I got half of your suggestion back then. I'll fix it (here and similar places) Thanks!

opm/grid/cpgrid/OrientedEntityTable.hpp
```diff
@@ -174,6 +174,7 @@ namespace Dune
             using super_t::clear;
             using super_t::appendRow;
             using super_t::allocate;
+            using super_t::swap;
```
blattms, line 177: I think this could be skipped.

opm/grid/CpGrid.hpp
blattms: Understood

blattms (approved):
Thanks a lot for the updates.

aritorto:
jenkins build this please

PR #724, opened 2024-05-28
Add scoped and named variable to avoid using out-of-scope temporary.

atgeirr:
Only called when using the bda GPU block jacobi ILU0 preconditioner. Risk of crashes should be very low, but not zero, so we fix it.


bska (approved):
Makes sense&ndash;thanks.  I'll merge into master.

PR #729, opened 2024-06-24
metis without scotch even if ptscotch is installed

lisajulia:
test if we can use purely metis even if ptscotch is installed (thats how it should be actually)
Then for scotch support, I'd add separate ScotchPartition -files


lisajulia:
jenkins build this opm-common=4115 please

PR #733, opened 2024-07-04
Fix Non-Zoltan/Non-Metis Build Configurations

bska:
The `partition_methods` object must also be defined in builds that have neither Zoltan nor Metis.  This will typically happen only for very spartan sequential builds.

While here, also mark the object as `const` to prevent inadvertent changes.


bska:

blattms:
Great catch, thanks.

lisajulia:
@bska: Thanks!

PR #732, opened 2024-06-28
Global refinement supported for corner point grids

aritorto:
This PR 
1. extends the support of DUNE-Grid interface for CpGrid (corner point grids), defining (and testing) CpGrid::globalRefine(int). The default refinement is to divide each cell into 8 refined cells ( 2 refined cells in each direction: x,-y-, and z-axis).
2. is based on OPM/opm-grid#731, therefore, it should be merged after it. 
3. does not affect the update of the OPM reference-manual.

PR #734, opened 2024-07-05
Inactive cells allowed in and/or outside LGRs for general CpGrid

aritorto:
This PR is (unnecessary /unfortunately) based on OPM/opm-grid#732.

Improvements of refinement for CpGrid, namely, fewer restrictions on the grid features.

BEFORE                                     NOW
- only Cartesian grid               - general Corner Point grid
- only all active cells                - there can be inactive cells anywhere (inside or outside LGRs).

This significantly increase the flexibility of the selection of cells to be refined.

Current restrictions:
- Serial run.
- When two different LGRs share at least one face, certain compatibility of the amount of refined cells from LGR1 and the refined cells fromLGR2 is required.


aritorto:
jenkins build this please

blattms:
Please rebase this on master once you deem this PR finished. That will ease the final review. Thanks.

aritorto:
It's rebased now on master, but the attempt to avoid std::find made some tests fail, therefore, draft mode for now. Thanks for your feedback!

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -2051,20 +2051,51 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
...
+    // Determine the assigned level for the refinement of each marked cell
+    std::vector<int> assignRefinedLevel(data_[0]->size(0));
+    for(const auto& element: elements(this->leafGridView())) {
+        std::cout<< "elemIdx: " << element.index() <<  " global: " << globalCell()[element.index()] << std::endl;
+        std::array<int,3> ijk;
+        getIJK(element.index(), ijk);
+        for (int level = 0; level < static_cast<int>(startIJK_vec.size()); ++level) {
+            bool belongsToLevel = true;
+            for (int c = 0; c < 3; ++c) {
+                belongsToLevel = belongsToLevel && ( (ijk[c] >= startIJK_vec[level][c]) && (ijk[c] <= endIJK_vec[level][c]) );
```
blattms, line 2067: Please double check the comparison. Shouldn't the first be `>` as startIJK is the index of the first cell to be refined? And maybe the last one  `<`?

aritorto: Great comment! That was the reason of failing tests

aritorto:
jenkins build this please

opm/grid/cpgrid/CpGrid.cpp
blattms: Well, I am definitely wrong about startIJK. :see_no_evil:

aritorto: The comment was pointing correctly to the issue :) Nevertheless, only endIJK comparison was wrong

blattms (changes requested):
This looks already good.

Would be perfect if you would throw exceptions instead of the new asserts.

The current test only execute the functions for adding LGrs and even ignore occurring exceptions.  I think we should add some real checks (number of elements, correct parent information, etc.) and an exception should be a failure (e.g. `BOOST_CHECK(false)` in the catch clauses)

aritorto:
jenkins build this please

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1999,6 +1999,11 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
             OPM_THROW_NOLOG(std::logic_error, "Adding LGRs to a distributed grid is not supported, yet.");
         }
     }
+
+    assert(cells_per_dim_vec.size() == startIJK_vec.size());
+    assert(cells_per_dim_vec.size() == endIJK_vec.size());
+    assert(cells_per_dim_vec.size() == lgr_name_vec.size());
```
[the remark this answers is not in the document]
aritorto: I extended the test, replaced asserts by throws, improved throw-messages, and throw also when an LGR contains all inactive cells. Many thanks for your feedback!

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1999,6 +1999,10 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
             OPM_THROW_NOLOG(std::logic_error, "Adding LGRs to a distributed grid is not supported, yet.");
         }
     }
+    if ( (cells_per_dim_vec.size() != startIJK_vec.size())  || (lgr_name_vec.size() != startIJK_vec.size())) {
+        OPM_THROW(std::invalid_argument, "Invalid size. Provide for each cell-block: subdivisions per cell and LGR name.");
```
blattms, line 2003:
```suggestion
        OPM_THROW(std::invalid_argument, "Sizes of provided vectors with start and end IJK and LGR names need to match.");
```

opm/grid/cpgrid/CpGridData.cpp
```diff
@@ -2204,13 +2204,21 @@ bool CpGridData::hasNNCs(const std::vector<int>& cellIndices) const
...
+                // valid startIJK and endIJK for each patch
+                validPatch = validPatch && (startIJK_vec[patch][c] < endIJK_vec[patch][c]);
+                if (!validPatch) {
+                    OPM_THROW(std::invalid_argument, "Invalid ijk-indices, e.g. end-i smaller than start-i.");
+                }
+            }
         }
     }
+    else {
+        OPM_THROW(std::invalid_argument, "Invalid size. Provide for each start-ijk an end-ijk.");
```
blattms, line 2220:
```suggestion
        OPM_THROW(std::invalid_argument, "Sizes of provided vectors with start and end IJK need to match");
```

opm/grid/cpgrid/CpGridData.cpp
```diff
@@ -2204,13 +2204,21 @@ bool CpGridData::hasNNCs(const std::vector<int>& cellIndices) const
...
-            // valid startIJK and endIJK for each patch
-            assert(startIJK_vec[patch][c] < endIJK_vec[patch][c]);
+    if (startIJK_vec.size() == endIJK_vec.size()) {
+        for (unsigned int patch = 0; patch < startIJK_vec.size(); ++patch) {
+            bool validPatch = true;
+            for (int c = 0; c < 3; ++c) {
+                // valid startIJK and endIJK for each patch
+                validPatch = validPatch && (startIJK_vec[patch][c] < endIJK_vec[patch][c]);
+                if (!validPatch) {
+                    OPM_THROW(std::invalid_argument, "Invalid ijk-indices, e.g. end-i smaller than start-i.");
```
blattms, line 2214:
```suggestion
                    OPM_THROW(std::invalid_argument, "Invalid ijk-indices. end[" std::to_string(c) "] not larger than start[" std::to_string(c) "]");
```
Not 100% sure whether this suggestion actually compiles. I did not understand the i in the original message

tests/cpgrid/inactiveCell_lgr_test.cpp
```diff
@@ -0,0 +1,583 @@
...
+        {
+            const auto& view = grid.levelGridView(l);
+            for (const auto& element: elements(view)){
+                BOOST_CHECK_EQUAL(element.level(), l);
+            }
+        }
+        //////////////////////////////
+    }
+    catch (const std::exception& e) {
+        std::cout << e.what() << "'\n";
```
blattms, line 302: I still don't like this one.

Imagine that there is an exception in try block, then we only print an error here, but the test will seem fine to ctest. Either don't do try-catch (-> test fails if there is an exception) or `BOOST_CHECK(false)` here.

blattms (commented):
Sorry, I have got a few new comments.

opm/grid/cpgrid/CpGridData.cpp
aritorto: I meant to say that at least one of the coordinates is invalid, meaning that, if startIJK = {i1, j1, k1} and endIJK={ i2,j2,k2}, then, i1>=i2. I agree, it was not clear. Thanks for the suggestions!

tests/cpgrid/inactiveCell_lgr_test.cpp
aritorto: I modified the test removing the try-catch, replacing it by BOOST_THROW in the last test-case "throw_allInactiveCells_in_lgr". I hope that this is kind of equivalent to what you suggested.

aritorto:
jenkins build this please

PR #735, opened 2024-07-10
[refactor] Local and global ids taking into account CpGrid with LGRs

aritorto:
Each entity has an unique id along the entire hierarchical grid. This means that an entity appearing in more than one level has exactly the same id in each of those levels. 

This PR refactors the computation of local and global ids to support CpGrids with LGRs.
This PR is not relevant for the update of the reference manual. 

To be improved: avoid, if possible, repetition of code.

### DIFF
--- a/opm/grid/CpGrid.hpp
+++ b/opm/grid/CpGrid.hpp
@@ -1025,8 +1025,8 @@ namespace Dune
                                            int preAdaptMaxLevel);
 
         void updateCornerHistoryLevels(const std::vector<std::vector<std::array<int,2>>>& cornerInMarkedElemWithEquivRefinedCorner,
-                                       std::map<std::array<int,2>,std::array<int,2>> elemLgrAndElemLgrCorner_to_refinedLevelAndRefinedCorner,
-                                       std::unordered_map<int,std::array<int,2>> adaptedCorner_to_elemLgrAndElemLgrCorner,
+                                       const std::map<std::array<int,2>,std::array<int,2>>& elemLgrAndElemLgrCorner_to_refinedLevelAndRefinedCorner,
+                                       const std::unordered_map<int,std::array<int,2>>& adaptedCorner_to_elemLgrAndElemLgrCorner,
                                        const int& corner_count,
                                        const std::vector<std::array<int,2>>& preAdaptGrid_corner_history,
                                        const int& preAdaptMaxLevel);
--- a/opm/grid/CpGrid.hpp
+++ b/opm/grid/CpGrid.hpp
@@ -1024,6 +1024,13 @@ namespace Dune
                                            const std::vector<std::array<int,3>>& cells_per_dim_vec,
                                            int preAdaptMaxLevel);
 
+        void updateCornerHistoryLevels(const std::vector<std::vector<std::array<int,2>>>& cornerInMarkedElemWithEquivRefinedCorner,
+                                       std::map<std::array<int,2>,std::array<int,2>> elemLgrAndElemLgrCorner_to_refinedLevelAndRefinedCorner,
+                                       std::unordered_map<int,std::array<int,2>> adaptedCorner_to_elemLgrAndElemLgrCorner,
+                                       const int& corner_count,
+                                       const std::vector<std::array<int,2>>& preAdaptGrid_corner_history,
+                                       const int& preAdaptMaxLevel);
+
         /// @brief Get the ijk index of a refined corner, given its corner index of a single-cell-refinement.
         ///
         /// Given a single-cell, we refine it in {nx, ny, nz} refined children cells (per direction). Then, this single-cell-refinement
--- a/opm/grid/cpgrid/CpGrid.cpp
+++ b/opm/grid/cpgrid/CpGrid.cpp
@@ -518,11 +518,10 @@ CpGrid::scatterGrid(EdgeWeightMethod method,
         setupRecvInterface(importList, *cell_scatter_gather_interfaces_);
 
         distributed_data_[0]->distributeGlobalGrid(*this,*this->current_view_data_, computedCellPart);
-        // global_id_set_.insertIdSet(*distributed_data_[0]);
         (*global_id_set_ptr_).insertIdSet(*distributed_data_[0]);
         distributed_data_[0]-> index_set_.reset(new cpgrid::IndexSet(distributed_data_[0]->cell_to_face_.size(),
                                                                      distributed_data_[0]-> geomVector<3>().size()));
-       
+
 
 
         current_view_data_ = distributed_data_[0].get();
@@ -1561,6 +1560,9 @@ bool CpGrid::adapt(const std::vector<std::array<int,3>>& cells_per_dim_vec,
     const int levels = static_cast<int>(cells_per_dim_vec.size());
     const int preAdaptMaxLevel = this->maxLevel();
 
+    // Copy corner history - needed to compute later ids, empty vector if the grid to be adapted is level 0 grid. 
+    const auto& preAdaptGrid_corner_history = (preAdaptMaxLevel>0) ? current_view_data_->corner_history_ : std::vector<std::array<int,2>>();
+
     // To store/build refined level grids.
     std::vector<std::vector<std::shared_ptr<Dune::cpgrid::CpGridData>>> refined_data_vec(levels,this -> data_);
     std::vector<std::shared_ptr<Dune::cpgrid::CpGridData>> refined_grid_ptr_vec(levels);
@@ -1948,11 +1950,23 @@ bool CpGrid::adapt(const std::vector<std::array<int,3>>& cells_per_dim_vec,
     (*data_[levels + preAdaptMaxLevel +1]).index_set_ = std::make_unique<cpgrid::IndexSet>(data_[levels + preAdaptMaxLevel +1]->size(0),
                                                                                            data_[levels + preAdaptMaxLevel +1]->size(3));
     (*data_[levels + preAdaptMaxLevel +1]).local_id_set_ = std::make_shared<const cpgrid::IdSet>(*data_[levels + preAdaptMaxLevel +1]);
-    // TO DO: How to modified logical_cartesian_size_
     (*data_[levels + preAdaptMaxLevel +1]).logical_cartesian_size_ =  (*data_[0]).logical_cartesian_size_;
 
     // Update the leaf grid view
     current_view_data_ = data_.back().get();
+
+    updateCornerHistoryLevels(cornerInMarkedElemWithEquivRefinedCorner,
+                              elemLgrAndElemLgrCorner_to_refinedLevelAndRefinedCorner,
+                              adaptedCorner_to_elemLgrAndElemLgrCorner,
+                              corner_count,
+                              preAdaptGrid_corner_history,
+                              preAdaptMaxLevel);
+
+    this->global_id_set_ptr_ = std::make_shared<cpgrid::GlobalIdSet>(*current_view_data_);
+    for (int level = 0; level < levels; ++level) {
+        const int refinedLevelGridIdx = level + preAdaptMaxLevel +1;
+        this->global_id_set_ptr_->insertIdSet(*data_[refinedLevelGridIdx]);
+    }
 
     // Print total amount of cells on the adapted grid
     Opm::OpmLog::info(std::to_string(markedElem_count) + " elements have been marked.\n");
@@ -3309,6 +3323,46 @@ void CpGrid::updateLeafGridViewGeometries( /* Leaf grid View Corners arguments *
                           cornerInMarkedElemWithEquivRefinedCorner,
                           cells_per_dim_vec,
                           preAdaptMaxLevel);
+}
+
+void CpGrid::updateCornerHistoryLevels(const std::vector<std::vector<std::array<int,2>>>& cornerInMarkedElemWithEquivRefinedCorner,
+                                       std::map<std::array<int,2>,std::array<int,2>> elemLgrAndElemLgrCorner_to_refinedLevelAndRefinedCorner,
+                                       std::unordered_map<int,std::array<int,2>> adaptedCorner_to_elemLgrAndElemLgrCorner,
+                                       const int& corner_count,
+                                       const std::vector<std::array<int,2>>& preAdaptGrid_corner_history,
+                                       const int& preAdaptMaxLevel)
+                                       //std::vector<std::vector<std::array<int,2>>>& levels_corner_history,
+                                       // std::vector<std::array<int,2>>& leaf_corner_history)
+{
+    // levels_corners_history.resize((this->maxLevel - preAdaptMaxLevel));
+    for (int level = preAdaptMaxLevel+1; level < (this->maxLevel()+1); ++level) {
+        std::cout<< "level: " << level << " preAdaptMaxLevel: " << preAdaptMaxLevel << " data_[level] -> size(3): " << data_[level]->size(3) <<std::endl;
+        data_[level]->corner_history_.resize( data_[level] ->size(3), std::array<int,2>({-1,-1}));
+        //  levels_corners_history[level].resize( data_[level] ->size(3), std::array<int,2>({-1,-1}));
+    }
+    // corner_history_ for levels 0, level 1, ..., preAdapt-maxLevel (maximum level before calling (again) adapt) should be already populated
+    // corner_history_[ new corner ] = {-1,-1}
+    // corner_history_[ corner equivalent to a corner in a previous level ] = { level where the corner was born, its index in that level grid}.
+    for (int corner = 0; corner < static_cast<int>(cornerInMarkedElemWithEquivRefinedCorner.size()); ++corner) {
+        if (!cornerInMarkedElemWithEquivRefinedCorner[corner].empty()) {
+            const auto& [refinedLevel, refinedCorner] = elemLgrAndElemLgrCorner_to_refinedLevelAndRefinedCorner[cornerInMarkedElemWithEquivRefinedCorner[corner].back()];
+                data_[refinedLevel]->corner_history_[refinedCorner] = preAdaptGrid_corner_history.empty() ? std::array<int,2>{{0, corner}} :  preAdaptGrid_corner_history[corner];
+                 std::cout<< "level: " << refinedLevel << " refinedCorner: " << refinedCorner << " corner: " << corner <<std::endl; 
+            }
+        }
+    // corner_history_ leaf grid view
+    for ( int leafCorner = 0; leafCorner < corner_count; ++leafCorner){
+        data_.back()->corner_history_.resize(corner_count);
+        const auto& [elemLgr, elemLgrCorner] = adaptedCorner_to_elemLgrAndElemLgrCorner[leafCorner];
+        if (elemLgr != -1) {
+            const auto& [refinedLevel, refinedCorner] = elemLgrAndElemLgrCorner_to_refinedLevelAndRefinedCorner[{elemLgr, elemLgrCorner}];
+            data_.back()->corner_history_[leafCorner] = { refinedLevel, refinedCorner};
+        }
+        else {
+            data_.back()->corner_history_[leafCorner] =  preAdaptGrid_corner_history.empty() ? std::array<int,2>{{0, elemLgrCorner}} : preAdaptGrid_corner_history[elemLgrCorner];
+        }
+        std::cout<<"leaf corner: " << leafCorner<<  " level: " <<  data_.back()->corner_history_[leafCorner][0] << " corner: " <<  data_.back()->corner_history_[leafCorner][1] <<std::endl; 
+    }
 }
 
 
--- a/opm/grid/cpgrid/CpGrid.cpp
+++ b/opm/grid/cpgrid/CpGrid.cpp
@@ -3331,8 +3331,6 @@ void CpGrid::updateCornerHistoryLevels(const std::vector<std::vector<std::array<
                                        const int& corner_count,
                                        const std::vector<std::array<int,2>>& preAdaptGrid_corner_history,
                                        const int& preAdaptMaxLevel)
-//std::vector<std::vector<std::array<int,2>>>& levels_corner_history,
-// std::vector<std::array<int,2>>& leaf_corner_history)
 {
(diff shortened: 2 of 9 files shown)


blattms (commented):
A few comments/questions.

opm/grid/cpgrid/Indexsets.hpp
```diff
@@ -236,15 +236,29 @@ namespace Dune
             IdType computeId(const EntityType& e) const
             {
                 IdType myId = 0;
-                for( int c=0; c<EntityType::codimension; ++c )
+                // Leaf grid view is a mixed of coarse and fined cells. When grid_ represents one of the levels 0,1,..., maxLevel,
+                // we sum up all the entities until the previous level of the provided entity e.
+                if (grid_.levelData().size() > 1) {
+                    // levelData() is a vector containing in each entry: level 0, level 1, ..., level maxLevel, leaf grid view (last entry).
+                    // Sum up the previos levels index sets. When grid_.level_ == 0, the following loop does nothing.
+                    for (int level = 0; level < grid_.getGridIdx(); ++level) { // getGridIdx() is an integer to access to the right grid_ in levelData() vector.
```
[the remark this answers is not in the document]
aritorto: For corners and faces we cannot do e.level() unfortunately

aritorto:
jenkins build this please

tests/cpgrid/lookupdataCpGrid_test.cpp
```diff
@@ -112,7 +112,7 @@ void lookup_check(const Dune::CpGrid& grid)
     const Dune::MultipleCodimMultipleGeomTypeMapper<Dune::CpGrid::LeafGridView> leafMapper(leaf_view, Dune::mcmgElementLayout());
     const Dune::MultipleCodimMultipleGeomTypeMapper<Dune::CpGrid::LevelGridView> level0Mapper(level0_view, Dune::mcmgElementLayout());
 
-    const auto& leaf_idSet = (*data.back()).localIdSet();
+    // const auto& leaf_idSet = (*data.back()).localIdSet();
```
blattms, line 115: Please remove the line.

tests/cpgrid/inactiveCell_lgr_test.cpp
```diff
@@ -294,6 +294,75 @@ void testInactiveCellsLgrs(const std::string& deckString,
...
+
+    const auto& global_id_set_ptr = grid.global_id_set_ptr_;
+
+    std::set<int> allIds_set;
+    std::vector<int> allIds_vec;
+    allIds_vec.reserve(data.back()->size(0) + data.back()->size(3));
+    const auto& leaf_view = grid.leafGridView();
+    for (const auto& element: elements(leaf_view)){
+        const auto& localId = data.back()->local_id_set_->id(element);
+        const auto& globalId = data.back()->global_id_set_->id(element);
```
blattms, line 306: Maybe use DUNE's grid methods to access the id sets:
```suggestion
        const auto& localId = grid.localIdSet().id(element);
        const auto& globalId = grid.globalIdSet().id(element);
```

tests/cpgrid/inactiveCell_lgr_test.cpp
```diff
@@ -294,6 +294,75 @@ void testInactiveCellsLgrs(const std::string& deckString,
...
+
+        for (const auto& element: elements(level_view)){
+            const auto& localId = (*level_localIdSet).id(element);
+            const auto& globalId = (*level_globalIdSet).id(element);
+            // In serial run, local and global id coincide:
+            BOOST_CHECK_EQUAL(localId, globalId);
+            levelIds_set.insert(localId);
+            levelIds_vec.push_back(localId);
+            if (element.isLeaf()) { // Check that the id of a cell not involved in any further refinement appears on the IdSet of the leaf grid view.
+                BOOST_CHECK( std::find(allIds_set.begin(), allIds_set.end(), localId) != allIds_set.end());
```
blattms, line 346: An alternative would be search in the elements of the leaf grid view for the element with the same id using `std::find_if). That should exists and coincide with `element`.

tests/cpgrid/grid_lgr_test.cpp
```diff
@@ -470,17 +470,93 @@ void refinePatch_and_check(Dune::CpGrid& coarse_grid,
         //
         const auto& leaf_view = coarse_grid.leafGridView();
         const auto& level0_view = coarse_grid.levelGridView(0);
+
         Dune::MultipleCodimMultipleGeomTypeMapper<Dune::CpGrid::LeafGridView> leafMapper(leaf_view, Dune::mcmgElementLayout());
         Dune::MultipleCodimMultipleGeomTypeMapper<Dune::CpGrid::LevelGridView> level0Mapper(level0_view, Dune::mcmgElementLayout());
+
         const auto& leaf_idSet = (*data[startIJK_vec.size()+1]).local_id_set_;
+        const auto& leaf_globalIdSet = (*data[startIJK_vec.size()+1]).global_id_set_;
```
blattms, line 478: Maybe do it the DUNE way and ask the grid for the id set (`grid.globalIdSet())?

tests/cpgrid/adapt_cpgrid_test.cpp
```diff
@@ -208,8 +208,8 @@ void markAndAdapt_check(Dune::CpGrid& coarse_grid,
         const auto& grid_view = coarse_grid.leafGridView();
 
         Dune::MultipleCodimMultipleGeomTypeMapper<Dune::CpGrid::LeafGridView> adaptMapper(grid_view, Dune::mcmgElementLayout());
-        const auto& adapt_idSet = adapted_leaf.local_id_set_;
-
+        const auto& leaf_localIdSet = data.back()->local_id_set_;
+
```
blattms, line 212: Nitpick: Please remove trailing whitespace

tests/cpgrid/adapt_cpgrid_test.cpp
```diff
@@ -208,8 +208,8 @@ void markAndAdapt_check(Dune::CpGrid& coarse_grid,
         const auto& grid_view = coarse_grid.leafGridView();
 
         Dune::MultipleCodimMultipleGeomTypeMapper<Dune::CpGrid::LeafGridView> adaptMapper(grid_view, Dune::mcmgElementLayout());
-        const auto& adapt_idSet = adapted_leaf.local_id_set_;
-
+        const auto& leaf_localIdSet = data.back()->local_id_set_;
```
blattms, line 211: Maybe get the id set from the grid.
```suggestion
        const auto& leaf_localIdSet = grid_view.grid().localIdSet();
```

blattms (commented):
I only reviewed the tests for now.

Would it be possible to anticipate the number of ids of all entities in some refined grid and test that number? Maybe for global refined ones?

tests/cpgrid/inactiveCell_lgr_test.cpp
aritorto: Good point. globalIdSet() was not defined for CpGridData. In the coming commit, it's defined and used!

aritorto: For CpGrid, both localIdSet() and globalIdSet() existed already. I misunderstood your comment. In any case, now there is also a method globalIdSet() for CpGridData

aritorto: In any case, for CpGrid, localIdSet() and globalIdSEt() both return the same (*global_id_set_prt_). Therefore, for testing purposes, I call the methods for each level grid and the leaf grid view (and not the entire grid)

tests/cpgrid/inactiveCell_lgr_test.cpp
aritorto: I added a few lines, in this direction, but also kept the old version checking only the id values.

tests/cpgrid/grid_lgr_test.cpp
aritorto: Same as before, I use localIdSet() and (new method) globalIdSet() for each level grid, or the leaf grid view (CpGridData objects)

tests/cpgrid/adapt_cpgrid_test.cpp
aritorto: Similar as before, I use the CpGridData ones, and grid.globalIdSet() sometimes too

aritorto:
In the last commit, the total ids is compared and checked with the total amount of cells and corners, per level grid, including the leaf grid view.  @blattms I went through all your comments, so I run jenkins again, and look forward to further feedback!

aritorto:
@blattms @bska Could you have a look or give any hint on how this refactor-of-ids make one test fail?? Thanks in advance!

bska:
> Could you have a look or maybe give me any hint on how this refactor-of-ids make one test fail?

I haven't looked at the details here.  I just notice that the [failure](https://ci.opm-project.org/job/opm-grid-PR-builder/lastCompletedBuild/testReport/(root)/mpi/compareECLFiles_flow_UDQ_01/) is of the form
> ```
> Unified restart files, sequence  1
> 
> Comparing SEQNUM ...  done.
> Comparing INTEHEAD ... 
> 
> Keyword: INTEHEAD, origin Restart, sequence 1
> 
> Value index                 = 206
> (first value, second value) = (2, 7)
> 
> Non floating point values not identical 
> 
> Keyword: INTEHEAD, origin Restart, sequence 1
>
> Value index                 = 207
> (first value, second value) = (24, 12)
> ```
meaning elements [206](https://github.com/OPM/opm-common/blob/bb5f8b4362332edfd211bd1d183a53f1f17c892e/opm/output/eclipse/VectorItems/intehead.hpp#L145) and [207](https://github.com/OPM/opm-common/blob/bb5f8b4362332edfd211bd1d183a53f1f17c892e/opm/output/eclipse/VectorItems/intehead.hpp#L146) of `INTEHEAD` differ at report step 1 between the candidate and reference solutions.

Since these two values are not supposed to change unless there's a change to the case itself, maybe we've accidentally changed the time step and/or clock?

aritorto:
Thanks @bska for your remark! let's see how it goes...

aritorto:
@bska Could you also give me a hint of what the numbers "first" and "second" mean for each element? For element with index 206, first =2, second =7. For element, 207, first = 24, second =12. Is it "first" representing "the candidate" and second, representing "the reference solution"? That would mean that "first" and "second" should coincide for each element. 

Do you mean that this PR refactoring entities ids could have affected the time step or the clock? Or that might have happened somewhere else?  

Any (extra) hint is welcome! Thanks in advance!

bska:
> Could you also give me a hint of what the numbers "first" and "second" mean for each element

Right, so those refer to the "first" and "second" cases, respectively.  Those, in turn, are determined by the order in which the result files are passed to the `compareECL` utility for analysis.  The log shows the following
> ```
> Keyword: INTEHEAD, origin Restart, sequence 1
>
> Value index                 = 206
> (first value, second value) = (2, 7)
>
> Program threw an exception: [/var/lib/jenkins/workspace/opm-grid-PR-builder/deps/opm-common/test_util/EclRegressionTest.cpp:171] Non floating point values not identical 
> Comparing '/var/lib/jenkins/workspace/opm-grid-PR-builder/deps/opm-tests/actionx/opm-simulation-reference/flow/UDQ-01' to '/var/lib/jenkins/workspace/opm-grid-PR-builder/mpi/build-opm-simulators/tests/results/flow+udq_undefined_2/UDQ-01'.
> ````
and especially the `Comparing ... to ... ` line here implies that `compareECL` is called as
```sh
compareECL reference candidate atol rtol
```
with `atol` and `rtol` being the absolute and relative tolerances, respectively.  This is further supported if we look at the Jenkins [`run-regressionTest.sh`](https://github.com/OPM/opm-simulators/blob/ed5f371133976fd58cc98b6354553be59ca32e2b/tests/run-regressionTest.sh#L62) comparison script.

In effect, the above is a long-winded way of saying that, at least in this case, "first" refers to the on-disk reference solutions while "second" refers to the simulation results obtained from running the PR.  I think I would argue that we should probably print "reference" and "candidate" in place of "first" and "second", but then we need to go through our complete testing framework to ensure we always call `compareECL` in that order.  I seem to recall there being at least one exception to that order in the past.

> That would mean that "first" and "second" should coincide for each element.

Normally, yes, and especially for these two elements.  The only case in which we expect changes to `INTEHEAD[206]` or `INTEHEAD[207]` is when we explicitly alter the report times in the input files&ndash;e.g., by changing `START`, `TSTEP`, or `DATES`.

> Do you mean that this PR refactoring entities ids could have affected the time step or the clock?

It's possible although, as I said, I haven't looked at the details here.  We assign these two items in [`InteHEAD::calendarDate()`](https://github.com/OPM/opm-common/blob/bb5f8b4362332edfd211bd1d183a53f1f17c892e/opm/output/eclipse/InteHEAD.cpp#L542-L543) and the input `timePoint` structure is constructed in [`getSimulationTimePoint()`](https://github.com/OPM/opm-common/blob/bb5f8b4362332edfd211bd1d183a53f1f17c892e/opm/output/eclipse/InteHEAD.cpp#L870-L871) based on the simulation start time&ndash;`START` keyword&ndash;and the ["elapsed" time](https://github.com/OPM/opm-common/blob/bb5f8b4362332edfd211bd1d183a53f1f17c892e/opm/output/eclipse/CreateInteHead.cpp#L629) in seconds.  The latter is ultimately determined on the simulator side, specifically in [`EclWriter<>::writeOutput()`](https://github.com/OPM/opm-simulators/blob/ed5f371133976fd58cc98b6354553be59ca32e2b/opm/simulators/flow/EclWriter.hpp#L460)

> Or that might have happened somewhere else?

Also possible.  Are you able to reproduce the symptom on your local development machine?  If not, debugging this could be a bit hard...

aritorto:
Many thanks for the explanation @bska. I'm not sure how to reproduce this on my machin, for debugging. Is it an issue appearing in parallel run? Would it look like "mpirun -np 2  ../../opm-simulators/.../flow  udq_undefined_2.DATA" ?? (due to "mpi.compareECLFiles_flow+UDQ-01" failing) Or is it related to EclRegressionTest.cpp?

bska:
> I'm not sure how to reproduce this on my machine, for debugging

The first thing to try is to "just" run the test locally and see if it fails at all.  Just go to your top-level opm-simulators **build** directory and type
```sh
ctest -VV -R UDQ-01
```
and look at the print-out.  If you do get a failure here, and especially if it looks like the same kind of failure as on the Jenkins CI system, then we have something to go on.

> Is it an issue appearing in parallel run?

Probably not. For what it's worth, the regression tests that compare against on-disk reference solutions are all run in serial&ndash;effectively as `mpirun -np 1 ...`.

aritorto:
Many thanks again @bska, I'll start with "ctest -VV -R UDQ-01" and see what I get locally

aritorto:
I get a few warnings:


43: Warning: Unsupported keywords or keyword items:
143: 
143:   NOECHO: keyword not supported
143:   In file: /home/aritorto/Opm/opm-tests/actionx/UDQ-01.DATA, line 122
143: 
143:   RPTSOL: keyword not supported
143:   In file: /home/aritorto/Opm/opm-tests/actionx/UDQ-01.DATA, line 146
143: 
143: Creating corner-point grid from keywords COORD, ZCORN and others
143: 3 fluid phases are active
143: 
143: Warning: Error in keyword RPTSOL, unrecognized mnemonic THPRES
143: In /home/aritorto/Opm/opm-tests/actionx/UDQ-01.DATA line 146.
143: 
143: Warning: Error in keyword RPTSOL, unrecognized mnemonic EQUIL
143: In /home/aritorto/Opm/opm-tests/actionx/UDQ-01.DATA line 146.

and:

Warning: Summary output requested for UDQ FUVAR1
143: In /home/aritorto/Opm/opm-tests/actionx/UDQ-01.DATA line 182
143: No definition for this UDQ found in the SCHEDULE section
143: 
143: Warning: Summary output requested for UDQ FUVAR2
143: In /home/aritorto/Opm/opm-tests/actionx/UDQ-01.DATA line 184
143: No definition for this UDQ found in the SCHEDULE section
143: 
143: Processing grid
143: Total number of active cells: 3 / total pore volume: 22500 RM3
143: 
143: Warning: Unhandled summary keyword DATE/DAY
143: In /home/aritorto/Opm/opm-tests/actionx/UDQ-01.DATA line 179
143: 
143: Warning: Unhandled summary keyword DATE/MONTH
143: In /home/aritorto/Opm/opm-tests/actionx/UDQ-01.DATA line 179
143: 
143: Warning: Unhandled summary keyword DATE/YEAR
143: In /home/aritorto/Opm/opm-tests/actionx/UDQ-01.DATA line 179


I would say that locally, I do not get the failure:

omparing summary files 
143: 
143: Checking 12  vectors  ...  done.
143: 
143: !Warning, summary and/or RSM - file not found, hence not compared. 
143: 
143: 
143: !Warning, rft files not found, hence not compared. 
143: 
143: 
143: Program finished 
143: 
1/1 Test #143: compareECLFiles_flow+UDQ-01 ......   Passed    0.64 sec

The following tests passed:
        compareECLFiles_flow+UDQ-01

100% tests passed, 0 tests failed out of 1

Total Test time (real) =   0.65 sec

aritorto:
[udq.log](https://github.com/user-attachments/files/16309731/udq.log)

bska:
> Apart from a few warnings, the test passes in my machine...

Agreed.  This all looks fine.  Let's try to rerun this PR to see if the symptom persists.

bska:
jenkins build this please

blattms (commented):
Sorry, there have been some pending comments that I never submitted as a review.
I probably did quit my browser before pressing the review button.

My apologies.

opm/grid/CpGrid.hpp
```diff
@@ -1024,6 +1024,13 @@ namespace Dune
                                            const std::vector<std::array<int,3>>& cells_per_dim_vec,
                                            int preAdaptMaxLevel);
 
+        void updateCornerHistoryLevels(const std::vector<std::vector<std::array<int,2>>>& cornerInMarkedElemWithEquivRefinedCorner,
+                                       std::map<std::array<int,2>,std::array<int,2>> elemLgrAndElemLgrCorner_to_refinedLevelAndRefinedCorner,
+                                       std::unordered_map<int,std::array<int,2>> adaptedCorner_to_elemLgrAndElemLgrCorner,
```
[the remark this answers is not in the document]
aritorto: I can double-check, but I remembered trying to pass argument with const & (once they existed, i.e., previously populated), but in the case of (unordered) maps, I was getting errors. I do not remember the error messages. Maybe something related to hash/(not)being-hashable. 

I can add the cont & and see which error appears, if any.

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -3311,6 +3325,41 @@ void CpGrid::updateLeafGridViewGeometries( /* Leaf grid View Corners arguments *
...
 }
 
+void CpGrid::updateCornerHistoryLevels(const std::vector<std::vector<std::array<int,2>>>& cornerInMarkedElemWithEquivRefinedCorner,
+                                       std::map<std::array<int,2>,std::array<int,2>> elemLgrAndElemLgrCorner_to_refinedLevelAndRefinedCorner,
+                                       std::unordered_map<int,std::array<int,2>> adaptedCorner_to_elemLgrAndElemLgrCorner,
+                                       const int& corner_count,
+                                       const std::vector<std::array<int,2>>& preAdaptGrid_corner_history,
+                                       const int& preAdaptMaxLevel)
+//std::vector<std::vector<std::array<int,2>>>& levels_corner_history,
+// std::vector<std::array<int,2>>& leaf_corner_history)
```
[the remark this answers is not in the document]
aritorto: Indeed

opm/grid/CpGrid.hpp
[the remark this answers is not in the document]
aritorto: My memory was kind of in the right direction, in the sense that I did try not to copy but there were errors so I removed the const references from the arguments. 

When I try to pass the map as a const reference instead of a copy, I get:

error: passing ‘const std::map<std::array<int, 2>, std::array<int, 2> >’ as ‘this’ argument discards qualifiers [-fpermissive]

...

usr/include/c++/12/bits/stl_map.h:502:7: note:   in call to ‘std::map<_Key, _Tp, _Compare, _Alloc>::mapped_type& std::map<_Key, _Tp, _Compare, _Alloc>::operator[](const key_type&) [with _Key = std::array<int, 2>; _Tp = std::array<int, 2>; _Compare = std::less<std::array<int, 2> >; _Alloc = std::allocator<std::pair<const std::array<int, 2>, std::array<int, 2> > >; mapped_type = std::array<int, 2>; key_type = std::array<int, 2>]’
  502 |       operator[](const key_type& __k)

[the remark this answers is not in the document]
atgeirr: You cannot use operator[] on a const map (or const map reference), because it might insert a new object. You can use at() instead (throws instead of inserting a default-constructed item if lookup fails for a key), or find() for more fine-grained control (returns end iterator if not found).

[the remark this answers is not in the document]
blattms: The reason for this error is that `operator[]` in maps is not there as a const method, because if there is no entry for the key in a map, then such an entry will be automatically inserted. This requires a mutable objects. That is probably not what you want here

Hence methods that get a const map may never use `operator[]`, but always need to use either `at(key)` which will return the value if it is there or throw or `find(key)` which will return an iterator (that might either be the end iterator or point to another entry) that you need to check.

So to avoid bugs it is always a good idea to use the `at` function that has a const version instead. It will throw an exception if there is no entry for the key:
```
const auto& val = map.at(key);
```
or use find and do the error checking yourself (cumbersome):
```
auto candidate = map.find(key);
assert (candidate != map.end && candidate->first == key); // use an assert if we are 100% sure that this value should exist.
const auto& val = candidate->second;
```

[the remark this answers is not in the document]
aritorto: Thanks @atgeirr and @blattms for the explanation!

aritorto:
jenkins build this please

blattms:
Copying will be fixed in an upcoming PR.

aritorto:
Thanks for merging Markus. Copying should be fixed here. I'm working on another branch to avoid copying in other methods (not involved in this PR). I'm not sure what you meant in the previous comment

PR #737, opened 2024-07-12
Correct the compiler macro for METIS support

lisajulia:
This caused an error when having only metis installed.

### DIFF
--- a/opm/grid/common/MetisPartition.cpp
+++ b/opm/grid/common/MetisPartition.cpp
@@ -38,7 +38,7 @@ namespace cpgrid
 {
 
 // We want to use METIS, but if METIS is installed as part of the ScotchMetis package, then the following options are not available.
-#if !(defined(IS_SCOTCH_METIS_HEADER) && IS_SCOTCH_METIS_HEADER)
+#if !IS_SCOTCH_METIS_HEADER
 void setMetisOptions(const std::map<std::string, std::string>& optionsMap, idx_t* options) {
 
     // Initialize all options to default values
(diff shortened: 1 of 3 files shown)


lisajulia:
jenkins build this please

lisajulia:
jenkins build this please

opm/grid/common/MetisPartition.cpp
```diff
@@ -38,7 +38,7 @@ namespace cpgrid
 {
 
 // We want to use METIS, but if METIS is installed as part of the ScotchMetis package, then the following options are not available.
-#if !(defined(IS_SCOTCH_METIS_HEADER) && IS_SCOTCH_METIS_HEADER)
+#if !(defined(IS_SCOTCH_METIS_HEADER) && IS_SCOTCH_METIS_HEADER > 0)
```
blattms, line 41: This looks a bit to complicated. Isn't the macro always defined?
How about changing FindMETIS a bit instead?
```
diff --git a/cmake/Modules/FindMETIS.cmake b/cmake/Modules/FindMETIS.cmake
index 1f3f623a8..4a7e2758e 100644
--- a/cmake/Modules/FindMETIS.cmake
+++ b/cmake/Modules/FindMETIS.cmake
@@ -35,6 +35,8 @@ if (METIS_INCLUDE_DIRS)
   string(FIND "${metisheader}" "SCOTCH_METIS_PREFIX" IS_SCOTCH_METIS_HEADER)
   if(NOT IS_SCOTCH_METIS_HEADER EQUAL "-1")
     set(IS_SCOTCH_METIS_HEADER 1)
+  else()
+    set(IS_SCOTCH_METIS_HEADER 0)
   endif()
 endif()
```
the we can do a simple 
```
#if IS_SCOTCH_METIS_HEADER
```

lisajulia:
jenkins build this opm-common=4140 please

PR #739, opened 2024-07-22
Pass maps with const ref to avoid copying

aritorto:
Replace [ ] operator by .at( ) for (unordered) maps so that passing maps as function arguments by constant reference (when they are not supposed to be modified by the functions) is possible.

### DIFF
--- a/opm/grid/cpgrid/CpGrid.cpp
+++ b/opm/grid/cpgrid/CpGrid.cpp
@@ -2128,7 +2128,7 @@ void CpGrid::refineAndProvideMarkedRefinedRelations( /* Marked elements paramete
                                                      int& cell_count,
                                                      std::vector<std::vector<int>>& preAdapt_level_to_leaf_cells_vec,
                                                      /* Additional parameters */
-                                                     const std::vector<std::array<int,3>>& cells_per_dim_vec)
+                                                     const std::vector<std::array<int,3>>& cells_per_dim_vec) const
 {   // Each marked element for refinement (mark equal to 1), will be refined individuality, creating its own Lgr. The element index will
     // be also used to identify its lgr. Even though, in the end, all the refined entities will belong to a unique level grid.
     // For this reason, we associate "-1" with those elements that are not involved in any refinement and will appear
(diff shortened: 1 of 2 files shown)


aritorto:
jenkins build this please

aritorto:
As I promised, @blattms here is a PR avoiding the unnecessary copying of (unordered) maps in all methods related to refine/adapt code :)

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -2236,10 +2235,10 @@ CpGrid::defineChildToParentAndIdxInParentCell(std::map<std::array<int,2>,std::ar
     for (int shiftedLevel = 0; shiftedLevel < static_cast<int>(refined_cell_count_vec.size()); ++shiftedLevel) {
         refined_child_to_parent_cells_vec[shiftedLevel].resize(refined_cell_count_vec[shiftedLevel]);
         refined_cell_to_idxInParentCell_vec[shiftedLevel].resize(refined_cell_count_vec[shiftedLevel]);
-        const auto& level =  shiftedLevel + startingGridIdx +1;
+        const auto& level =  shiftedLevel + (this->maxLevel()) +1;
```
blattms, line 2238: This change seems unrelated. Please double check and explain.

blattms (commented):
This looks great. I only have a question about one line that I do not understand.

opm/grid/cpgrid/CpGrid.cpp
aritorto: It is now correct, I would say. I should have added a comment though. When calling multiple times adapt, or refining a grid that has been already refined, "startingGridIdx" is the integer to access "the current view data" that will be deleted once you refine/adapt for a second/third/... time. 
In the begging, I was not aware of this deleting process, and I wrote the code (in one of my first interpretations) in such a way that we still store those mixed grids, which is not the case in master branch (and in my merged PRs). 

Maybe this should have been part of a tiny PR called bugfix or similar, this indeed it is not related to passing maps as const references.

aritorto: To illustrate, if you have a grid and refine it once adding 2 levels, then the first mixed grid is stored in data_ in the entry 3.
data_[0] = level 0, data_[1] = level 1 grid, data_[2] = level 2, data_[3] = "current view data"
Now, you mark elements to be refined, in "current view data". Later on, this "current view data" is deleted/replaced for the refined level grid 3. Then
data_[0] = level 0, data_[1] = level 1 grid, data_[2] = level 2, data_[3] = level 3, data_[4] = (new) "current view data".


Therefore, old line (in this example), with shiftedLEvel =0 e.g.:
level = startingGridIdx + 1 =  3 + 1 = 4 which is wrong,
new line:
level = (this->maxLevel()) + 1 = 2+ 1 =3 which is correct!

I hope this helps!

blattms: Thanks. So this is a fix to a bug, that was not triggered by any of our tests.

aritorto: Exactly. Found out thanks to .at() which from now, I'll keep it in mind!

PR #740, opened 2024-07-25
Avoid capturing structured binding [&key].

atgeirr:
This is not supported until C++20, and at least for clang on mac, is not yet supported in OpenMP:

```
/Users/atgeirr/opm/src/opm-grid/opm/grid/common/MetisPartition.cpp:111:112: warning: captured structured bindings are a C++20 extension [-Wc++20-extensions]
        } else if (std::any_of(metisOptionKeysForOtherMethods.begin(), metisOptionKeysForOtherMethods.end(), [&key](const std::string& otherKey) { return otherKey == key; })) {
                                                                                                               ^
/Users/atgeirr/opm/src/opm-grid/opm/grid/common/MetisPartition.cpp:95:23: note: 'key' declared here
    for (const auto& [key,value] : optionsMap) {
                      ^
/Users/atgeirr/opm/src/opm-grid/opm/grid/common/MetisPartition.cpp:111:112: error: capturing a structured binding is not yet supported in OpenMP
        } else if (std::any_of(metisOptionKeysForOtherMethods.begin(), metisOptionKeysForOtherMethods.end(), [&key](const std::string& otherKey) { return otherKey == key; })) {
                                                                                                               ^
/Users/atgeirr/opm/src/opm-grid/opm/grid/common/MetisPartition.cpp:95:23: note: 'key' declared here
    for (const auto& [key,value] : optionsMap) {
```
I replaced the std::any_of() with calling the count() function of the set, which is simpler in any case (contains() is not yet available).

Also improve formatting nearby avoiding overlong lines.

### DIFF
--- a/opm/grid/common/MetisPartition.cpp
+++ b/opm/grid/common/MetisPartition.cpp
@@ -88,9 +88,15 @@ void setMetisOptions(const std::map<std::string, std::string>& optionsMap, int& 
         {"METIS_DBG_CONNINFO", METIS_DBG_CONNINFO}, 
         {"METIS_DBG_CONTIGINFO", METIS_DBG_CONTIGINFO}
     };
+    // Option keys not valid for the METIS_PartGraphRecursive or METIS_PartGraphKway functions.
+    std::unordered_set<std::string> metisOptionKeysForOtherMethods = {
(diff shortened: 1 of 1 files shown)


atgeirr:
jenkins build this please

PR #741, opened 2024-07-25
Recover running in parallel in a very hacky way.

blattms:
I am not proud of this, but it seems to be the easiest way of recovering our parallel runs broken by #735 which accidentally removed usage of mappings for global id set in parallel. Apologies for not catching that during my review.

This seems to recover parallel Norne runs, but I had to deactivate some of the LGR tests that fail with this change. I need more time to fix those. Hence expect some further PRs for that. Just not today.


blattms:
jenkins build this please

atgeirr:
I can confirm it solves the immediate problem, and as it is a very short and simple (if temporary) fix I will merge it right away.

PR #742, opened 2024-07-25
Add possible future connections to all methods that handle the well c…

lisajulia:
…onnections
Description is here: https://github.com/OPM/opm-simulators/pull/5488


blattms (commented):
Looks good to me.

Would be cool if we would:
- Extend the description of the PR a bit to make clear that this change is there connections that might be opened in ACTIONX and that we neglected those before.
- Add the new argument to the doxcygen docs where applicable.

blattms (approved):

PR #743, opened 2024-07-26
Fix the id sets for LGRs and reactivate tests.

blattms:
This is a follow up of #741 and completes the fix of the parallel regression and restores the state for LGRs.

To accomplish this we again ask the local id set for the global id in a sequential run. To do this correctly with multiple levels we need to make sure that we use the full entity as we need level information. 

In #741 we used the EntityRep which made our tests of ids for higher levels fail as points on finer levels that also appeared on lower ones suddenly had different ids. In addition we needed to make sure that the id_ pointer of the global id set actually pointed to the correct local id set. This is guaranteed by the CpGridData constructor and we should never reset the local_id_set_ member of CpGridData as this breaks the guarantee (unless we also reset global_id_set_).

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -112,15 +112,15 @@ list (APPEND TEST_SOURCE_FILES
 
 if(Boost_VERSION_STRING VERSION_GREATER 1.53)
 	list(APPEND TEST_SOURCE_FILES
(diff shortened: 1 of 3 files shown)


blattms:
jenkins build this please

atgeirr:
Thanks for the fix, merging.

PR #747, opened 2024-08-05
Change type of possibleFutureConnections to std::unordered_map<std::string, std::set<int>>

lisajulia:
Now it contains the global ids of the perforated cells for each well instead of the coordinates ijk.

This PR is not relevant for the reference manual.


opm/grid/CpGrid.hpp
```diff
@@ -1225,7 +1225,7 @@ namespace Dune
         ///         perforated cells local to the process, for all wells (sorted by name)
         std::pair<bool,std::vector<std::pair<std::string,bool>>>
         loadBalance(const std::vector<cpgrid::OpmWellType> * wells,
-                    const std::unordered_map<std::string, std::set<std::array<int,3>>>* possibleFutureConnections = nullptr,
+                    const std::unordered_map<std::string, std::set<int>>* possibleFutureConnections = nullptr,
```
blattms, line 1228: Same question here as in opm-simulators. Would this also work:
```suggestion
                    const std::unordered_map<std::string, std::set<int>>& possibleFutureConnections = {},
```
and internally you can test the size?

blattms (commented):
Only wondering about the same as in the companion PR in opm-simulators: Do we really need the pointers?

PR #748, opened 2024-08-07
Add fully-interior LGRs on a distributed grid

aritorto:
This PR 
- refactors some methods related to refinement to take into account the case when the grid has been distributed (before refinement),
-  tests a few cases, adding LGRs after the grid has been distributed,
-  does not imply any updates on the reference manual.

Once merged, refinement of a distributed grid will be supported in the particular case of each LGR being fully interior on a process. Namely, every selected cell for refinement must be surrounding by other interior cells of the process. 

Notice that each process can contain more than one LGR as long as each LGR is surrounding by other interior cells of the process. 

In future PRs, this restriction will be weakened/removed.


aritorto:
jenkins build this please

aritorto:
jenkins build this please

opm/grid/CpGrid.hpp
```diff
@@ -398,6 +398,9 @@ namespace Dune
         /// @brief Returns either data_ or distributed_data_(if non empty).
         const std::vector<std::shared_ptr<Dune::cpgrid::CpGridData>>& chooseData() const;
 
+        /// @brief Returns either data_ or distributed_data_(if non empty).
+        std::vector<std::shared_ptr<Dune::cpgrid::CpGridData>>& getData();
+
```
blattms, line 403: Can we stick to one name for the const and mutable function, please? (Either chooseData or getData or somthing even better)

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -624,6 +624,11 @@ const std::vector<std::shared_ptr<Dune::cpgrid::CpGridData>>& CpGrid::chooseData
     }
 }
 
+std::vector<std::shared_ptr<Dune::cpgrid::CpGridData>>& CpGrid::getData()
+{
+    return !distributed_data_.empty() ? distributed_data_ : data_;
```
blattms, line 629: The decision should be based on what current_view_data_ points to.
```suggestion
    return current_view_data_ == data_.back() ? data_ : distributed_data_;
```

blattms (commented):
I need to run, just a start.

aritorto:
[addLgrsDistGrid.log](https://github.com/user-attachments/files/16538038/addLgrsDistGrid.log)

I get different output. In any case, I set the PR as draft for now.

opm/grid/CpGrid.hpp
aritorto: Tiny improvement of the name: now both are called "currentData"

aritorto:
jenkins build this please

aritorto:
jenkins build this please

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -653,26 +656,25 @@ std::string CpGrid::name() const
...
-    if (!distributed_data_.empty()){
-        return cpgrid::Iterator<codim, All_Partition>(*current_view_data_, 0, true);
+    bool areThereLgrs =  (currentData().size() > 1);
+    if(areThereLgrs) {
+        return  cpgrid::Iterator<codim, All_Partition>( *(*current_data_)[level], 0, true);
     }
-    else{
-        return cpgrid::Iterator<codim, All_Partition>(*data_[level], 0, true);
+    else {
+        return cpgrid::Iterator<codim, All_Partition>(*current_view_data_, 0, true);
```
blattms, line 677: Do we really need if-else? Isn't the level grid view always in current_data_[level]?

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -684,11 +686,12 @@ typename CpGridTraits::template Codim<codim>::LevelIterator CpGrid::lend (int le
...
-    if (!distributed_data_.empty()){
-        return cpgrid::Iterator<codim, All_Partition>(*current_view_data_, size(codim), true);
+    bool areThereLgrs =  (currentData().size() > 1);
+    if(areThereLgrs) {
+        return  cpgrid::Iterator<codim, All_Partition>( *(*current_data_)[level], size(level, codim), true);
     }
-    else{
-        return cpgrid::Iterator<codim,All_Partition>(*data_[level], size(level, codim), true );
+    else {
+        return cpgrid::Iterator<codim, All_Partition>(*current_view_data_, size(codim), true);
```
blattms, line 694: Same here and also below: superfluous if-else?

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1673,15 +1673,15 @@ bool CpGrid::adapt(const std::vector<std::array<int,3>>& cells_per_dim_vec,
     std::vector<std::vector<int>> preAdapt_level_to_leaf_cells_vec(preAdaptMaxLevel +1);
     for (int preAdaptLevel = 0; preAdaptLevel < preAdaptMaxLevel +1; ++preAdaptLevel) {
         // Resize with the corresponding amount of cells of the preAdapt level. Deafualt {-1, empty vector} when the cell has no children.
-        if ( (*data_[preAdaptLevel]).parent_to_children_cells_.empty()){
-            preAdapt_parent_to_children_cells_vec[preAdaptLevel].resize(data_[preAdaptLevel]->size(0), std::make_pair(-1, std::vector<int>{}));
+        if ( (*data[preAdaptLevel]).parent_to_children_cells_.empty()){
+            preAdapt_parent_to_children_cells_vec[preAdaptLevel].resize(currentData()[preAdaptLevel]->size(0), std::make_pair(-1, std::vector<int>{}));
```
blattms, line 1677: Maybe use `data[preAdaptLevel]` for consistency?

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1703,13 +1703,13 @@ bool CpGrid::adapt(const std::vector<std::array<int,3>>& cells_per_dim_vec,
...
+    
     // Update/define parent_to_children_cells_ and level_to_leaf_cells_ for all the existing level grids (level 0, 1, ..., preAdaptMaxLevel), before this call of adapt.
     for (int preAdaptLevel = 0; preAdaptLevel < preAdaptMaxLevel +1; ++preAdaptLevel) {
-        (*data_[preAdaptLevel]).parent_to_children_cells_ = preAdapt_parent_to_children_cells_vec[preAdaptLevel];
-        (*data_[preAdaptLevel]).level_to_leaf_cells_ =  preAdapt_level_to_leaf_cells_vec[preAdaptLevel];
+        (*data[preAdaptLevel]).parent_to_children_cells_ = preAdapt_parent_to_children_cells_vec[preAdaptLevel];
+        (*data[preAdaptLevel]).level_to_leaf_cells_ =  preAdapt_level_to_leaf_cells_vec[preAdaptLevel];
     }
-
+
```
blattms, line 1712: Maybe remove trailing whitespace?

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1724,7 +1724,7 @@ bool CpGrid::adapt(const std::vector<std::array<int,3>>& cells_per_dim_vec,
                  adapted_cell_to_idxInParentCell] = defineChildToParentAndIdxInParentCell(refinedLevelAndRefinedCell_to_elemLgrAndElemLgrCell,
                                                                                           refined_cell_count_vec,adaptedCell_to_elemLgrAndElemLgrCell,
                                                                                           cell_count);
-
+
```
blattms, line 1727: Maybe remove trailing whitespace?

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1797,7 +1801,7 @@ bool CpGrid::adapt(const std::vector<std::array<int,3>>& cells_per_dim_vec,
                           assignRefinedLevel,
                           faceInMarkedElemAndRefinedFaces,
                           cells_per_dim_vec);
-
+
```
blattms, line 1804: Maybe remove trailing whitespace?

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1829,7 +1833,7 @@ bool CpGrid::adapt(const std::vector<std::array<int,3>>& cells_per_dim_vec,
                                     markedElemAndEquivRefinedCorn_to_corner,
                                     cornerInMarkedElemWithEquivRefinedCorner,
                                     cells_per_dim_vec);
-
+
```
blattms, line 1836: Maybe remove trailing whitespace?

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1863,75 +1867,76 @@ bool CpGrid::adapt(const std::vector<std::array<int,3>>& cells_per_dim_vec,
...
         Dune::cpgrid::EntityVariableBase<PointType>& level_mutable_face_normals = level_face_normals;
         level_mutable_face_normals.swap(mutable_refined_face_normals_vec[level]);
-
+        
         // Further Refined grid Attributes
         //
         // Populate some attributes of the level LGR
-        (*data_[refinedLevelGridIdx]).level_data_ptr_ = &(this -> data_);
-        (*data_[refinedLevelGridIdx]).level_ = refinedLevelGridIdx;
+        (*data[refinedLevelGridIdx]).level_data_ptr_ = &(this -> currentData());
```
blattms, line 1917: `&data` for consistency?

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -2079,19 +2083,20 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
...
+        // Do not throw if all cells of an LGR are inactive in a parallel run (The process might not 'see' those cells.)
+        if ((lgrs_with_at_least_one_active_cell[level] == 0) && distributed_data_.empty()) {
             OPM_THROW(std::logic_error, "LGR" + std::to_string(level+1) + " contains only inactive cells, remove it or extend the corresponding region.\n");
         }
         else {
             ++non_empty_lgrs;
         }
     }
-
+
```
blattms, line 2094: Maybe remove trailing whitespace?

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -2079,19 +2083,20 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
 
     int non_empty_lgrs = 0;
     for (int level = 0; level < static_cast<int>(startIJK_vec.size()); ++level) {
-        if (lgrs_with_at_least_one_active_cell[level] == 0) {
+        // Do not throw if all cells of an LGR are inactive in a parallel run (The process might not 'see' those cells.)
+        if ((lgrs_with_at_least_one_active_cell[level] == 0) && distributed_data_.empty()) {
```
blattms, line 2087: I think there are two options here:
1. We never throw for empty LGRs, but just issue a warning via OpmLog
2. Also do the check in parallel.

In both cases we will need to do a sum of the cells of each lgr on all processes:
```
auto localLgrCells = countActiveCellsInLGR();
auto globalLgrCells = comm_.sum(localLgrCells);
if(globalLgrCells==0) ....
```

blattms (commented):
Looks already really good. Just some minor things. for now.

I have not looked at the test, yet.

opm/grid/cpgrid/CpGrid.cpp
aritorto: You're right. No need for this if-statement (I assume also for the ones you point out below). I'll remove them. Thanks!

opm/grid/cpgrid/CpGrid.cpp
aritorto: Definitively!

opm/grid/cpgrid/CpGrid.cpp
aritorto: Now, we never throw for empty LGRs and issue a warning via Opm::log. The check is done in parallel, in the sense that, each process sees only its own cells, and then a process could conclude that an LGR is empty, when for the global view it's not empty (it contains active cells not seen by the current process).

aritorto: Now, we also have such a check when global lgrs only contain inactive cells.

aritorto:
jenkins build this please

aritorto:
[Last commit](https://github.com/OPM/opm-grid/pull/748/commits/3afcb21d1f9148ad05f65fe9b20ff15a4dad7286)  inspired in [suggestion on draft PR](https://github.com/aritorto/opm-grid/pull/16) from @blattms

aritorto:
jenkins build this please

aritorto:
I'm marking this PR ready for review ,even though I need to work a bit more on the test file, which currently does not fail when you consider "one test case per run". (This explains why there are large commented pieces on the test). It seems that it can be improved selecting in which process each cell should be sent. In the meantime, I run jenkins too.

aritorto:
jenkins build this please

aritorto:
jenkins build this please

aritorto:
Something might need to be fixed since running the all the tests defined in addLgrsOnDistrubutedGrid_test.cpp fail, but running each of them separately, it does not fail.

aritorto:
jenkins build this please

akva2:
did not look at details but sounds like a fixture initializing mpi et al.

aritorto:
thanks @akva2 I'm a bit new into mpi, so I'm not 100% about what you meant. In any case, I also noticed that, when requirements for refinements are not fulfilled, I should "throw in all the processes" (terminate) - not only one of them. I'll fix that soon. Does this happen to be related to your comment?

akva2:
not really, what i mean is if you have a test fixture which calls MPI_Init / MPI_Finalize (diretly or indirectly), you are in trouble because those can only be called once per program invocation. So it would work fine running one test, but since the fixture is constructed / destructed once per test, you'd be in trouble running multiple tests.

aritorto:
Thanks a lot for the explanation, so these last commits where I separated the text cases (one per file) in the end might make sense, if I got it correctly. I'll think about it! Thanks again @akva2

akva2:
you can see an example of the typical way of solving this in https://github.com/OPM/opm-simulators/blob/master/tests/test_broadcast.cpp

in particular note the use of BOOST_TEST_NO_MAIN and implementation of a main method.

blattms:
@akva2 We actually found out what the problem was:

It seems like Zoltan has some kind of internal state (maybe a random number generator) and you cannot rely on the partitioning being the same the second time you call it (or if you previously computed another partitioning in addition). 

I guess this might even be a feature: If you compute a partitioning and later realize that it was rubbish, you could start over and hope to get a better partitioning by chance. :sweat_smile: 

Even for our small example computing another partitioning before has quite an impact:

Below I print the owned cell indices per rank.

With an additional previous partitioning call to zoltan:
rank 0:0 1 4 5 8 9 16 20 21
rank 1:12 13 17 24 25 28 29 32 33
rank 2:2 3 6 7 10 11 18 22 23
rank 3:14 15 19 26 27 30 31 34 35

Without the previous partitioning:
rank 0:14 15 18 19 26 27 30 31 35
rank 1:2 3 6 7 10 11 22 23 34
rank 2:12 13 16 17 24 25 28 29 33
rank 3:0 1 4 5 8 9 20 21 32

Hence we need to force our partitions in this case to make this stable.

akva2:
ah that old fun. there's a PHG_RANDOMIZE_INPUT parameter that might help.

blattms:
Cool, thanks. The result may still differ between different versions, etc.

tests/cpgrid/addLgrsOnDistributedGrid_test.cpp
```diff
@@ -397,9 +397,35 @@ BOOST_AUTO_TEST_CASE(threeLgrs)
...
+                        parts[elemIdx] = 1;
+                    }
+                }
+                if (i>4){
+                    if (j<4) {
+                        parts[elemIdx] = 2;
+                    }
+                    if (j>3){
+                        parts[elemIdx] = 3;
+                    }
```
blattms, line 421: From a short glimpse:

Is there a particular reason why you do not use if-else? It seems like the second if condition always covers what a usual else branch would cover, too.

If-else might be easier to understand.

aritorto: no particular reason, I'll change it for else-if in coming commit

aritorto:
Thanks @akva2 and @blattms for your feedback. The PR is ready for review, after a jenkins check

aritorto:
jenkins build this please

blattms:

blattms (commented):
Review of new change in most files. Tests are still missing.

I will need to take another look at the global id computation, though.

aritorto:

tests/cpgrid/inactiveCell_lgr_test.cpp
```diff
@@ -81,297 +81,305 @@ void testInactiveCellsLgrs(const std::string& deckString,
     Dune::CpGrid grid;
     grid.processEclipseFormat(&es.getInputGrid(), &es, false, false, false);
 
+    if (grid.comm().size() == 0)
```
blattms, line 84: Is this really needed? 

As long as you do not call loadBalance on the grid, I think all the code should work in parallel, too. In this case all cells of the grid are on process 0 and for it this looks like  a serial grid. The other processes simply see an empty grid.

tests/cpgrid/inactiveCell_lgr_test.cpp
```diff
@@ -81,297 +81,305 @@ void testInactiveCellsLgrs(const std::string& deckString,
...
-                BOOST_CHECK( std::find(allIds_set.begin(), allIds_set.end(), localId) != allIds_set.end());
+        }
+
+        std::set<int> allIds_set;
+        std::vector<int> allIds_vec;
+        allIds_vec.reserve(data.back()->size(0) + data.back()->size(3));
+        for (const auto& element: elements(grid.leafGridView())){
+            const auto& localId = data.back()->localIdSet().id(element);
+            const auto& globalId = data.back()->globalIdSet().id(element);
+            if (grid.comm().size() == 0) { // In serial runs, local and global ids coincide.
```
blattms, line 305: Not needed as there is no loadBalance call.

tests/cpgrid/inactiveCell_lgr_test.cpp
```diff
@@ -81,297 +81,305 @@ void testInactiveCellsLgrs(const std::string& deckString,
...
+            // Check that the global_id_set_ptr_ has the correct id (id from the level where the entity was born).
+            BOOST_CHECK_EQUAL( grid.globalIdSet().id(element), (*data[element.level()]).globalIdSet().id(element.getEquivLevelElem()));
+        }
+        // Check injectivity of the map local_id_set_ (and, indirectly, global_id_set_) after adding cell ids.
+        BOOST_CHECK( allIds_set.size() == allIds_vec.size());
+
+        for (const auto& point: vertices(grid.leafGridView())){
+            const auto& localId = data.back()->localIdSet().id(point);
+            const auto& globalId = data.back()->globalIdSet().id(point);
+            if (grid.comm().size() == 0) { // In serial runs, local and global ids coincide.
```
blattms, line 319: Not needed?

aritorto:

aritorto:

aritorto:

aritorto:

blattms:
Thanks a lot for all the work. I am ready to merge this. Should we maybe reduce the massive amount of commit by rewriting history, first?

aritorto:

PR #752, opened 2024-08-27
Refactor PartitionTypeIndicator to use only Entities.

blattms:
Those have more information than EnitityRep which might be useful with LGRs/adaptivity.

This WIP as it is just there for demonstration.

### DIFF
--- a/opm/grid/cpgrid/CpGridData.cpp
+++ b/opm/grid/cpgrid/CpGridData.cpp
@@ -217,9 +217,9 @@ struct GetRowType<std::vector<E,A> >
 };
 
 PartitionType getPartitionType(const PartitionTypeIndicator& p, const EntityRep<1>& f,
-                               const CpGridData&)
+                               const CpGridData& grid)
 {
-    return p.getPartitionType(f);
+    return p.getPartitionType(Entity<3>(grid, f.index(), f.orientation()));
 }
 
 PartitionType getPartitionType(const PartitionTypeIndicator& p, int i,
--- a/opm/grid/cpgrid/PartitionTypeIndicator.cpp
+++ b/opm/grid/cpgrid/PartitionTypeIndicator.cpp
@@ -9,18 +9,18 @@ namespace Dune
 {
 namespace cpgrid
 {
-PartitionType PartitionTypeIndicator::getPartitionType(const EntityRep<0>& cell_entity) const
+PartitionType PartitionTypeIndicator::getPartitionType(const Entity<0>& cell_entity) const
 {
     if(cell_indicator_.size())
         return PartitionType(cell_indicator_[cell_entity.index()]);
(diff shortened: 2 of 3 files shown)


blattms:
jenkins build this please

opm/grid/cpgrid/PartitionTypeIndicator.cpp
```diff
@@ -9,18 +9,18 @@ namespace Dune
...
-PartitionType PartitionTypeIndicator::getPartitionType(const EntityRep<0>& cell_entity) const
+PartitionType PartitionTypeIndicator::getPartitionType(const Entity<0>& cell_entity) const
 {
     if(cell_indicator_.size())
         return PartitionType(cell_indicator_[cell_entity.index()]);
     return InteriorEntity;
 }
 
-PartitionType PartitionTypeIndicator::getPartitionType(const EntityRep<1>& face_entity) const
+PartitionType PartitionTypeIndicator::getPartitionType(const Entity<1>& face_entity) const
```
aritorto, line 19: Do we support Entity<1> for CpGrid? Don't we always use EntityRep<1> instead?

opm/grid/cpgrid/CpGridData.cpp
```diff
@@ -217,9 +217,9 @@ struct GetRowType<std::vector<E,A> >
 };
 
 PartitionType getPartitionType(const PartitionTypeIndicator& p, const EntityRep<1>& f,
-                               const CpGridData&)
+                               const CpGridData& grid)
 {
-    return p.getPartitionType(f);
+    return p.getPartitionType(Entity<3>(grid, f.index(), f.orientation()));
```
aritorto, line 222: Typo in the codim? for face, 1, for point 3. Same as below, just wondering if we also need some refactor in Entity to support a few methods there for Entity<1> for corner point grid

opm/grid/cpgrid/PartitionTypeIndicator.cpp
blattms: I think we need some further refactoring to really drop that. When distributing the grid we ask for global ids on faces.

At least this change seems to work/compile.

aritorto: I see, I'm not familiar yet with global ids for faces (I did work on global ids for cells and points, and ignore faces so far). Step by step then!

opm/grid/cpgrid/CpGridData.cpp
blattms: Good catch. I'll fix that.

blattms:
I accidentally pushed this to OPM instead of my repo. I will deleted the branch here.

blattms:
Update commit is c492f477a62065aefda98a493a971a23856d0eed

PR #753, opened 2024-08-30
Remove a few static_cast<int> 

aritorto:


aritorto:

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1528,11 +1528,11 @@ bool CpGrid::adapt(const std::vector<std::array<int,3>>& cells_per_dim_vec,
                    const std::vector<std::array<int,3>>& endIJK_vec)
 {
     // To do: support coarsening.
-    assert(static_cast<int>(assignRefinedLevel.size()) == current_view_data_->size(0));
+    assert( int(assignRefinedLevel.size()) == current_view_data_->size(0));
```
bska, line 1531: Please don't do this.  The `int()` cast is strictly worse, from a semantic point of view, than `static_cast<int>()`.  Yes, the latter is more verbose, but the verbosity does have a real benefit here.

aritorto:
int() is gone now

aritorto:

opm/grid/cpgrid/CpGrid.cpp
aritorto: Thanks for your feedback, it's gone now.

bska (approved):

PR #754, opened 2024-09-02
fixed: mark variable maybe_unused

akva2:
quells warning without mpi


akva2:

aritorto:
Thanks for catching that, my bad!

bska:
> I would merge it, but it seems I'm not allowed to.

In that case I'll merge the PR.  Thanks for looking into this, @aritorto!

PR #756, opened 2024-09-04
Precalculate TRANZ because of PINCH option 4 ALL and include as NNC.

blattms:
We calculate all transmissibilities for intersections in between the two active cells that are connected with an NNC due to PINCH. Then an NNC is added with harmonic average of those transmissibilities. That NNC is marked as stemming from PINCH for correct treatment in the simulator.

There is still some cleanup needed.

Needs OPM/opm-common#4181


opm/grid/cpgrid/processEclipseFormat.cpp
```diff
@@ -200,15 +209,106 @@ namespace cpgrid
...
+                    // Apply Multipliers
+
+                    // \todo FIXME We assume here that MULTZ does not change in the SCHEDULE or such changes have
+                    //       no effect here. This might be wrong and need fixing. Which basically means we need to store
+                    //       all the intermediate transmissibilities for the harmonic average and later apply additional
+                    //       multipliers
+                    const auto& transMult = ecl_state->getTransMult();
+                    trans = trans_between.begin();
+
+                    for (std::size_t cell_top = cell1, cell_bottom = cell_top + ecl_grid.getNX() * ecl_grid.getNY();
```
atgeirr, line 283: In this loop, trans is incremented in the loop body, unlike the one above. Is there a reason for the difference, or could it be unified?

atgeirr (commented):
It seems to me the PinchedCellInformation class is not used?

Other than that, the main algorithmic change looks fine.

opm/grid/MinpvProcessor.cpp
```diff
@@ -287,20 +288,14 @@ MinpvProcessor::process(const std::vector<double>& thickness,
                         // and sum of gaps is below threshold
                         const std::array<double, 8> cz_below = getCellZcorn(ii, jj, kk_iter, zcorn);
                         const std::array<double, 8> cz_above = getCellZcorn(ii, jj, k_above, zcorn);
-                        nnc_allowed = nnc_allowed && (computeGap(cz_above, cz_below) < max_gap);
+                        // top cell might not have been inspected for option 4 ALL before
+                        option4ALLZero = option4ALLZero || (!permz.empty() && permz[c_above] == 0.0) || multz(c_above) == 0.0;
+                        nnc_allowed = nnc_allowed && (computeGap(cz_above, cz_below) < max_gap) && (!pinchOption4ALL || !option4ALLZero) ;
```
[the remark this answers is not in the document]
blattms: Yes that is intended, because PERMZ and MULTZ only play a role for item 4 all. Only in this case the transmissibiities of the cells between the active ones are used. If one of those is zero (due to PERMZ or MULTZ), we omit the NNC.

blattms:
Thanks a lot for the comments. They should all be addressed now

opm/grid/cpgrid/processEclipseFormat.cpp
blattms: Changed the loop above to also increment in the for statement. Seemed safer.

tests/test_minpvprocessor.cpp
```diff
@@ -120,14 +120,22 @@ BOOST_AUTO_TEST_CASE(Pinch4ALL)
     BOOST_CHECK_EQUAL(minpv_result.nnc.size(), 0);
 
     permz = {2, 2, 1, 2.5}; // Not supported
```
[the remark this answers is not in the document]
blattms: corrected

opm/grid/cpgrid/processEclipseFormat.cpp
```diff
@@ -200,15 +209,110 @@ namespace cpgrid
...
+                    auto bottomIJK = ecl_grid.getIJK(cell2);
+                    std::vector<double> trans_between(bottomIJK[2]-topIJK[2]);
+                    std::vector<std::size_t> cells_between;
+                    cells_between.reserve(trans_between.size());
+
+                    // first we need to calculate transmissibilities from permeability and geometry
+                    auto trans = trans_between.begin();
+                    auto bottom_cell_info = ecl_grid.getCellAndBottomCenterNormal(cell1);
+
+                    // Wecalculate the transmissibilty tran for each intersection between the
```
atgeirr, line 230: Typo.

opm/grid/cpgrid/processEclipseFormat.cpp
```diff
@@ -200,15 +209,110 @@ namespace cpgrid
...
+                    // two active cells. For each intersection we need geometry information
+                    // (distance from cell center to face center, face area and normal) of the
+                    // the top (cell_top) and the bottom cell (cell_bottom).
+                    // we start at the top (top cell is actived) and iterate until the bottom
```
atgeirr, line 234: actived->active?

Also: the word active may be confused with the usual active cell concept, but I think it means something else here?

opm/grid/cpgrid/processEclipseFormat.cpp
```diff
@@ -200,15 +209,110 @@ namespace cpgrid
...
+                        else
+                            isZero = true;
+
+                    if (isZero)
+                        average = 0;
+                    else
+                        average = 1.0 / average;
+
+                    // Set nnc and transmissibility, last param indicates that this from pinch
+                    // It is needed to overwrite transmissibilties instead of adding to existing ones.
```
atgeirr, line 310: transmissibilties -> transmissibilities

atgeirr (approved):
A few minor typos in the new comments, but otherwise this is done, and the whole chain can be merged.

opm/grid/cpgrid/processEclipseFormat.cpp
blattms: Done.

To me it is the same. Those two cells are active due to the GRID and EDIT section. (MinPVProcessor might just have made more inactive).

opm/grid/cpgrid/processEclipseFormat.cpp
blattms: Done

opm/grid/cpgrid/processEclipseFormat.cpp
blattms: Done

PR #745, opened 2024-08-02
bump minimum dune version to 2.9

akva2:
Downstream of https://github.com/OPM/opm-common/pull/4157

PR #758, opened 2024-09-07
Prepare OPM for DUNE 2.10

blattms:
DUNE_VERSION_NEWER has been removed. Falling back to DUNE_VERSION_GTE which exists since at least DUNE 2.7

Use Dune::Communcation instead of removed Dune::CollectiveCommunication (there since at least DUNE 2.7)l

### DIFF
--- a/examples/finitevolume/finitevolume.cc
+++ b/examples/finitevolume/finitevolume.cc
@@ -36,7 +36,7 @@ template<class G>
 void timeloop(const G& grid, double tend)
 {
     // make a mapper for codim 0 entities in the leaf grid
-#if DUNE_VERSION_NEWER(DUNE_GEOMETRY, 2, 8)
+#if DUNE_VERSION_GTE(DUNE_GEOMETRY, 2, 8)
     Dune::MultipleCodimMultipleGeomTypeMapper<typename G::LeafGridView>  mapper(grid.leafGridView(), Dune::mcmgElementLayout());
 #else
     Dune::LeafMultipleCodimMultipleGeomTypeMapper<G> mapper(grid, Dune::mcmgElementLayout());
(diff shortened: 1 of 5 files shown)


blattms:
jenkins build this please

timokoch:
@blattms This has not made it into the 2024.10 release? So currently there is no version compatible with dune 2.10 right? Or can this is be backported?

bska:
> So currently there is no version compatible with dune 2.10 right?

Correct.

> Or can this is be backported?

In my capacity as the release manager for the 2024.10 OPM release: It cannot as that would trigger a **huge** round of testing of the release packages.  Dune 2.10 did not exist when we created the release branches for OPM 2024.10.  Besides, the proper solution to the problem is arguably #745 (ping @akva2) and the other PRs listed in OPM/opm-common#4157.

On another note, I'm **personally** rather hesitant to declare "support" for Dune 2.10 since the OPM project has a stated policy of supporting the two most recent Ubuntu LTS releases (22.04 and 24.04 at the time of writing) and Dune 2.10 does not exist on either of those.  In a sense OPM does not "need" Dune 2.10 until our 2026.04 release, although experience suggests we bring in some support for newer Dune releases earlier than that.  It's a double-edged sword, though, because such support should really be backed by build testing as well which has a non-zero cost.

timokoch:
The things replaced here have been deprecated for years now. The replacement is available since dune 2.7. I personally think independent of whether/when #745 will be completed/merged this PR is a good fix. Official support for dune 2.10 is not so important for me. The cleanup changes proposed here happen to be the _only_ things that I need to change to make opm-common+opm-grid compile with dune 2.10. That would have been great :) Maybe changing the title of the PR helps ;)?

I fully understand if backporting is too much work and I for sure didn't expect that it will end up in packages. Would be also useful on the release branch. Still useful on master. In any case, we'll have to provide a patch downstream. (We also do test this in the CI on our side and report if something doesn't work, that's how I found this PR. Personally, I think even if there is no official support from OPM side it's nice to accept small changes that do no harm and make things work smoother for other users). My comment was not indented to criticise that there is no official support yet, just to ask for clarification.

BTW Dune 2.10 (release branch) existed before the release branches for OPM 2024.10 were created (according to git history) but not important for this PR.

PR #759, opened 2024-09-12
fixed: zero-initialize array

akva2:
This caused issues for serialization tests with the debug iterator build. Random data cause integer overflow in the getSize method. Granted, those tests probably use classes in states they are not intended to be used, but it's also good to initialize.


akva2:
jenkins build this please

PR #761, opened 2024-09-16
Clean up a couple of friend declarations in CpGridData.hpp

aritorto:

### DIFF
--- a/opm/grid/cpgrid/CpGridData.cpp
+++ b/opm/grid/cpgrid/CpGridData.cpp
@@ -1710,6 +1710,31 @@ void CpGridData::computeCommunicationInterfaces([[maybe_unused]] int noExistingP
 #endif
 }
 
+int CpGridData::numFaces() const
+{
+    return face_to_cell_.size();
+}
+
+int CpGridData::numFaceVertices(int face_idx) const
+{
+    return face_to_point_[face_idx].size();
+}
+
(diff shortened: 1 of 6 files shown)


aritorto:
jenkins build this please

blattms (commented):
Unfortunately, the cleanup also seems to remove quite some part of the checks. Is that intended?

Normally refactoring should not change the amount of tests. Why is this different here? Might we miss problems due to this?

aritorto:
It's true that the clean up removed lines on the tests, however, those checked implementation details instead of features. For example, when we use "internally" -1 to denote that an entity does not verify certain property. Or when we check properties of children cells, it should be sufficient to use the hierarchic iterators instead of accessing via parent_to_children_cells_ . Sorry if "Clean up" on the title does not fully describe the PR intend

blattms:
I see. Fair enough. 

I see that we have the checks using child_to_parent as we are checking `Entity::father`.
But the checks the other way around (using parent_to_children) are removed. I think we should still have those, but refactored using HierarchicIterator. That should be possible, right?

I think moving more to the standard DUNE interface even in the tests is a very good change. Thanks a lot.

aritorto:
True, I haven't tried to use the iterators "in the other direction". That's a great idea. I'll draft this PR for now. Thanks for your feedback!

PR #762, opened 2024-09-25
Level  global cell for CpGrid with LGRs

aritorto:
This PR is relevant for including wells on LGRs (CpGrid). Potentially relevant for output files too. 

For CpGrid with LGRs, we determine the `global_cell_` values for refined level grids.
If an LGR has `NX * NY * NZ` dimension (amount of cells in each direction x,y,z respectively), then each refined cell can be associated with a (local) IJK and its (local) Cartesian index.
For a leaf grid view, we assign the `global_cell_` values of either the parent cell or the equivalent cell from level zero.

### DIFF
--- a/opm/grid/cpgrid/CpGrid.cpp
+++ b/opm/grid/cpgrid/CpGrid.cpp
@@ -633,6 +633,59 @@ const std::vector<int>& CpGrid::globalCell() const
     return currentData().back() -> global_cell_;
 }
 
+void CpGrid::computeGlobalCellLgr(const int& level, const std::array<int,3>& startIJK, std::vector<int>& global_cell_lgr)
+{
+    assert(level);
+    for (const auto& element : elements(levelGridView(level))) {
+        // Element belogns to an LGR, therefore has a father. Get IJK of the father in the level grid the father was born.
+        // For CARFIN, parent cells belong to level 0.
+        std::array<int,3> parentIJK = {0,0,0};
+        currentData()[element.father().level()]->getIJK(element.father().index(), parentIJK);
+        // Each parent cell has been refined in cells_per_dim[0]*cells_per_dim[1]*cells_per_dim[2] child cells.
+        // element has certain 'position' inside its parent cell that can be described with 'IJK' indices, let's denote them by ijk,
+        // where 0<= i < cells_per_dim[0], 0<= j < cells_per_dim[1], 0<= k < cells_per_dim[2].
+        const auto& cells_per_dim = currentData()[level]->cells_per_dim_;
+        //
+        // Refined cell (here 'element') has "index in parent cell": k*cells_per_dim[0]*cells_per_dim[1] + j*cells_per_dim[0] + i
+        // and it's stored in  cell_to_idxInParentCell_.
+        auto idx_in_parent_cell =  currentData()[level]-> cell_to_idxInParentCell_[element.index()]; // deep copy
+        // Find ijk.
+        std::array<int,3> childIJK = {0,0,0};
+
+        childIJK[0] = idx_in_parent_cell % cells_per_dim[0];
+        idx_in_parent_cell -= childIJK[0]; // k*cells_per_dim[0]*cells_per_dim[1] + j*cells_per_dim[0]
+        idx_in_parent_cell /= cells_per_dim[0]; // k*cells_per_dim[1] + j
+        childIJK[1] = idx_in_parent_cell % cells_per_dim[1];
+        idx_in_parent_cell -= childIJK[1]; // k*cells_per_dim[1]
+        childIJK[2] = idx_in_parent_cell /cells_per_dim[1];
+        // The corresponding lgrIJK can be computed as follows:
+        const std::array<int,3>& lgrIJK = { ( (parentIJK[0] - startIJK[0])*cells_per_dim[0] ) + childIJK[0],  // Shift parent index according to the startIJK of the LGR.
+                                            ( (parentIJK[1] - startIJK[1])*cells_per_dim[1] ) + childIJK[1],
+                                            ( (parentIJK[2] - startIJK[2])*cells_per_dim[2] ) + childIJK[2] };
+        // Dimensions of the "patch of cells" formed when providing startIJK and endIJK for an LGR
+        const auto& lgr_logical_cartesian_size = currentData()[level]->logical_cartesian_size_;
+        global_cell_lgr[element.index()] = (lgrIJK[2]*lgr_logical_cartesian_size[0]*lgr_logical_cartesian_size[1]) + (lgrIJK[1]*lgr_logical_cartesian_size[0]) + lgrIJK[0];
+    }
+}
+
+void CpGrid::computeGlobalCellLeafGridViewWithLgrs(std::vector<int>& global_cell_leaf)
+{
+    for (const auto& element: elements(leafGridView()))
+    {
+        // When refine via CpGrid::addLgrsUpdateGridView(/*...*/), level-grid to lookup global_cell_ is equal to level-zero-grid 
+        // In the context of allowed nested refinement, we lookup for the oldest ancestor, also belonging to level-zero-grid.
+        auto ancestor = element.getOrigin();
+        int origin_in_level_zero = ancestor.index();
+        while (ancestor.level()>0){
+            ancestor = ancestor.getOrigin();
+            origin_in_level_zero = ancestor.getOrigin().index();
+        }
+        assert(ancestor.level()==0);
+        assert(origin_in_level_zero < currentData().front()->size(0));
+        global_cell_leaf[element.index()] = currentData().front()-> global_cell_[origin_in_level_zero];
+    }
+}
+
 void CpGrid::getIJK(const int c, std::array<int,3>& ijk) const
 {
     current_view_data_->getIJK(c, ijk);
@@ -1809,7 +1862,6 @@ bool CpGrid::adapt(const std::vector<std::array<int,3>>& cells_per_dim_vec,
                                     cornerInMarkedElemWithEquivRefinedCorner,
                                     cells_per_dim_vec);
 
-    std::vector<int> adapted_global_cell(cell_count, 0);
     updateLeafGridViewGeometries( /* Leaf grid View Corners arguments */
                                   adapted_corners,
                                   corner_count,
@@ -1822,7 +1874,6 @@ bool CpGrid::adapt(const std::vector<std::array<int,3>>& cells_per_dim_vec,
                                   /* Leaf grid View Cells argumemts  */
                                   adapted_cells,
                                   adapted_cell_to_point,
-                                  adapted_global_cell,
                                   cell_count,
                                   adapted_cell_to_face,
                                   adapted_face_to_cell,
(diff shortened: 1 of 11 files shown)


aritorto:
jenkins build this please

aritorto:
jenkins build this please

aritorto:
jenkins build this please

aritorto:
@blattms maps for global cell from level grids and the leaf index set have been added. Could you take a look? 

@bska do you think these changes - apart from potential use for wells location on LGRs - are also relevant for including LGRs in output files?

blattms:
If we refine without LGRs, what size and values does globalCell have for the levels?

tests/cpgrid/adapt_cpgrid_test.cpp
```diff
@@ -207,6 +208,11 @@ void markAndAdapt_check(Dune::CpGrid& coarse_grid,
         const auto& grid_view = coarse_grid.leafGridView();
         Dune::MultipleCodimMultipleGeomTypeMapper<Dune::CpGrid::LeafGridView> adaptMapper(grid_view, Dune::mcmgElementLayout());
 
+        auto itMin = std::min_element((data.back() -> global_cell_).begin(),  (data.back()-> global_cell_).end());
+        auto itMax = std::max_element((data.back() -> global_cell_).begin(),  (data.back() -> global_cell_).end());
+        BOOST_CHECK_EQUAL( *itMin, 0);
+        BOOST_CHECK_EQUAL( *itMax, data.front()-> size(0) -1);
```
blattms, line 214: Maybe change this test to work also for grid with inactive cells?
You could check that each value is between zero and the maximum cartesian index possible.

tests/cpgrid/adapt_cpgrid_test.cpp
```diff
@@ -370,6 +372,14 @@ void markAndAdapt_check(Dune::CpGrid& coarse_grid,
             } // end-preAdaptElements-for-loop
         } // end-startingGridIdx==0
 
+        for (int level = 1; level < coarse_grid.maxLevel(); ++level)
+        {
+            auto itMinLevel = std::min_element((data[level] -> global_cell_).begin(),  (data[level] -> global_cell_).end());
+            auto itMaxLevel = std::max_element((data[level] -> global_cell_).begin(),  (data[level] -> global_cell_).end());
+            BOOST_CHECK_EQUAL( *itMinLevel, 0);
+            BOOST_CHECK_EQUAL( *itMaxLevel, data[level]-> size(0) -1);
```
blattms, line 380: Same here and below.  There could be inactive parent cells in the region to refine and hence missing kids (in the future?).

blattms (commented):
Looks good.
Just a few suggestions for better code reuse.

There are also comments on checks that won't hold if there are inactive cells. Maybe that is tested elsewhere and irrelevant?

aritorto:
> If we refine without LGRs, what size and values does globalCell have for the levels?
Good point. I didn't comment on that in the PR description. 

When we refine without LGRs, the size of global_cell_ coincides with the total amount of cells on the level grid/leaf grid view. For  refined level grids, the values are 0,1,2,..., total amount of cells in that refined level grid. 

For the leaf grid view, both refine-methods (with or without LGRs) do exactly the same: <leaf_grid_view>.global_cell_[ leaf cell index ] = global_cell_ value of the oldest ancestor from level zero. 

One idea to make these two refine-approaches compatible is to detect minimum and maximums ijk values of marked cells, and mimic what's done for LGRs in the without-LGRs refinement, with these "fake startIJK and endIJK"

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -633,6 +633,59 @@ const std::vector<int>& CpGrid::globalCell() const
...
+    assert(level);
+    for (const auto& element : elements(levelGridView(level))) {
+        // Element belogns to an LGR, therefore has a father. Get IJK of the father in the level grid the father was born.
+        // For CARFIN, parent cells belong to level 0.
+        std::array<int,3> parentIJK = {0,0,0};
+        currentData()[element.father().level()]->getIJK(element.father().index(), parentIJK);
+        // Each parent cell has been refined in cells_per_dim[0]*cells_per_dim[1]*cells_per_dim[2] child cells.
+        // element has certain 'position' inside its parent cell that can be described with 'IJK' indices, let's denote them by ijk,
+        // where 0<= i < cells_per_dim[0], 0<= j < cells_per_dim[1], 0<= k < cells_per_dim[2].
+        const auto& cells_per_dim = currentData()[level]->cells_per_dim_;
```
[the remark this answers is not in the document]
aritorto: I see that it might seem that we do not need it. Maybe there is a way to get rid of it as a data member. I'll think about it. 
In any case, we do use it, for example, in Entity::geometryInFather()[ here](https://github.com/OPM/opm-grid/blob/36069f06e5fe321c2dd161a7614dc95be5d05978/opm/grid/cpgrid/Entity.hpp#L535)

If we keep it, we can also change the name to cells_per_dim_per_parent_cell_. What do you think?

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -633,6 +633,59 @@ const std::vector<int>& CpGrid::globalCell() const
...
+    }
+}
+
+void CpGrid::computeGlobalCellLeafGridViewWithLgrs(std::vector<int>& global_cell_leaf)
+{
+    for (const auto& element: elements(leafGridView()))
+    {
+        // When refine via CpGrid::addLgrsUpdateGridView(/*...*/), level-grid to lookup global_cell_ is equal to level-zero-grid 
+        // In the context of allowed nested refinement, we lookup for the oldest ancestor, also belonging to level-zero-grid.
+        auto ancestor = element.getOrigin();
```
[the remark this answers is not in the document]
aritorto: I'm rewriting it a few lines below: ancestor = ancestor.getOrigin();

In any case, I tried the const auto& and get the error: passing ‘const Dune::cpgrid::Entity<0>’ as ‘this’ argument discards qualifiers [-fpermissive]
  674 |             ancestor = ancestor.getOrigin();
related to Entity operator=

So I would leave it as 'auto'

blattms:
> One idea to make these two refine-approaches compatible is to detect minimum and maximums ijk values of marked cells, and mimic what's done for LGRs in the without-LGRs refinement, with these "fake startIJK and endIJK"

I don't think we should make them compatible as they are not. cartesian indices on the level only makes sense for LGRs. In the long term we should probably throw if useers use them in adaptively refined grids. Let's postpone this, though,

opm/grid/cpgrid/CpGridData.cpp
```diff
@@ -1710,16 +1710,17 @@ void CpGridData::computeCommunicationInterfaces([[maybe_unused]] int noExistingP
...
 {
     // Refined cells in parent cell: k*cells_per_dim[0]*cells_per_dim[1] + j*cells_per_dim[0] + i
     std::array<int,3> ijk = {0,0,0};
-    ijk[0] = idxInParentCell % cells_per_dim[0];
-    idxInParentCell -= ijk[0]; // k*cells_per_dim[0]*cells_per_dim[1] + j*cells_per_dim[0]
-    idxInParentCell /= cells_per_dim[0]; // k*cells_per_dim[1] + j
-    ijk[1] = idxInParentCell % cells_per_dim[1];
-    idxInParentCell -= ijk[1]; // k*cells_per_dim[1]
-    ijk[2] = idxInParentCell /cells_per_dim[1];
+    int idx_copy = idx_in_parent_cell;
```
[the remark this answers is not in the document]
aritorto: Not needed

tests/cpgrid/adapt_cpgrid_test.cpp
aritorto: There are checks in opm-grid/tests/cpgrid/inactiveCells_lgr_test.cpp, e.g. [here](https://github.com/OPM/opm-grid/blob/36069f06e5fe321c2dd161a7614dc95be5d05978/tests/cpgrid/inactiveCell_lgr_test.cpp#L182). However, OPM/opm-grid#761 intends to remove all the implementation details, in particular the previous check. 

In this case, we can replace data.front()-> size(0)-1 by grid.logicalCartesianSize()[0]*...[1]*....*[2]

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -633,6 +633,59 @@ const std::vector<int>& CpGrid::globalCell() const
...
+        auto idx_in_parent_cell =  currentData()[level]-> cell_to_idxInParentCell_[element.index()]; // deep copy
+        // Find ijk.
+        std::array<int,3> childIJK = {0,0,0};
+
+        childIJK[0] = idx_in_parent_cell % cells_per_dim[0];
+        idx_in_parent_cell -= childIJK[0]; // k*cells_per_dim[0]*cells_per_dim[1] + j*cells_per_dim[0]
+        idx_in_parent_cell /= cells_per_dim[0]; // k*cells_per_dim[1] + j
+        childIJK[1] = idx_in_parent_cell % cells_per_dim[1];
+        idx_in_parent_cell -= childIJK[1]; // k*cells_per_dim[1]
+        childIJK[2] = idx_in_parent_cell /cells_per_dim[1];
```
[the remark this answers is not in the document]
aritorto: This is now in CpGridData::getInParentCellIJK()

aritorto:
> There are also comments on checks that won't hold if there are inactive cells. Maybe that is tested elsewhere and irrelevant?

Now there are a few lines in inactiveCell_test.cpp checking that the max value of global cell is below the max Cartesian Index for level grids, and for the leaf (which corresponds to the max Cartesian Index value of level zero grid)

tests/cpgrid/adapt_cpgrid_test.cpp
aritorto: Now the check is done with inequalities, meaning that the maximum value of global cell should be between 0 and the maximum Cartesian Index (the later might vary for each LGR, and it's equal to the maximum Cartesian Index from level zero for the leaf grid view)

aritorto:
jenkins build this please

aritorto:
@blattms thanks for your feedback! (which is now incorporated)

opm/grid/cpgrid/CpGrid.cpp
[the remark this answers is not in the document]
blattms: Yep. Indeed that was a bad idea of mine. We are overwriting it in the loop. Sorry for the noise.

opm/grid/cpgrid/CpGridData.hpp
```diff
@@ -309,6 +309,31 @@ class CpGridData
...
+    /// direction (x,y,and z) is given by std::array<int,3> cells_per_dim. In this case, CpGrid::adapt(/* with args */)
+    /// has been called.
+    /// - If refinement has been done via adapt() (without arguments) the default amount of children of each parent
+    /// cell in each direction is cells_per_dim = {2,2,2}.
+    ///
+    /// @param [in] idx_in_parent_cell      Integer between 0 and cells_per_dim[0]*cells_per_dim[1]*cells_per_dim[2]-1
+    ///                                     (total amount of children of a parent cell minus one since starting index is zero).
+    /// @param [in] cells_per_dim
+    /// @param [out] ijk                    In-parent-cell Cartesian index triplet
+    void getInParentCellIJK(int idx_in_parent_cell, const std::array<int,3>& cells_per_dim,  std::array<int,3>& ijk) const
```
blattms, line 324: Isn't this much more general than the name and the documentation suggests? It seems to convert a cartesian index to ijk and has no relation to refinement. You just happen to use it during refinement, currently. Why not simply name it getIJK?

In addition we could just return the ijk directly and only have two parameters.
Then you can do this in line 306: `ijk = getIK(c, logical_cartesian_size_);` and that method is done.

aritorto: It sounds doable. I kept both methods since CpGrid::getIJK  uses current_view_data_ ->getIJK (which is CpGridData::getIJK), and the integer argument "c" then it's used to get global_cell_[c] and then with this new integer, the "usual repeated deduction of ijk is done"

aritorto: It's probably your comment combined with ijk = getIJK(global_cell_[c], logical_cartesian_size_)

aritorto:
jenkins build this please

PR #763, opened 2024-09-25
fixed: do not add parallel test if executable was not built

akva2:
This is already if'd out in CMakeLists_files.cmake so we need to do the same here..


akva2:
jenkins build this please

akva2:
jenkins build this please

blattms:
thanks a lot. Which redhat version is this needed for?

akva2:
the one that will never die - 7.

PR #767, opened 2024-10-04
Report the number of generated pinch-out connections.

blattms:


blattms:
jenkins build this please

opm/grid/cpgrid/processEclipseFormat.cpp
```diff
@@ -312,6 +314,9 @@ namespace cpgrid
                 }
             }
 
+            Opm::OpmLog::info(fmt::format("{} pinch-out connections generated",
+                                          nnc_cells[PinchNNC].size()));
```
bska, line 318: So this will print

> 0 pinch-out connections generated

if there are no additional connections.  That's okay, if a little redundant.  Worse, however, is that it will print

> 1 pinch-out connection**S** generated

i.e., a plural, for the case of a single pinch-out connection.  That's not ideal and there's no reason to do that here because we can easily query the number of connections emit a message in concord in this case.

bska (commented):
I do wish you'd stop pushing your branches directly to the main OPM repository.

Other than that, this is okay save for a smaller grammar issue.

blattms:
jenkins build this please

opm/grid/cpgrid/processEclipseFormat.cpp
```diff
@@ -312,6 +314,10 @@ namespace cpgrid
                 }
             }
 
+            Opm::OpmLog::info(fmt::format("{} pinch-out connection{} generated",
+                                          nnc_cells[PinchNNC].size(),
+                                          nnc_cells[PinchNNC].size()>1?"s":""));
```
bska, line 319: Tiny nit: The condition needs to be `size() != 1` here.

blattms:
Thanks a lot for catching that. Always amazing what I mange to do wrong in the tiniest PRs ever :sweat_smile:

blattms:
jenkins build this please

bska (approved):

PR #768, opened 2024-10-07
Download External FMT Library if Not Found

bska:
Unless the FMT library is installed in a central location, our configure time checks won't find it and we'll get a build failure.


bska:
jenkins build this please

blattms:
I rather change my PR. I think this is overkill

PR #769, opened 2024-10-07
[fix] Omit fmt::format usage.

blattms:
We do not require that package for building.
In the long run we should make the one in opm-common available, though.

For the time being this is better than #768 

I apologize for the fallout from #767

### DIFF
--- a/opm/grid/cpgrid/processEclipseFormat.cpp
+++ b/opm/grid/cpgrid/processEclipseFormat.cpp
@@ -58,6 +58,7 @@
 #include <iostream>
 #include <initializer_list>
 #include <set>
+#include <string>
 #include <utility>
 
 namespace Dune
--- a/opm/grid/cpgrid/processEclipseFormat.cpp
+++ b/opm/grid/cpgrid/processEclipseFormat.cpp
@@ -53,8 +53,6 @@
 #include <opm/grid/RepairZCORN.hpp>
 #include <opm/grid/utility/StopWatch.hpp>
 
-#include <fmt/format.h>
-
 #include <cstddef>
 #include <fstream>
 #include <iostream>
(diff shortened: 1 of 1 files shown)


blattms:
jenkins build this please

opm/grid/cpgrid/processEclipseFormat.cpp
```diff
@@ -314,9 +312,8 @@ namespace cpgrid
                 }
             }
 
-            Opm::OpmLog::info(fmt::format("{} pinch-out connection{} generated",
-                                          nnc_cells[PinchNNC].size(),
-                                          (nnc_cells[PinchNNC].size() != 1)? "s" : ""));
+            auto suffix = (nnc_cells[PinchNNC].size() != 1)? "s" : "";
```
bska, line 315: This will make `suffix` be a `const char*` whence the sub-expression
```C++
" pinch-out connection"+suffix+" generated"
```
is strictly speaking undefined. That, in turn, means that we're a little bit dependent on how the compiler chooses to parse the overall expression
```C++
to_string(...) + "..." + suffix + "..."
```
for whether or not we get a build failure.  The easiest way to guarantee that the expression is well defined is to force `suffix` to be a `std::string` through an initialisation expression of the form
```C++
const auto suffix = std::string { (nnc_cells[PinchNNC].size() != 1)? "s" : "" };
```
By the way, you should `#include <string>` to guarantee that we have a declaration of `std::to_string()` in scope.

bska (commented):
Thanks.  Other than one potential parsing issue, this looks good to me.

opm/grid/cpgrid/processEclipseFormat.cpp
blattms: All comments addressed.
My comment on the compiler dependency  (I am probably missing something...):

AFAIK the addition like it was seems well-defined, because the very first argument in the addition is a string, addition is left-to-right associative, see https://en.cppreference.com/w/cpp/language/operator_precedence and the definitions of operator+ for const char* and basic_string is there for both directions, see https://en.cppreference.com/w/cpp/string/basic_string/operator%2B .  Hence there is always a string in each operation.

IMHO, we would need to do something like this to get into trouble:
```
std::to_string(nnc_cells.size()) + (" pinch-out connection" + suffix) + " generated";
```

blattms:
jenkins build this please

bska (approved):

PR #770, opened 2024-10-09
From static member to a local static

aritorto:
Only instantiate in_father_reference_elem_corner_indices if geometryInfather() is called. Declare the variable (local) static extends its lifetime beyond geometryInFather().  This is needed since the returned value of geometryInFather() points at in_father_reference_elem_corner_indices.  

### DIFF
--- a/opm/grid/cpgrid/Entity.hpp
+++ b/opm/grid/cpgrid/Entity.hpp
@@ -528,7 +528,7 @@ Dune::cpgrid::Geometry<3,3> Dune::cpgrid::Entity<codim>::geometryInFather() cons
     // Indices of corners in entity's geometry in father reference element.
     static constexpr std::array<int,8> in_father_reference_elem_corner_indices = {0,1,2,3,4,5,6,7};
     // 'static': The returned object Geometry<3,3> stores a pointer to in_father_reference_elem_corner_indices. Therefore,
-    // this variable is declated static to prolongate its lifetime beyond this function (static storage duration).
+    // this variable is declared static to prolongate its lifetime beyond this function (static storage duration).
 
     auto idx_in_parent_cell = pgrid_ -> cell_to_idxInParentCell_[this->index()];
     if (idx_in_parent_cell !=-1) {
--- a/opm/grid/cpgrid/Entity.hpp
+++ b/opm/grid/cpgrid/Entity.hpp
@@ -524,11 +524,12 @@ Dune::cpgrid::Geometry<3,3> Dune::cpgrid::Entity<codim>::geometryInFather() cons
     if (!(this->hasFather())){
         OPM_THROW(std::logic_error, "Entity has no father.");
     }
-    // static    to not need any extra storage per Enitity. One object used for all instances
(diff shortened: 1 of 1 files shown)


aritorto:
jenkins build this please

opm/grid/cpgrid/Entity.hpp
```diff
@@ -529,6 +524,11 @@ Dune::cpgrid::Geometry<3,3> Dune::cpgrid::Entity<codim>::geometryInFather() cons
     if (!(this->hasFather())){
         OPM_THROW(std::logic_error, "Entity has no father.");
     }
+    // static    to not need any extra storage per Enitity. One object used for all instances
+    // constexpr to allow for in-class instantiation
+    /// \brief Indices of corners in entity's geometry in father reference element, for the Geometry returned by geometryInFather
+    static constexpr std::array<int,8> in_father_reference_elem_corner_indices_ = {0,1,2,3,4,5,6,7};
```
blattms, line 530: Nice.

The only thing that I am wondering about is whether we really need the static. Should we maybe drop it?

This is a local variable now. Hence the storage argument is a bit different. For a local variable, that is not static, the storage is freed after the function call. So storage is only used during this function. By making it static, the variable (and storage) is allocated and initialized at program start and stays until the end.
One reason to make it static would be if the initialization is costly. This does not hold here.

aritorto: Thanks for the explanation. Dropping it creates a warning (storing the address of a local variable in the return value of geometryInFather()), so I would keep it. 

I fixed the documentation too

blattms: I see. Did not look how it is used. Well the function name should have told me. Please add documentation that the Geometry that we return stores a pointer to this and that therefore we need to make sure that the lifetime goes beyond this function.

aritorto: Comment has been added, thanks!

aritorto:
jenkins build this please

opm/grid/cpgrid/Entity.hpp
```diff
@@ -529,6 +524,12 @@ Dune::cpgrid::Geometry<3,3> Dune::cpgrid::Entity<codim>::geometryInFather() cons
     if (!(this->hasFather())){
         OPM_THROW(std::logic_error, "Entity has no father.");
     }
+
+    // Indices of corners in entity's geometry in father reference element.
+    static constexpr std::array<int,8> in_father_reference_elem_corner_indices = {0,1,2,3,4,5,6,7};
+    // 'static': The returned object Geometry<3,3> stores a pointer to in_father_reference_elem_corner_indices. Therefore,
+    // this variable is declated static to prolongate its lifetime beyond this function (static storage duration).
```
blattms, line 531: declated -> declared.

blattms (approved):
Looks good. Just a tiny typo to fix.

aritorto:
Typo is fixed!

aritorto:
jenkins build this please

PR #775, opened 2024-10-15
Feature/graph of grid

akva2:
Graph representation of the grid.
Purpose: Loadbalancing of the grid with wells. Each well is represented with a single vertex in the graph repr. of the grid, therefore no well can be spread over multiple processes.

What is implemented so far:

    GraphofGrid -class that stores the representation. Needs a grid (currently CpGrid) to build the graph. Has functions for searching for cells by their global IDs, adding wells, and vertex contraction.
        Edge weights (transmissibilities) currently start equal to 1, and are possibly added up when contracted vertices have a common neighbor
    wrappers for Zoltan callback functions (not tested with Zoltan yet though)
    unit test for the above

This is #774 reopened due to a github glitch.

PR #774, opened 2024-10-15
Feature/graph of grid

michal-toth:
Graph representation of the grid.
Purpose: Loadbalancing of the grid with wells. Each well is represented with a single vertex in the graph repr. of the grid, therefore no well can be spread over multiple processes.

What is implemented so far:
- GraphofGrid -class that stores the representation. Needs a grid (currently CpGrid) to build the graph. Has functions for searching for cells by their global IDs, adding wells, and vertex contraction.
- - Edge weights (transmissibilities) currently start equal to 1, and are possibly added up when contracted vertices have a common neighbor
- wrappers for Zoltan callback functions (not tested with Zoltan yet though)
- unit test for the above

PR #776, opened 2024-10-16
Feature: graph representation of the grid

michal-toth:
This is my second attempt to create a pull request after the first one got bugged (see #774 ). akva2 doubled it already (#775), so this one is to determine if I can create a working PR. The original description follows:

Graph representation of the grid.
Purpose: Loadbalancing of the grid with wells. Each well is represented with a single vertex in the graph repr. of the grid, therefore no well can be spread over multiple processes.

What is implemented so far:

GraphofGrid -class that stores the representation. Needs a grid (currently CpGrid) to build the graph. Has functions for searching for cells by their global IDs, adding wells, and vertex contraction.
    Edge weights (transmissibilities) currently start equal to 1, and are possibly added up when contracted vertices have a common neighbor
wrappers for Zoltan callback functions (not tested with Zoltan yet though)
unit test for the above


blattms:
jenkins build this please

blattms:
jenkins build this please

blattms (changes requested):
Thanks a lot.
I have not checked the tests very diligently yet I might come back to that later. Just a few comments there for now.

Please note that i might always miss something or misunderstand things.

opm/grid/GraphOfGrid.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+  // add up vertex weights
+  graph[gID1].weight += graph[gID2].weight;
+
+  // Merge the list of neighbors,
+  // for common neighbors add up edge weights.
+  // Remove the edge between gID1, gID2.
+  auto& v1e = graph[gID1].edges;
+  v1e.erase(gID2);
+  for (const auto& edge : graph[gID2].edges)
```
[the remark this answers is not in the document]
michal-toth: Indeed, we are changing all edges that connected gID2's neighbors to gID2. Either the target is changed to gID1, or, if gID1 is already a neighbor, the transmissibilities are added up. Edges in gID1 are changed accordingly. This is done in the for-cycle at the end of the section you highlighted.

opm/grid/GraphOfGrid.hpp
```diff
@@ -0,0 +1,162 @@
...
+
+  /// \brief List of neighbors for given vertex
+  EdgeList edgeList(int gID) const
+  {
+    // get iterator to the vertex or the well containing it
+    auto pgID = find(gID);
+    if (pgID==graph.end())
+    {
+      return EdgeList{};
+    }
```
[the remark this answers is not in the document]
michal-toth: Reworked. The edgeList now returns const reference or throws if the ID is not in the graph or wells.

opm/grid/GraphOfGrid.cpp
```diff
@@ -0,0 +1,167 @@
...
+      {
+        if (w->find(gID)!=w->end())
+        {
+          // gID is in another well => remap it and join wells
+          if (wID==gID)
+            wID = *(w->begin());
+          gID = *(w->begin());
+          newWell.insert(w->begin(),w->end());
+          wells.erase(w);
+          break; // wells are assumed disjoint, each gID has max 1 match
```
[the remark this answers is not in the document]
michal-toth: What do you mean by "this"? If it is whether the wells are disjoint, then catering for the case when they would not be would require a lot of rework and several inefficiencies.
We do not expect users to provide disjoint wells, wells with a common cell are internally merged into one. Adding wells checks for intersections to ensure this . (By default, it is possible to skip the check to improve speed but onus is on the user.) 
Having disjoint wells (internally) is necessary for having uniquely defined graph. They are simply a set of cells that got contracted into one vertex. (We could consider renaming them but their intended use case is for wells.)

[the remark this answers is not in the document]
blattms: Thanks. So this comment seems to be about storage of the wells internally. Maybe we can make this more clear in the comment.

When I read it, I interpreted it as: "There cannot be two wells that have a connection to the same cell in the simulaton"

blattms (commented):
I am sending this now to reduce the feedback delay. I am not finished yet, though. Still need to look at the import / export lists.

tests/test_graphofgrid.cpp
```diff
@@ -0,0 +1,290 @@
...
+		Looking up vertex not present in GraphofGrid gives dummy with weight 0.");
+
+    auto v5e = gog.getVertex(5).edges;
+	BOOST_REQUIRE(v5e==gog.edgeList(5));
+	BOOST_REQUIRE(v5e==gog.edgeList(6)); // 5 and 6 have the same neighbors (1, 2 got merged)
+	BOOST_REQUIRE(v5e!=gog.edgeList(7));
+
+}
+
+#include <opm/grid/GraphOfGridWrappers.hpp>
```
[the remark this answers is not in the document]
michal-toth: The main reason was to hide Wrappers from the first few tests. I was considering splitting the file into two after reading your first comment, but it would cut it into very small pieces. I will move it to the top.

tests/test_graphofgrid.cpp
```diff
@@ -0,0 +1,392 @@
...
+    BOOST_REQUIRE(gog.edgeList(3).size()==2);
+    BOOST_REQUIRE(gog.edgeList(3).at(0)==2);
+    // contracting vertices removes higher ID from the graph
+    // (when well is added, IDs removed from the graph are stored in the well)
+    BOOST_REQUIRE_THROW(gog.getVertex(1),std::logic_error);
+
+    auto v5e = gog.getVertex(5).edges;
+    BOOST_REQUIRE(v5e==gog.edgeList(5));
+    BOOST_REQUIRE(v5e==gog.edgeList(6)); // 5 and 6 have the same neighbors (1, 2 got merged)
+    BOOST_REQUIRE(v5e!=gog.edgeList(7));
```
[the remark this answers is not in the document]
michal-toth: I considered it to be a little excessive here. However, the test IntersectingWells checks all edges at one point.

blattms:
jenkins build this please

blattms:

opm/grid/GraphOfGridWrappers.hpp
```diff
@@ -0,0 +1,270 @@
...
+/// (like process number and ownership) of their representative cell (well ID).
+template<typename TheTuple>
+void extendImportExportList (const GraphOfGrid<Dune::CpGrid>& gog,
+                                       std::vector<TheTuple>& cellList)
+{
+    // using TheTuple = std::tuple<int,int,char>; or std::tuple<int,int,char,int>
+    using CellList = std::vector<TheTuple>;
+    // make a list of wells for easy identification. Contains ID, begin, end
+    using iter = std::set<int>::const_iterator;
+    std::list<std::tuple<int,iter,iter>> wellList;
```
blattms, line 222: Please add a comment here why having this list is beneficial: Because we can make the search space smaller below.

We do a (linear) search in this container for the first cell below. Maybe we should use a vector sorted by the first cell? Then we could do  a binary search with std::lower_bound below.

We cannot shrink the container though, but maybe it is still more performant?

Could also be premature optimization (and turn out slower or not that much faster).IF we do such a change we should actually benchmark both on some cases to see which is better. It is not like we have thousands of wells...

So maybe this is for the future.

opm/grid/GraphOfGridWrappers.hpp
```diff
@@ -0,0 +1,270 @@
...
+        wellList.push_front(std::make_tuple(*well.begin(),well.begin(),well.end()));
+    }
+
+    CellList addToList;
+    // iterate once through the original cellList
+    for (const auto& cell : cellList)
+    {
+        // if a cell is in any well, add cells of the well to cellList
+        // and remove the well from the wellList to quicken the search
+        for (auto wID=wellList.begin(); wID!=wellList.end(); )
```
blattms, line 234: It is hard to see and understand what is what below. Maybe using  a structured binding might help?
```
const auto& [ firstCell, begin, end] = *wID;

File: opm/grid/GraphOfGridWrappers.hpp [hunk 8]
```diff
@@ -0,0 +1,270 @@
...
+    using iter = std::set<int>::const_iterator;
+    std::list<std::tuple<int,iter,iter>> wellList;
+    for (const auto& well : gog.getWells())
+    {
+        wellList.push_front(std::make_tuple(*well.begin(),well.begin(),well.end()));
+    }
+
+    CellList addToList;
+    // iterate once through the original cellList
+    for (const auto& cell : cellList)
```
blattms, line 230: I don't have a better name, but when I read cell I think this an integer.

opm/grid/GraphOfGridWrappers.hpp
```diff
@@ -0,0 +1,270 @@
...
+            }
+        }
+        if (wellList.empty())
+        {
+            break;
+        }
+    }
+    std::sort(addToList.begin(),addToList.end(),[](const auto& a, const auto& b){return std::get<0>(a)<std::get<0>(b);});
+    int origSize = cellList.size();
+    int totsize = origSize+addToList.size();
```
blattms, line 262: Use auto to get 64bit?

opm/grid/GraphOfGridWrappers.hpp
```diff
@@ -0,0 +1,270 @@
...
+    cellList.reserve(totsize);
```
blattms, line 263: I think we do know who many entries will be added in the function already at the beginning, right?
In that case we might be able to omit the extra container addToList and push_back to cellList directly.

I might be missing something here, though.

blattms (commented):
Nice and I think I am finished.

Last part is mostly minor things and stylistic comments.

blattms:
There are some build failures (due to me, sorry):
```
/var/lib/jenkins/workspace/opm-grid-PR-builder/tests/test_graphofgrid.cpp:358:38: error: ‘AttributeSet’ has not been declared
  358 |     BOOST_CHECK(std::get<2>(imp[5])==AttributeSet::copy);
      |                                      ^~~~~~~~~~~~
```
I think you need to explicitly include dune/istl/owneroverlapcopy.hh to fix this.

opm/grid/GraphOfGridWrappers.hpp
michal-toth: I changed the container from std::list to std::unordered_map. Removing entries no longer improves search speed, but allows us to leave the loop sooner when the map gets empty.

opm/grid/GraphOfGridWrappers.hpp
michal-toth: I also ran out of ideas. I renamed it to cellProperties.

opm/grid/GraphOfGridWrappers.hpp
michal-toth: The number of cells to be added to the cellList depends on which wells are in the list. Unless all wells are always in the list we do not know the final size beforehand.

michal-toth:
> There are some build failures (due to me, sorry):
> 
> ```
> /var/lib/jenkins/workspace/opm-grid-PR-builder/tests/test_graphofgrid.cpp:358:38: error: ‘AttributeSet’ has not been declared
>   358 |     BOOST_CHECK(std::get<2>(imp[5])==AttributeSet::copy);
>       |                                      ^~~~~~~~~~~~
> ```
> 
> I think you need to explicitly include dune/istl/owneroverlapcopy.hh to fix this.

I will add the include. However, I can not reproduce this error so I can not check if it helped.

blattms:

blattms (approved):

blattms:
:sunglasses: Thanks  a lot for making this build in serial. In it goes.

tests/test_graphofgrid.cpp
```diff
@@ -0,0 +1,517 @@
...
+    std::array<double,3> size{1.,1.,1.};
+    grid.createCartesian(dims,size);
+    Opm::GraphOfGrid gog(grid);
+
+    int err;
+    int nVer = getGraphOfGridNumVertices(&gog,&err);
+    BOOST_REQUIRE(err==ZOLTAN_OK);
+    BOOST_REQUIRE(nVer == 60);
+
+    int gIDs[nVer], lIDs[0];
```
atgeirr, line 121: I get many warnings now from this and other places in the code: arrays with variable length is not C++, it is a GNU C extension, so we should not use it. Use vector instead:
```
std::vector<int> gIDs(nVer);
```
And then if you need to pass the pointer to the start of the data use `gIDs.data()`.

However, the `lIDs[0]` has me confused: a zero length array? Why? In general, it seems to me this should have the same size as the other array. For getGraphOfGridVerticesList() specifically, I think it does not look at that array at all (it is flagged as maybe_unused), then passing `nullptr` here is appropriate (although I think the arrays, sizes and content should be similar to how it will be used in real code if possible).

@michal-toth can you address these issues?

michal-toth: Thanks, I was not aware of the std::vector's .data() function. I cleared it out, hopefully I found all instances.

PR #778, opened 2024-10-16
changed: avoid use of deprecated Factorial struct

akva2:
Quells a warning with clang.

"Use function factorial instead! Will be removed after Dune 2.9"


akva2:
jenkins build this please

atgeirr (approved):

PR #779, opened 2024-10-16
[clean-up] Avoid a few copies and remove/silence unused variables

aritorto:


aritorto:
jenkins build this please

PR #783, opened 2024-10-23
Generalize assignment of global ids for refined cells and points.

aritorto:
This PR generalizes the assignment of global ids for refined level grid cells and points, paving the way for supporting  distributed LGRs.

Note that the restriction requiring LGR cells to be fully interior is still in place here. Removing this restriction would require an additional communication step, along with defining parallel indices and a level cell index set for each level grid. This additional work is addressed in OPM/opm-grid#785.

### DIFF
--- a/opm/grid/cpgrid/CpGrid.cpp
+++ b/opm/grid/cpgrid/CpGrid.cpp
@@ -2141,7 +2141,6 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
         getIJK(element.index(), ijk);
         for (std::size_t level = 0; level < startIJK_vec.size(); ++level) {
             bool belongsToLevel = true;
-            int marked_elem_level_count = 0;
             for (int c = 0; c < 3; ++c) {
                 belongsToLevel = belongsToLevel && ( (ijk[c] >= startIJK_vec[level][c]) && (ijk[c] < endIJK_vec[level][c]) );
                 if (!belongsToLevel)
@@ -2165,7 +2164,6 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
                     }
                     this-> mark(1, element);
                     assignRefinedLevel[element.index()] = level+1; // shifted since starting grid is level 0, and refined grids levels are >= 1.
-                    ++marked_elem_level_count;
                     lgr_with_at_least_one_active_cell[level] = 1;
                 }
             } // end-if-belongsToLevel
@@ -2310,6 +2308,7 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
             }
             // For the general case where the LGRs might be also distributed, a communication step is needed to assign global ids
             // for overlap cells and points.
+            /** TODO: Set up the parallel index set correctly. */
 
             // Global id set for each (refined) level grid.
             if(lgr_with_at_least_one_active_cell[level-1]>0) {
@@ -2322,7 +2321,7 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
         ////////////////////////////////
 
         // Global id for the cells in leaf grid view
-        std::vector<int> leafCellIds(current_data_->back()->size(0), 0);
+        std::vector<int> leafCellIds(current_data_->back()->size(0));
         for(const auto& element: elements(leafGridView())){
             // Notice that for level zero cells the global_id_set_ is given, for refined level grids was defined
             // under the assumption of each lgr being fully contained in the interior of a process.
--- a/opm/grid/cpgrid/CpGrid.cpp
+++ b/opm/grid/cpgrid/CpGrid.cpp
@@ -2241,12 +2241,9 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
         std::vector<std::size_t> point_ids_needed_by_proc(comm().size());
         std::size_t local_point_ids_needed = 0;
         for (std::size_t level = 1; level < cells_per_dim_vec.size()+1; ++level){
-            // Compute the partition type for point, if the level grid is not empty, i.e., it contains
-            // at least one active cell.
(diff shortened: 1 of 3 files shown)


aritorto:
jenkins build this please

aritorto:
jenkins build this please

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -2206,159 +2204,122 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
...
-                            ++local_owned_points_per_level[level];
+        }
+        comm().allgather(&local_cell_ids_needed, 1, cell_ids_needed_by_proc.data());
+
+        std::vector<std::size_t> point_ids_needed_by_proc(comm().size());
+        std::size_t local_point_ids_needed = 0;
+        for (std::size_t level = 1; level < cells_per_dim_vec.size()+1; ++level){
+            // Compute the partition type for point
+            if(active_cell_count_per_level[level-1]>0) {
+                (*current_data_)[level]->computePointPartitionType();
```
blattms, line 2243: Maybe put comment here that a TODO is still to set up the parallel index set correctly?
Currently this call will mark all elements as owner. I guess that is probably expected as we don't allow LGRs touching the faces of overlap cells, yet?

tests/cpgrid/addLgrsOnDistributedGrid_test.cpp
```diff
@@ -588,3 +597,222 @@ BOOST_AUTO_TEST_CASE(globalRefine2)
...
+                                                     {2,3,6,7,10,11,18,22,23},
+                                                     {14,15,19,26,27,30,31,34,35} };
+    for (int rank = 0; rank < 4; ++rank) {
+        for (const auto& elemIdx : cells_per_rank[rank]) {
+            parts[elemIdx] = rank;
+        }
+    }
+    if(grid.comm().size()>1)
+    {
+        grid.loadBalance(parts);
```
blattms, line 627: Not sure: Do we need to check whether grid.comm().size() >= 4 or is that done in loadBalance?

blattms (commented):
Thanks. I think this is a good improvement. 

We still do not allow LGRS to touch faces of overlap cells and the approach does not work for adaptive refinement. That is fine as we probably take it step by step.

Just a few minor comments.

blattms:
Better safe than sorry:

aritorto:
> We still do not allow LGRS to touch faces of overlap cells and the approach does not work for adaptive refinement. That is fine as we probably take it step by step.

Thanks for the feedback! I realize the PR description was quite brief—sorry about that!

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -2163,7 +2163,7 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
                     this-> mark(1, element);
                     assignRefinedLevel[element.index()] = level+1; // shifted since starting grid is level 0, and refined grids levels are >= 1.
                     ++marked_elem_level_count;
-                    lgrs_with_at_least_one_active_cell[level] = marked_elem_level_count;
+                    active_cell_count_per_level[level] = marked_elem_level_count;
```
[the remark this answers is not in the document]
aritorto: Unfortunately, the name isn’t very clear. The variable active_cell_count_per_level[in certain level] stores the total number of active coarse cells that will be refined, with their child cells belonging to that specific level. I completely agree that it can be confusing, sorry for that!

Currently, we only support values of 0 and 1 for potential marks, which indicate 'do nothing to this element' and 'refine this element,' respectively. The variable marked_elem_level_count increases each time we encounter an element marked for refinement (line 2265 in the PR). At the end of the loops, this count represents the total number of marked elements to be refined, potentially across different level grids.

opm/grid/cpgrid/CpGrid.cpp
aritorto: I will add a TODO comment, but it's sad to realize that it’s not working as I expected. I initially thought that computePointPartition() would categorize points as interior, border, front, or overlap based on [CpGridData::computePointPartition() ](https://github.com/OPM/opm-grid/blob/master/opm/grid/cpgrid/CpGridData.cpp#L1631)

tests/cpgrid/addLgrsOnDistributedGrid_test.cpp
aritorto: It seems that I should check it, it's not done in loadBalance.

aritorto:
Thanks again for your feedback, all comments have been addressed!

aritorto:

opm/grid/cpgrid/CpGrid.cpp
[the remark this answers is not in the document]
blattms: Seems like `marked_elem_level_count` is never really used now. Please remove it.

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -2206,159 +2207,122 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
...
         } // end-for-loop-level
-        assert(globalIdCell <= max_globalId_levelZero + global_refined_cell_count +1);
-        assert(globalIdPoint <= max_globalId_levelZero + global_refined_cell_count + almost_global_refined_point_count+1);
-        
 
         ////////////////////////////////
 
         // Global id for the cells in leaf grid view
-        std::vector<int> leafCellIds(current_data_->back()->size(0));
+        std::vector<int> leafCellIds(current_data_->back()->size(0), 0);
```
blattms, line 2325: I probably should have commented on all these. Here is another one with an access parameter that is the default value, anyway. Maybe remove it, too.

opm/grid/cpgrid/CpGrid.cpp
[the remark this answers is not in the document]
aritorto: Done, thanks!

opm/grid/cpgrid/CpGrid.cpp
aritorto: I could have also checked, removed!

aritorto:

PR #784, opened 2024-10-24
Data handle for parent-to-children-global ids

aritorto:
This PR introduces a structure to manage the global IDs of parent-to-children cells, paving the way for distributing Local Grid Refinements (LGRs) across different processes. This structure ensures the correct assignment of global cell IDs in refined grids, gathering already-assigned IDs for interior cells and distributing them for overlap cells.

The next step will focus on implementing a data handle for points/vertices (Entity<3>).

It contains OPM/opm-grid#783

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -199,6 +199,7 @@ list (APPEND PUBLIC_HEADER_FILES
   opm/grid/LookUpCellCentroid.hh
   opm/grid/LookUpData.hh
   opm/grid/cpgrid/OrientedEntityTable.hpp
+  opm/grid/cpgrid/ParentToChildrenCellGlobalIdHandle.hpp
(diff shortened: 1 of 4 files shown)


aritorto:
jenkins build this please

blattms:
I have not really looked at this in any detail, yet.

This has only one commit, but includes another PR. Is this intentional? 

I think the changes related to the subject of this PR should be in their own commit. 
There should also be some kind of test for the DataHandle.

aritorto:
> This has only one commit, but includes another PR. Is this intentional?

Unfortunately, I messed things up while I fixed-up some commits. It was not intentional. I'll split the change and add some tests for the data handle. Turn it into draft now.

aritorto:
Replaced and improved by OPM/opm-grid#785

PR #785, opened 2024-10-29
Support distributed LGRs for CpGrid

aritorto:
This PR introduces a structure to manage the global IDs of parent-to-children cells, allowing support of distributing Local Grid Refinements (LGRs) across different processes, for CpGrid. This structure (ParentToChildrenCellGlobalIdsHandle) ensures the correct assignment of global cell IDs in refined grids, gathering already-assigned IDs for interior cells and distributing them for overlap cells.

### DIFF
--- a/opm/grid/cpgrid/CpGrid.cpp
+++ b/opm/grid/cpgrid/CpGrid.cpp
@@ -2197,14 +2197,18 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
         // Refined level grid cells:
         //    1. Inherit their partition type from their parent cell (i.e., element.father().partitionType()).
         //    2. Assign global ids only for interior cells.
+        //    3. Communicate the already assigned cell global ids from interior to overlap refined cells.
         // Refined level grid points/vertices:
-        //    1. Challenge: <refined level grid>.computePointPartitionType(). This requires face and cell partition type to be defined. For
-        //       a grid with LGRs, this can be done on the fly. Under the assumption of fully interior LGRs, all refined level grid
-        //       points are interior.
-        //    2. Previous step allows us to assign uniquily global ids to non-overlap points.
+        // There are 4 partition types: interior, border, front, overlap. This classification requires that both
+        // cell and face partition types are already defined, not available yet for refined level grids.
+        //    1. Assign for all partition type points a 'candidate of global id' (unique in each process).
+        //       Except the points that coincide with a point from level zero.
+        //    2. To make point ids globally unique, re-write the values for points that are corners of overlap
+        //       refined cells, via communication.
         // Under the assumption of LGRs fully-interior, no communication is needed. In the general case, communication will be used
         // to populate overlap cell/point global ids on the refined level grids.
 
+        // Predict how many new cell ids per process are needed.
         std::vector<std::size_t> cell_ids_needed_by_proc(comm().size());
         std::size_t local_cell_ids_needed = 0;
         for ( const auto& element : elements( levelGridView(0), Dune::Partitions::interior) ) {
@@ -2212,11 +2216,15 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
             bool hasBeenMarked = currentData().front()->getMark(element) == 1;
             if ( hasBeenMarked ) {
                 const auto& level = assignRefinedLevel[element.index()];
+                // Shift level (to level -1) since cells_per_dim_vec stores number of subdivisions in each direction (xyz)
+                // per parent cell, per level, starting from level 1, ..., maxLevel.
                 local_cell_ids_needed += cells_per_dim_vec[level-1][0]*cells_per_dim_vec[level-1][1]*cells_per_dim_vec[level-1][2];
             }
         }
         comm().allgather(&local_cell_ids_needed, 1, cell_ids_needed_by_proc.data());
 
+        // Overestimate ('predict') how many new point ids per process are needed.
+        // Assign for all partition type points a 'candidate of global id' (unique in each process).
         std::vector<std::size_t> point_ids_needed_by_proc(comm().size());
         std::size_t local_point_ids_needed = 0;
         for (std::size_t level = 1; level < cells_per_dim_vec.size()+1; ++level){
@@ -2236,8 +2244,8 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
         comm().allgather(&local_point_ids_needed, 1, point_ids_needed_by_proc.data());
 
         auto expected_max_globalId_cell = std::accumulate(cell_ids_needed_by_proc.begin(),
-                                                                      cell_ids_needed_by_proc.end(),
-                                                                      max_globalId_levelZero + 1);
+                                                          cell_ids_needed_by_proc.end(),
+                                                          max_globalId_levelZero + 1);
         auto min_globalId_cell_in_proc = std::accumulate(cell_ids_needed_by_proc.begin(),
                                                          cell_ids_needed_by_proc.begin()+comm().rank(),
                                                          max_globalId_levelZero + 1);
--- a/opm/grid/cpgrid/CpGrid.cpp
+++ b/opm/grid/cpgrid/CpGrid.cpp
@@ -50,6 +50,7 @@
 #endif
 
 #include "../CpGrid.hpp"
+#include "ParentToChildrenCellGlobalIdHandle.hpp"
 #include <opm/grid/common/MetisPartition.hpp>
 #include <opm/grid/common/ZoltanPartition.hpp>
 //#include <opm/grid/common/ZoltanGraphFunctions.hpp>
@@ -2119,13 +2120,7 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
             OPM_THROW_NOLOG(std::logic_error, "Subdivisions of neighboring LGRs sharing at least one face do not coincide. Not suppported yet.");
         }
     }
-
-    // LGRs Fully Interior: Currently, adding LGRs on a distributed grid is supported only in the case where each LGR is fully contained
-    //                      in the interior of a process, i.e., each cell that is marked for refinement and has a neighboring cell, this
-    //                      neighboring cell has to be also interior for the process. In other words, marked element for refinement cannot
-    //                      have overlap neighboring cells.
-    // To check "LGRs fully interior" for all processes.
-    bool lgrsFullyInteriorHasFailed = false;
+    
     // Non neighboring connections: Currently, adding LGRs whose cells have NNCs is not supported yet.
     // To check "Non-NNCs (non neighboring connections)" for all processes.
     bool nonNNCsHasFailed = false;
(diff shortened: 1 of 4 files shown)


tests/cpgrid/addLgrsOnDistributedGrid_test.cpp
```diff
@@ -276,8 +276,8 @@ void refinePatch_and_check(Dune::CpGrid& coarse_grid,
                 BOOST_CHECK( level_cellIdx[0] == entity.level());
             }
             else{
-                BOOST_CHECK_THROW(entity.father(), std::logic_error);
-                BOOST_CHECK_THROW(entity.geometryInFather(), std::logic_error);
+                //  BOOST_CHECK_THROW(entity.father(), std::logic_error);
```
aritorto, line 279: This was commented during debugging. I'll fix it once the checks are done! (same with other calls)

tests/cpgrid/addLgrsOnDistributedGrid_test.cpp
```diff
@@ -497,9 +519,21 @@ BOOST_AUTO_TEST_CASE(threeLgrs)
...
+        localPointIds_vec.reserve(grid.currentData().back()->size(3));
+        for (const auto& point : vertices(grid.leafGridView())) {
+            // Notice that all partition type points are pushed back. Selecting only interior points does not bring us to the expected value.
+            localPointIds_vec.push_back(grid.currentData().back()->globalIdSet().id(point));
+        }
+        auto [allGlobalIds_points, displPoint ] = Opm::allGatherv(localPointIds_vec, grid.comm());
+        const std::set<int> allGlobalIds_points_set(allGlobalIds_points.begin(), allGlobalIds_points.end());
+
+        // Leaf grid grid total ampunt of points: 11x9x9 + (75-18) + (64-8) + (125-8) = 1121
+        BOOST_CHECK( allGlobalIds_points_set.size() == 1121 );
```
blattms, line 536: This is not of much use here as global ids for points are part of the next step.  Maybe remove it for now?

tests/cpgrid/addLgrsOnDistributedGrid_test.cpp
```diff
@@ -542,6 +576,19 @@ BOOST_AUTO_TEST_CASE(atLeastOneLgr_per_process_attempt)
...
+        for (const auto& point : vertices(grid.leafGridView())) {
+            // Notice that all partition type points are pushed back. Selecting only interior points does not bring us to the expected value.
+            localPointIds_vec.push_back(grid.currentData().back()->globalIdSet().id(point));
+        }
+        auto [allGlobalIds_points, displPoint ] = Opm::allGatherv(localPointIds_vec, grid.comm());
+        const std::set<int> allGlobalIds_points_set(allGlobalIds_points.begin(), allGlobalIds_points.end());
+
+        // Total global ids in leaf grid view for points: 80 + 33 + 56 + 117 + 33 = 319
+        BOOST_CHECK( allGlobalIds_points_set.size() == 319 );
     }
```
blattms, line 592: Same here: Next step?

tests/cpgrid/addLgrsOnDistributedGrid_test.cpp
```diff
@@ -575,9 +622,26 @@ BOOST_AUTO_TEST_CASE(throw_not_fully_interior_lgr)
...
+        }
+        auto [allGlobalIds_points, displPoint ] = Opm::allGatherv(localPointIds_vec, grid.comm());
+        const std::set<int> allGlobalIds_points_set(allGlobalIds_points.begin(), allGlobalIds_points.end());
+
+        // Total global ids in leaf grid view for points: 80 + 33 + 56 + 117 + 33 = 319
+        // To be included when correction/uniqueness of global ids for points of refined level grid is implemented
+        // std::cout<< allGlobalIds_points_set.size() <<std::endl; // Prints 436
+        // BOOST_CHECK( allGlobalIds_points_set.size() == 319 ); // 319 + 117 = 436
+        // It counts double the point global ids from LGR3: once when cell 7 is refined in rank 2, and
+        // once more when cell 7 is refined as an overlap cell in rank 3.
```
blattms, line 644: Same here: Next step?

tests/cpgrid/addLgrsOnDistributedGrid_test.cpp
```diff
@@ -636,42 +697,25 @@ BOOST_AUTO_TEST_CASE(distributed_lgr)
...
+        auto [allGlobalIds_points, displPoint ] = Opm::allGatherv(localPointIds_vec, grid.comm());
+        const std::set<int> allGlobalIds_points_set(allGlobalIds_points.begin(), allGlobalIds_points.end());
+
+        // Total global ids in leaf grid view for points: 80 + 33 + 56 + 117 + 33 = 319
+        // To be included when correction/uniqueness of global ids for points of refined level grid is implemented
+        //   std::cout<< allGlobalIds_points_set.size() <<std::endl; // Prints 436
+        // BOOST_CHECK( allGlobalIds_points_set.size() == 319 ); // 319 + 117 = 436
+        // It counts double the point global ids from LGR3: once when cell 7 is refined in rank 2, and
+        // once more when cell 7 is refined as an overlap cell in rank 3.
+
```
blattms, line 718: Same here: Next step?

tests/cpgrid/addLgrsOnDistributedGrid_test.cpp
```diff
@@ -709,38 +753,22 @@ BOOST_AUTO_TEST_CASE(distributed_lgr_II)
...
-        // Commented since communication step is not implemented yet, therefore refined level grids lack of
-        // global ids for non-interior entities, and the leaf cell index of the leaf grid vie
-        // refinePatch_and_check(grid, cells_per_dim_vec, startIJK_vec, endIJK_vec, lgr_name_vec);
+        auto [allGlobalIds_points, displPoint ] = Opm::allGatherv(localPointIds_vec, grid.comm());
+        const std::set<int> allGlobalIds_points_set(allGlobalIds_points.begin(), allGlobalIds_points.end());
+
+        // Total global ids in leaf grid view for points: 80 (coarse grid points) + 63 (refined points) - 16 (parent cell corners) = 127
+        // To be included when correction/uniqueness of global ids for points of refined level grid is implemented
+        // std::cout<< allGlobalIds_points_set.size() <<std::endl; // Prints 160 (expected value: 127)
+        //BOOST_CHECK( allGlobalIds_points_set.size() == 127 ); // 127 + 33 = 160
```
blattms, line 771: Same here: Next step?

I will stop commenting here. Please assume the same below whenever you check point ids

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -2272,52 +2253,91 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
...
-                    }
+                // Check if it coincides with a corner from level zero. In that case, no global id is needed.
+                const auto& bornLevel_bornIdx =  (*current_data_)[level]->corner_history_[point.index()];
+                if (bornLevel_bornIdx[0] != -1)  { // Corner in the refined grid coincides with a corner from level 0.
+                    // Therefore, search and assign the global id of the previous existing equivalent corner.
+                    const auto& equivPoint = cpgrid::Entity<3>(*( (*current_data_)[bornLevel_bornIdx[0]]), bornLevel_bornIdx[1], true);
+                    localToGlobal_points_per_level[level-1][point.index()] = current_data_->front()->global_id_set_->id( equivPoint );
+                }
+                else {
+                    // Assign new global id only to non-overlap points that do not coincide with
```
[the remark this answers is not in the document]
aritorto: Indeed, the comment is wrong. At this stage in the code, using Dune::Partitions::interior in vertices(..., ...::interior) actually includes all points, since the partition type for points in the refined-level grids hasn’t been set yet. This is handled a bit further down, after cellIndexSet is resized and the partition types for cells in the refined grids are computed.

Partition type of points is a bit tricky since there are 4 types: interior, border, front, overlap. This classification requires that both cell and face partition types are already defined.

Thanks for catching that! I’ll remove the incorrect comment.

tests/cpgrid/addLgrsOnDistributedGrid_test.cpp
aritorto: This is one of the test cases for fully interior LGRs, and I’d like to keep it. The reason this check appears here now is that the previous check for non-duplicated global ids for points in fully interior LGRs was removed. That old check relied on manual calculations using parent cell blocks and LGR dimensions based on ijk-computations, rather than a more stable, long-term solution using DUNE’s grid interface.

I promise to clean this up as soon as the implementation for unique global ids for points is integrated!

tests/cpgrid/addLgrsOnDistributedGrid_test.cpp
aritorto: See my comment above, please. It's another test of fully interior LGRs. I would like to keep it as well :)

tests/cpgrid/addLgrsOnDistributedGrid_test.cpp
aritorto: I will remove the checks that correspond to the general case and keep the ones of fully interior LGRs, if that's okay with you too.

aritorto:
> Concerning the checks for unique point ids:
> 
> * Is there a decent check for the case that the LGR is in the interior without touching border faces?
> * What should be do with the checks, where know that the check is actually wrong. We should either remove those or add a bug disclaimer that the check is wrong because this is not correctly implemented yet. What do you think.

The check for the cases when LGRs are fully interior are [here ](https://github.com/OPM/opm-grid/blob/ba32782f2bb45d84fa9228b057e7cb6fa366cf9f/tests/cpgrid/addLgrsOnDistributedGrid_test.cpp#L536)and [here](https://github.com/OPM/opm-grid/blob/ba32782f2bb45d84fa9228b057e7cb6fa366cf9f/tests/cpgrid/addLgrsOnDistributedGrid_test.cpp#L591).  Not sure if they count as 'decent'. The reason why this is done separately, i.e., not in refinePatch_and_check(...) definition, is that the implementation for the general case (on correct/unique point global ids) is not done yet.

Now, the wrong-checks are removed. However, there was in each wrong check [a small comment](https://github.com/OPM/opm-grid/blob/671f15bcb7a24f118551499b139a178e598e3613/tests/cpgrid/addLgrsOnDistributedGrid_test.cpp#L824).  I think now it looks cleaner, only the checks for fully interior LGRs are left.

tests/cpgrid/addLgrsOnDistributedGrid_test.cpp
```diff
@@ -104,8 +104,8 @@ void refinePatch_and_check(Dune::CpGrid& coarse_grid,
         {
             Dune::cpgrid::Entity<0> entity = Dune::cpgrid::Entity<0>(*data[0], cell, true);
             BOOST_CHECK( entity.hasFather() == false);
-            BOOST_CHECK_THROW(entity.father(), std::logic_error);
-            BOOST_CHECK_THROW(entity.geometryInFather(), std::logic_error);
+            //BOOST_CHECK_THROW(entity.father(), std::logic_error);
+            //BOOST_CHECK_THROW(entity.geometryInFather(), std::logic_error);
```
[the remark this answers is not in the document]
aritorto: Indeed, it was commented out to avoid throw messages during debugging and needs to be uncommented again. Thanks!

aritorto:
> What happens if somebody marks elements including some that touch border faces and calls adapt? Will the the user get an exception? Is this tested?
> What happens if cells of a distributed grid are marked of which all faces are interior?

Currently, the code to correctly assign global ids for refined level grids and leaf grid view when the coarse grid has been distributed is located in CpGrid::addLgrsUpdateLeafGRidView(...). I would need to go through the code in details, but in principle, there is no exception. I can also add a test just calling adapt in such a situation. In any case, adapt(...) is invoked in addLgrsUpdateLeafGRidView(...) definition, after marking elements to be refined, regardless of the partition type of their faces.

aritorto:
All comments have been addressed, thanks your your review!

aritorto:

opm/grid/cpgrid/CpGrid.cpp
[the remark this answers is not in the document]
blattms: Thanks for the clarification. Should we add a comment about partitions not being ready to use, if it is not there already? This might not be obvious to everyone.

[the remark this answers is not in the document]
aritorto: There is a  comment [here](https://github.com/OPM/opm-grid/blob/abe6a69546031cc4a4dcf0c02c2290c216c452e4/opm/grid/cpgrid/CpGrid.cpp#L2200) but it's not 100% clear. I'll re-write that, thanks!

tests/cpgrid/addLgrsOnDistributedGrid_test.cpp
blattms: Good, then let's not forget.

blattms (commented):
Just a tine newline missing. I will accept my own suggestion...

blattms:
jenkins build this please

blattms:
And then he forgot about this one. Sorry.

PR #786, opened 2024-10-31
Change C arrays into std::vectors in test_graphofgrid

michal-toth:


michal-toth:
jenkins build this please

bska:
> Thanks for the quick fix! Merging.

Just out of curiosity: Did you verify that this fixes the CLang build?

atgeirr:
> Just out of curiosity: Did you verify that this fixes the CLang build?

Yes, no warning anymore. However, there is another, unrelated in the same file:
```
/Users/atgeirr/opm/src/opm-grid/tests/test_graphofgrid.cpp:154:25: warning: adding 'value_type' (aka 'int') to a string does not append to the string [-Wstring-plus-int]
  153 |     std::string message("Expecting an error message from getGraphOfGridEdgeList, the vertex "
      |                         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  154 |                         + gIDs[16] + std::string(" has a wrong number of edges."));
      |                         ^~~~~~~~~~
```
Can you fix this as well @michal-toth ?

michal-toth:
Fix is on its way.

PR #787, opened 2024-10-31
Clear warning, properly cast int into string.

michal-toth:


michal-toth:
jenkins build this please

tests/test_graphofgrid.cpp
```diff
@@ -151,7 +151,7 @@ BOOST_AUTO_TEST_CASE(WrapperForZoltan)
 
     numEdges[16] = 8;
     std::string message("Expecting an error message from getGraphOfGridEdgeList, the vertex "
-                        + gIDs[16] + std::string(" has a wrong number of edges."));
+                        + std::to_string(gIDs[16]) + std::string(" has a wrong number of edges."));
```
blattms, line 154: Looking at the next line, I am wondering whether we really want to use the logger here as we do not register any of those.
When do want this displayed? Would [`BOOST_MESSAGE`](https://www.boost.org/doc/libs/1_31_0/libs/test/doc/components/test_tools/reference/BOOST_MESSAGE.html) be the better option?

michal-toth: This test fails the check inside getGraphOfGridEdgeList function comparing the number of edges the vertex has in the GraphOfGrid, and the number of edges expected by Zoltan. Since it is a callback function to Zoltan, it does not throw but sets *err to ZOLTAN_FATAL.
I wanted to give additional information about the failure and have put it in OpmLog::error. The test thus leaves an error in the logger. I did not want to leave an error in the log without any context, so that is how that message ended in OpmLog::info.

atgeirr: Just be aware that any logging ends up in nothing if there are no logging backends set up, for example in a test fixture. For standalone code, the simplest way to do this is to call the static function `OpmLog::setupSimpleDefaultLogging()`. OPM Flow has a more complex setup, logging to .PRT and .DBG files as well as standard out.

bska: > for example in a test fixture

If you wish to do this you could have a look at [`test_LogOutputHelper.cpp`](https://github.com/OPM/opm-simulators/blob/e81cf62e796b66ebaa6ec7476926cc3aba25d370/tests/test_LogOutputHelper.cpp#L105) in the OPM-Simulators repository. which not only sets up logging at the start of a test (fixture's [constructor](https://github.com/OPM/opm-simulators/blob/e81cf62e796b66ebaa6ec7476926cc3aba25d370/tests/test_LogOutputHelper.cpp#L113)), but also tears down the logging support at the end of the test (fixture's [destructor](https://github.com/OPM/opm-simulators/blob/e81cf62e796b66ebaa6ec7476926cc3aba25d370/tests/test_LogOutputHelper.cpp#L118)).

> I wanted to give additional information about the failure and have put it in OpmLog::error. The test thus leaves an error in the logger. I did not want to leave an error in the log without any context, so that is how that message ended in OpmLog::info.

As @atgeirr says, calling `OpmLog::error()` (or `OpmLog::info()`) when there's no logging backend just means that all messages are lost.  For unit testing you're generally much better off using the `BOOST_TEST_MESSAGE` or `BOOST_MESSAGE` macros unless you're specifically testing the logging system.

PR #792, opened 2024-11-05
Get xyz from vertices in LGRs and 8-corners-of-refined-cells

aritorto:
This PR aims to contribute to the generation of output files for simulations involving Corner Point Grids (CpGrid) with Local Grid Refinement (LGR). It demonstrates how to extract the necessary values for defining COORD and ZCORN in CpGrids with LGRs. Note that the ordering presented here does not follow the "pillars" construction. To correctly output COORD and ZCORN, a reordering step must be implemented.

This [commit](https://github.com/OPM/opm-grid/pull/792/commits/bf6bd6ab4e0ef006891dc0b01c15f23fcbaec0ac) translates one of the test cases from OPM/opm-common#4255 using the grid interface.

### DIFF
--- a/tests/cpgrid/coord_zcorn_lgr_test.cpp
+++ b/tests/cpgrid/coord_zcorn_lgr_test.cpp
@@ -98,8 +98,7 @@ checkGlobalCellLgr(Dune::CpGrid& grid)
         //    belongs to a refined level grid). This default Cartesian Index for leaf cells coincide with the globalCell
         //    values.
         const auto& cartesian_idx_from_leaf_elem = mapper.cartesianIndex(element.index());
-        const auto& global_cell_idx_leaf
-            = grid.globalCell()[element.index()]; // parent cell index when the cell is a refined one.
+        const auto& global_cell_idx_leaf = grid.globalCell()[element.index()]; // parent cell index when the cell is a refined one.
         BOOST_CHECK_EQUAL(cartesian_idx_from_leaf_elem, global_cell_idx_leaf);
         // global_ijk represents the ijk values of the equivalent cell on the level zero, or parent cell if the leaf
         // cell is a refined one. Notice that all the refined cells of a same parent cell will get the same global_cell_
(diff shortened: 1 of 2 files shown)


tests/cpgrid/coord_zcorn_lgr_test.cpp
```diff
@@ -0,0 +1,277 @@
...
+        // Each refined cell has its 8 corners stored in <refined_level_grid>.cell_to_point_[ refined element index ]
+        // in the following order:
+        //     BOTTOM FACE          TOP FACE
+        //     2 --- 3              6 --- 7
+        //    /     /              /     /
+        //   0 --- 1              4 --- 5
+        //
+        std::vector<std::vector<std::array<double,3>>> xyz_for_zcorn(grid.currentData()[level]->size(0)); // size total amount of cells in the level refined grid.
+        for (const auto& element : elements(grid.levelGridView(level))) {
+            xyz_for_zcorn[element.index()].reserve(8);
```
blattms, line 182: Maybe also get the local global index. Similar to above. That together with the order you documented above will be all the information needed.

blattms (commented):
Looks great (from a short glimpse).

tests/cpgrid/coord_zcorn_lgr_test.cpp
```diff
@@ -0,0 +1,277 @@
...
+        // points are entered with the X axis cycling fastest. Next come the depths of the bottom of the first layer. The
+        // top of layer two follows etc.
+        //
+        // Expected values in each LGR: 8*refined_cells_in_that_lgr
+        // Each refined cell has its 8 corners stored in <refined_level_grid>.cell_to_point_[ refined element index ]
+        // in the following order:
+        //     BOTTOM FACE          TOP FACE
+        //     2 --- 3              6 --- 7
+        //    /     /              /     /
+        //   0 --- 1              4 --- 5
```
bska, line 178: It typically does not matter, but for what it's worth, the default coordinate system is right-handed with the origin in the upper, left, back corner and Z axis pointing down.  In other words, the **typical** node ordering is as outlined in the picture in MRST's [`processGRDECL`](https://github.com/SINTEF-AppliedCompSci/MRST/blob/5bcfe7cd64c2a9c3bdb298af46c7fe4c0df8feb9/core/gridprocessing/processGRDECL.m#L136-L149) function.

aritorto: I see, in the grid, we follow the described ordering. Therefore, for output, we need to re-order them according to your description. Thanks for your feedback!

tests/cpgrid/coord_zcorn_lgr_test.cpp
aritorto: I will add that here then. Thanks for your feedback!

tests/cpgrid/coord_zcorn_lgr_test.cpp
bska: > Therefore, for output, we need to re-order them according to your description.

Sorry for being too brief: It's not that we need to do anything in particular to the order&ndash;it's actually the same in both views&ndash;it's just that in terms of labelling the **top** face is normally comprised of vertices 0..3 while the **bottom** face normally comprises vertices 4..7.

aritorto: Thanks for clarifying! I added your comment into the documentation.

aritorto:
[coord_zcorn.log](https://github.com/user-attachments/files/17647065/coord_zcorn.log)

aritorto:
I will run Jenkins and leave the decision to merge this PR up to you (@blattms, @bska), as it is more illustrative than testing actual features.

aritorto:
jenkins build this please

PR #795, opened 2024-11-06
use constexpr instead of anonymous enums

akva2:
c++-20 warns for arithmetics using enums


akva2:
jenkins build this please

bska (approved):
Much appreciated.  It's annoying, tough, that we have to wait for C++23 to get [`to_underlying()`](https://en.cppreference.com/w/cpp/utility/to_underlying).

PR #796, opened 2024-11-06
Replace outdated macro (removed in dune 2.10)

timokoch:
The macro redefinition already existed in 2.7 so it shouldn't be a problem to change this.

Of course an alternative is to remove the version check altogether


timokoch:
duplicate of #758

PR #798, opened 2024-11-15
Bug fix pinch-out log output

daavid00:
PinchNNC was missing in nnc_cells. Here also adding to not write this message if no connections are generated, but I could remove it if this is required.


daavid00:
jenkins build this please

opm/grid/cpgrid/processEclipseFormat.cpp
```diff
@@ -323,8 +323,10 @@ namespace cpgrid
                 }
             }
 
-            auto suffix = std::string{(nnc_cells[PinchNNC].size() != 1)? "s" : ""};
-            Opm::OpmLog::info(std::to_string(nnc_cells.size()) + " pinch-out connection" + suffix + " generated");
+            if (nnc_cells[PinchNNC].size() > 0) {
+                auto suffix = std::string{(nnc_cells[PinchNNC].size() != 1)? "s" : ""};
+                Opm::OpmLog::info(std::to_string(nnc_cells[PinchNNC].size()) + " pinch-out connection" + suffix + " generated");
+            }
```
bska, line 329: This doesn't do anything other than remove the print-out for the zero pinch-connection case.  Is there some part of this PR missing?

On another note, please use `! .empty()` instead of `.size() > 0` when the container is a [`vector<>`](https://github.com/OPM/opm-grid/blob/c2c7478fae50e639e7df354db4c301717a36660c/opm/grid/cpgrid/processEclipseFormat.cpp#L215)

bska: > Before this PR, `2 pinch-out connections generated` is always written, even in cases without nnc.

Doh, yeah I missed that.  It's `nnc_cells.size()` vs. `nnc_cells[PinchNNC].size()` and since `nnc_cells` is an `array<NNCMap,2>` that will of course always print 2.  Thanks for the hint.

bska:
jenkins build this please

PR #799, opened 2024-11-15
Fix GPL license declarations for two tests.

blattms:
Those were missing the plus and the short description of the license.


blattms:
jenkins build this please

alfbr:
Thanks for the fix, in it goes!

PR #800, opened 2024-11-19
Correct point ids

aritorto:


aritorto:
Replaced/improved by OPM/opm-grid#801

PR #801, opened 2024-11-19
Unique point global ids for refined level grids

aritorto:
This PR introduces a structure to manage the point global IDs of  all refined child cells per parent. This structure ensures the correct assignment of global point IDs in refined grids, gathering already-assigned IDs for interior cells and distributing them for overlap cells. 

Current approach avoids duplicated point ids in the following setting:
Let's say a block of cells to be refined is distributed in P_{i_0}, ..., P_{i_n}, with n+1 < grid.comm().size().
If all cells to be refined are not "seen"(neither interior nor overlap) for every process P with P != P_{i_j}, j = 0, ..., n.

When this fails, duplicated ids appear on edges or faces "shared" with unseen cells. 

Keeping this potential duplication of point ids in mind, this PR allows distributing Local Grid Refinements (LGRs) across different processes.


aritorto:

aritorto:
Improve by OPM/opm-grid#802

PR #802, opened 2024-11-21
Point ids for refined grids and addLgrsUpdateLeafGridView refactorization

aritorto:
This PR introduces a structure to manage the point global IDs of all refined child cells per parent. This structure ensures the correct assignment of global point IDs in refined grids, gathering already-assigned IDs for interior cells and distributing them for overlap cells.

Current approach avoids duplicated point ids in the following setting:
Let's say a block of cells to be refined is distributed in P_{i_0}, ..., P_{i_n}, with n+1 < grid.comm().size().
If all cells to be refined are not "seen"(neither interior nor overlap) for every process P with P != P_{i_j}, j = 0, ..., n.

Points that may have multiple ids as shown and explained in OPM/opm-grid#806. 
Cells that only share corners (not faces) with interior cells of a process are not included in its overlap layer, therefore, some point may have multiple ids. 

Keeping this potential multiple point ids in mind, this PR allows distributing Local Grid Refinements (LGRs) across different processes, subject to first distribute level zero grid, and after that add the LGRs. To be able to run the simulation, some changes to respect this order must be done in opm-simulators.

Improvement/replacement for OPM/opm-grid#801

Not relevant for the Reference Manual

### DIFF
--- a/opm/grid/cpgrid/CpGrid.cpp
+++ b/opm/grid/cpgrid/CpGrid.cpp
@@ -51,6 +51,7 @@
 
 #include "../CpGrid.hpp"
 #include "ParentToChildrenCellGlobalIdHandle.hpp"
+#include "ParentToChildCellToPointGlobalIdHandle.hpp"
 #include <opm/grid/common/MetisPartition.hpp>
 #include <opm/grid/common/ZoltanPartition.hpp>
 //#include <opm/grid/common/ZoltanGraphFunctions.hpp>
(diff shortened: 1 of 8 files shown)


aritorto:

akva2:
some new warnings, please address

aritorto:

aritorto:

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -2082,37 +2415,9 @@ void CpGrid::addLgrsUpdateLeafView(const std::vector<std::array<int,3>>& cells_p
...
-                    compatibleSubdivisionsHasFailed = true;
-                    break;
-                }
-            } // end-otherLevel-for-loop
-        } // end-level-for-loop
-    }// end-if-patchesShareFace
-    compatibleSubdivisionsHasFailed = comm().max(compatibleSubdivisionsHasFailed);
-    if(compatibleSubdivisionsHasFailed) {
+    bool compatibleSubdivisions = current_view_data_->compatibleSubdivisions(cells_per_dim_vec, startIJK_vec, endIJK_vec);
+    compatibleSubdivisions= comm().max(compatibleSubdivisions);
```
blattms, line 2419: I think there is a mistake here as the maximum between false and true would be true.
Please make compatibleSubdivisions an int as MPI within DUNE does not support bool directly.

opm/grid/cpgrid/CpGridData.hpp
```diff
@@ -412,6 +412,19 @@ class CpGridData
...
+    /// @brief Check compatibility of number of subdivisions of neighboring LGRs.
+    ///
+    /// Check shared faces on boundaries of LGRs. Not optimal since the code below does not take into account
+    /// active/inactive cells, instead, relies on "ijk-computations".
+    ///
+    /// @param [in]  cells_per_dim_vec    Vector of expected subdivisions per cell, per direction, in each LGR.
+    /// @param [in]  startIJK_vec         Vector of Cartesian triplet indices where each patch starts.
+    /// @param [in]  endIJK_vec           Vector of Cartesian triplet indices where each patch ends.
+    ///                                   Last cell part of the lgr will be {endIJK_vec[patch][0]-1, ..., endIJK_vec[patch][2]-1}.
+    bool compatibleSubdivisions(const std::vector<std::array<int,3>>& cells_per_dim_vec,
```
blattms, line 424: Please add doc for `/// @return`

opm/grid/cpgrid/ParentToChildCellToPointGlobalIdHandle.hpp
```diff
@@ -0,0 +1,172 @@
...
+        // value (e.g., '42').
+        if ( (element.partitionType() != Dune::InteriorEntity) || (level==-1)) {
+            buffer.write(42);
+            return;
+        }
+        // Store the children's corner global ids in the buffer when the element is interior and has children.
+        // Write the rank first, for example via the "corner 0" of cell_to_point_ of the first child:
+        // First child: children[0]
+        // First corner of first child:  level_cell_to_point_[ level -1 ][children[0]] [0]
+        buffer.write(level_winning_ranks_[level-1][ level_cell_to_point_[ level -1 ][children[0]] [0] ]); // winner rank
```
blattms, line 118: I think this is wrong. We need to send the rank of the process here (`comm.rank()`);

tests/cpgrid/addLgrsOnDistributedGrid_test.cpp
```diff
@@ -104,8 +104,8 @@ void refinePatch_and_check(Dune::CpGrid& coarse_grid,
         {
             Dune::cpgrid::Entity<0> entity = Dune::cpgrid::Entity<0>(*data[0], cell, true);
             BOOST_CHECK( entity.hasFather() == false);
-            BOOST_CHECK_THROW(entity.father(), std::logic_error);
-            BOOST_CHECK_THROW(entity.geometryInFather(), std::logic_error);
+            // BOOST_CHECK_THROW(entity.father(), std::logic_error);
+            // BOOST_CHECK_THROW(entity.geometryInFather(), std::logic_error);
```
blattms, line 108: Why is this commented out?

tests/cpgrid/addLgrsOnDistributedGrid_test.cpp
```diff
@@ -276,8 +276,8 @@ void refinePatch_and_check(Dune::CpGrid& coarse_grid,
                 BOOST_CHECK( level_cellIdx[0] == entity.level());
             }
             else{
-                BOOST_CHECK_THROW(entity.father(), std::logic_error);
-                BOOST_CHECK_THROW(entity.geometryInFather(), std::logic_error);
+                // BOOST_CHECK_THROW(entity.father(), std::logic_error);
+                // BOOST_CHECK_THROW(entity.geometryInFather(), std::logic_error);
```
blattms, line 280: Why is this commented out?

tests/cpgrid/addLgrsOnDistributedGrid_test.cpp
```diff
@@ -409,21 +398,6 @@ void refinePatch_and_check(Dune::CpGrid& coarse_grid,
...
-    for (const auto& point: vertices(coarse_grid.leafGridView(), Dune::Partitions::interior)) {
-        localInteriorPointIds_vec.push_back(data.back()->globalIdSet().id(point));
-        ++local_interior_points_count;
-    } 
-    auto global_point_count  = coarse_grid.comm().sum(local_interior_points_count);
-    auto [allGlobalIds_points, displ_point] = Opm::allGatherv(localInteriorPointIds_vec, coarse_grid.comm());
-
-    const std::set<int> allGlobalIds_points_set(allGlobalIds_points.begin(), allGlobalIds_points.end());
-    BOOST_CHECK( static_cast<int>(allGlobalIds_points.size()) == global_point_count);
-    BOOST_CHECK( allGlobalIds_points.size() == allGlobalIds_points_set.size() );
```
blattms, line 425: This is deactivated because there is a bug right? Please add a prominent comment and create a github issue about this.

tests/cpgrid/addLgrsOnDistributedGrid_test.cpp
```diff
@@ -714,14 +803,48 @@ BOOST_AUTO_TEST_CASE(distributed_lgr_II)
 
         grid.addLgrsUpdateLeafView(cells_per_dim_vec, startIJK_vec, endIJK_vec, lgr_name_vec);
         refinePatch_and_check(grid, cells_per_dim_vec, startIJK_vec, endIJK_vec, lgr_name_vec);
+
+        // Check global id is not duplicated for points for each LGR
+        // LGR1 dim 6x2x2 -> 7x3x3 = 63 points
+        std::vector<int> local_point_ids;
+        local_point_ids.reserve(63); // expected_point_ids in LGR1
+        for (const auto& element : elements(grid.levelGridView(1))) {
```
blattms, line 811: What is the reason why we do not iterate over vertices directly?

tests/cpgrid/addLgrsOnDistributedGrid_test.cpp
```diff
@@ -753,6 +876,57 @@ BOOST_AUTO_TEST_CASE(distributed_in_all_ranks_lgr)
...
+            for (int corner = 0; corner < 8; ++corner)
+            {
+                const auto& point = element.subEntity<3>(corner);
+                local_point_ids.push_back(grid.currentData()[1]->globalIdSet().id(point));
+            }
+        }
+        auto [all_point_ids, displPoint ] = Opm::allGatherv(local_point_ids, grid.comm());
+        const std::set<int> all_point_ids_set(all_point_ids.begin(), all_point_ids.end());
+        // Difference with preveious test case:
```
blattms, line 893: previous

tests/cpgrid/addLgrsOnDistributedGrid_test.cpp
```diff
@@ -753,6 +876,57 @@ BOOST_AUTO_TEST_CASE(distributed_in_all_ranks_lgr)
...
+        // cell 13 = { overlap in P0, interior in P1, does not exist in P2, overlap in P3}
+        // cell 14 = { does not exist in P0, overlap in P1, overlap in P2, interior in P3}
+        // cell 17 = { overlap in P0, interior in P1, overlap in P2, does not exist in P3}
+        // cell 18 = { does not exist in P0, overlap in P1, interior in P2, overlap in P3}
+        // Remark: the LGR is distributed in ALL processes BUT
+        // - P0 does NOT see cell 18, but sees all the others.
+        // - P1 does NOT see cells 2 and 6, but sees the rest of them.
+        // - P2 does NOT see cell 13, but sees all the others.
+        // - P3 does NOT see cell 1, 5 and 17, but sees all the others.
+        // Remark: the LGR is distributed only in 2 processes which are the only two processes seeing these cells.
```
blattms, line 908: I don't understand this sentence. Are there more than 1 LGR (which is said to be distributed among all processes above)

tests/cpgrid/addLgrsOnDistributedGrid_test.cpp
```diff
@@ -753,6 +876,57 @@ BOOST_AUTO_TEST_CASE(distributed_in_all_ranks_lgr)
...
+            for (int corner = 0; corner < 8; ++corner)
+            {
+                const auto& point = element.subEntity<3>(corner);
+                local_point_ids.push_back(grid.currentData()[1]->globalIdSet().id(point));
+            }
+        }
+        auto [all_point_ids, displPoint ] = Opm::allGatherv(local_point_ids, grid.comm());
+        const std::set<int> all_point_ids_set(all_point_ids.begin(), all_point_ids.end());
+        // Difference with preveious test case:
+        // Global cell ids of cells to be refined = {1,2,5,6,13,14,17,18}
```
blattms, line 894: Please indicate that this a bug originating from the way that we add overlap cells.

opm/grid/cpgrid/ParentToChildCellToPointGlobalIdHandle.hpp
```diff
@@ -0,0 +1,172 @@
...
+    bool fixedSize(std::size_t, std::size_t)
+    {
+        return false;
+    }
+    // Only communicate values attached to cells.
+    bool contains(std::size_t, std::size_t codim)
+    {
+        return codim == 0;
+    }
+    // Communicate variable size: 1 (rank) + (8* amount of child cells) from an interior parent cell from level zero grid.
```
blattms, line 87: These comments seem more like comments about the implementation. Maybe move them inside the methods?

blattms (changes requested):
Thanks a lot.
I have a few suggestions where you should probably change the code.

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1212,6 +1213,344 @@ Dune::cpgrid::Intersection CpGrid::getParentIntersectionFromLgrBoundaryFace(cons
...
+    for(const auto& element: elements(this->leafGridView())) {
+        std::array<int,3> ijk;
+        getIJK(element.index(), ijk);
+        for (std::size_t level = 0; level < startIJK_vec.size(); ++level) {
+            bool belongsToLevel = true;
+            for (int c = 0; c < 3; ++c) {
+                belongsToLevel = belongsToLevel && ( (ijk[c] >= startIJK_vec[level][c]) && (ijk[c] < endIJK_vec[level][c]) );
+                if (!belongsToLevel)
+                    break;
+            }
```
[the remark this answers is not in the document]
aritorto: Code is slightly simplified in a coming-soon-commit.  In nonNNCsSelectedCellsLGR(...) we only check if the selected cells for refinement (via blocks with startIJK endIJK values) have NNCs. If at least one cell of one block has NNCs, we return false. 

We mark elements for refinement in CpGrid::markElemeAssignLevelDetectActiveLgrs(...)

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1212,6 +1213,344 @@ Dune::cpgrid::Intersection CpGrid::getParentIntersectionFromLgrBoundaryFace(cons
...
+    //    3. Communicate the already assigned cell global ids from interior to overlap refined cells.
+    // Refined level grid points/vertices:
+    // There are 4 partition types: interior, border, front, overlap. This classification requires that both
+    // cell and face partition types are already defined, not available yet for refined level grids.
+    //    1. Assign for all partition type points a 'candidate of global id' (unique in each process).
+    //       Except the points that coincide with a point from level zero.
+    //    2. To make point ids globally unique, re-write the values for points that are corners of overlap
+    //       refined cells, via communication.
+    // Under the assumption of LGRs fully-interior, no communication is needed. In the general case, communication will be used
+    // to populate overlap cell/point global ids on the refined level grids.
```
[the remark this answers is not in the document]
aritorto: Comment added and documentation moved where the method is invoked.

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1212,6 +1213,344 @@ Dune::cpgrid::Intersection CpGrid::getParentIntersectionFromLgrBoundaryFace(cons
...
+
+    auto expected_max_globalId_cell = std::accumulate(cell_ids_needed_by_proc.begin(),
+                                                      cell_ids_needed_by_proc.end(),
+                                                      max_globalId_levelZero + 1);
+    min_globalId_cell_in_proc = std::accumulate(cell_ids_needed_by_proc.begin(),
+                                                cell_ids_needed_by_proc.begin()+comm().rank(),
+                                                max_globalId_levelZero + 1);
+    min_globalId_point_in_proc = std::accumulate(point_ids_needed_by_proc.begin(),
+                                                 point_ids_needed_by_proc.begin()+ comm().rank(),
+                                                 expected_max_globalId_cell);
```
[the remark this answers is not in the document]
aritorto: I added the plus 1. I couldn't tell now if I forgot to add it, or if it's not necessary, but sure thing is that that does not hurt since, in some cases, we are overestimating anyway.

tests/cpgrid/addLgrsOnDistributedGrid_test.cpp
aritorto: Not to see million lines of errors when debugging. I forgot to undo this for the PR, sorry for that!

opm/grid/cpgrid/CpGrid.cpp
aritorto: Thanks! Nice catch, it's indeed incorrect.

opm/grid/cpgrid/ParentToChildCellToPointGlobalIdHandle.hpp
aritorto: In principle, I would say that "level_winning_ranks_[level-1][ level_cell_to_point_[ level -1 ][children[0]] [0] ]" is exactly comm().rank(), but I might be wrong, based on how level_winning_ranks gets populated (see screenshot).  To be sure, I'll follow your suggestion then. 
Screenshot from 2024-11-26 10-53-29

aritorto: Thanks, I\ll correct that.

tests/cpgrid/addLgrsOnDistributedGrid_test.cpp
aritorto: Good point!

tests/cpgrid/addLgrsOnDistributedGrid_test.cpp
aritorto: I will try to improve the comment. There is only one LGR here. The block of parent cells  has dimension 2x2x2, the parent cell indices are {1,2,5,6,13,14,17,18}. Each parent cell gets refined into 2x2x2 children, therefore the LGR dimension is 4x4x4.

aritorto:
@blattms all commnets have been addressed, many thanks for your review!

aritorto:

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1212,6 +1213,328 @@ Dune::cpgrid::Intersection CpGrid::getParentIntersectionFromLgrBoundaryFace(cons
...
+        for (const auto& element : elements(levelGridView(level))) {
+            // For interior cells, rewrite the rank value - later used in "point global id competition".
+            if (element.partitionType() == InteriorEntity) {
+                for (const auto& corner : currentData()[level]->cell_to_point_[element.index()]){
+                    int rank = comm().rank();
+                    level_winning_ranks[level -1][corner] = rank;
+                }
+            }
+        }
+    }
```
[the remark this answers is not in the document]
aritorto: I see your point, but the Handle does not see any grid, so changing the for-loop to elemIdx with size of level_cell_to_point_  would be okay, but then to detect if the each element is interior we need an Entity<0>. To create an Entity, we need each refined level grid. 
It can be done, but implies passing to the constructor of the handle all refined level grids.

aritorto:

blattms (approved):
Thanks. Fair enough. Merging

PR #805, opened 2024-11-27
[bug fixed] Use idset_ for sequential run in getMaxCodimGlobalId()

aritorto:
When computing the maximum global id from level zero grid on a distributed grid, it's necessary to distinguish between two cases: sequential and parallel. This PR adds the  missing case of sequential.


aritorto:

PR #806, opened 2024-11-27
On loadBalance(...) "addCornerCells" should be called "addEdgeCells"?

aritorto:
In an attempt to assign unique point ids on a distributed grid with LGRs, I invoked loadBalance, enabling the inclusion of cells that share corners with interior cells on each process by setting addCornerCells = true (one of the function arguments in loadBalance). However, instead of adding corner-sharing cells, this configuration primarily adds cells that share edges, in addition to the overlap layer of size 1. As a result, cells that share only a corner are not included, which can impact the uniqueness of point ids in certain cases.

In the test addLgrsOnDistributedGrid_test.cpp, both escenarios are included:
- cases where setting addCornerCells=true fix the issue of unique point ids 
- cases where setting addCornerCells=true is not enough to achieve unique point ids (cells that share only corners - and not edges - are not included)

Based on OPM/opm-grid#802. 

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -202,6 +202,7 @@ list (APPEND PUBLIC_HEADER_FILES
   opm/grid/LookUpData.hh
   opm/grid/cpgrid/OrientedEntityTable.hpp
   opm/grid/cpgrid/ParentToChildrenCellGlobalIdHandle.hpp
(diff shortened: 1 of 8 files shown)


aritorto:

aritorto:
REplaced by OPM/opm-grid#808 (too many merge conflicts here)

PR #803, opened 2024-11-25
Global refinement partially supported on distributed CpGrid

aritorto:
Based on OPM/opm-grid#802.

This PR moves the definition of global ids from addLgrsUpdateLeafView(...) to adapt(...), allowing refinement in a more general setting. For example, global refinement (currently, only possible to invoke once with argument equal to 1), or refinement of elements not necessary forming a block of cells.

Not relevant for the Reference Manual.

PR #807, opened 2024-11-28
Global refinement partially supported on distributed CpGrid

aritorto:
This PR moves the definition of global ids from addLgrsUpdateLeafView(...) to adapt(...), allowing refinement in a more general setting. For example, global refinement (currently, only possible to invoke once with argument equal to 1), or refinement of elements not necessary forming a block of cells.

Replacement for OPM/opm-grid#803.

### DIFF
--- a/opm/grid/cpgrid/CpGrid.cpp
+++ b/opm/grid/cpgrid/CpGrid.cpp
@@ -1020,16 +1020,15 @@ void CpGrid::globalRefine (int refCount)
             OPM_THROW(std::logic_error, "Global refinement of a mixed grid with coarse and refined cells is not supported yet.");
         }
     }
-    // For a distributed grid, global refinement can be called only once - for now.
-    if ( (refCount>1) && (!distributed_data_.empty()) ) {
-        OPM_THROW(std::logic_error, "Multiple calls of global refinement on a distributed grid are not supported yet.");
-    }
     if (refCount>0) {
         const std::vector<std::array<int,3>>& cells_per_dim_vec = {{2,2,2}}; // Arbitrary chosen values.
         for (int refinedLevel = 0; refinedLevel < refCount; ++refinedLevel) {
 
             std::vector<int> assignRefinedLevel(current_view_data_-> size(0));
             const auto& preAdaptMaxLevel = this ->maxLevel();
+            std::vector<std::string> lgr_name_vec = { "GR" + std::to_string(preAdaptMaxLevel +1) };
+            bool isCARFIN = true;
+            const std::array<int,3>& endIJK = currentData().back()->logicalCartesianSize();
 
             for(const auto& element: elements(this-> leafGridView())) {
                 // Mark all the elements of the current leaf grid view for refinement
@@ -1037,9 +1036,6 @@ void CpGrid::globalRefine (int refCount)
                 assignRefinedLevel[element.index()] = preAdaptMaxLevel +1;
             }
 
-            std::vector<std::string> lgr_name_vec = { "GR" + std::to_string(preAdaptMaxLevel +1) };
-            bool isCARFIN = true;
-            const std::array<int,3>& endIJK = currentData().back()->logicalCartesianSize();
             preAdapt();
             adapt(cells_per_dim_vec, assignRefinedLevel, lgr_name_vec, isCARFIN, {{0,0,0}}, {endIJK});
             postAdapt();
@@ -1260,30 +1256,11 @@ bool CpGrid::nonNNCsSelectedCellsLGR( const std::vector<std::array<int,3>>& star
     return true;
 }
 
-std::pair<std::vector<int>, std::vector<int>> CpGrid::markElemAssignLevelDetectActiveLgrs(const std::vector<std::array<int,3>>& startIJK_vec,
-                                                                                          const std::vector<std::array<int,3>>& endIJK_vec,
-                                                                                          bool onlyDetectLgrs)
+template<class T>
+void CpGrid::computeOnLgrParents(const std::vector<std::array<int,3>>& startIJK_vec,
+                                 const std::vector<std::array<int,3>>& endIJK_vec,
+                                 T func)
 {
-    // Auxiliary function.
-    // If onlyDetectLgrs == true, then only activeLgrs gets populated.
-    // Otherwise, both activeLgrs and assignLevel get populated.
-    auto detectLgrs_maybeAssignLevel = [this](const Dune::cpgrid::Entity<0>& element,
-                                              int level,
-                                              std::vector<int>& active_lgrs,
-                                              std::vector<int>& assign_level,
-                                              bool only_detect_lgrs)
-    {
-        // shifted since starting grid is level 0, and refined grids levels are >= 1.
-        active_lgrs[level] = 1;
-        if (!only_detect_lgrs)
-        {
-            this-> mark(1, element);
-            assign_level[element.index()] = level+1; // shifted since starting grid is level 0, and refined grids levels are >= 1.
-        }
-    };
-
-    std::vector<int> activeLgrs(startIJK_vec.size());
-    std::vector<int> assignLevel(onlyDetectLgrs ? 0 : currentData().back()->size(0));
     // Find out which (ACTIVE) elements belong to the block cells defined by startIJK and endIJK values.
     for(const auto& element: elements(this->leafGridView())) {
         std::array<int,3> ijk;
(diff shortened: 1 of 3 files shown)


aritorto:

aritorto:

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -2356,105 +2404,7 @@ bool CpGrid::adapt(const std::vector<std::array<int,3>>& cells_per_dim_vec,
...
-    if(globalActiveLgrs == 0) {
-        Opm::OpmLog::warning("All the LGRs contain only inactive cells.\n");
-    }
-
-    preAdapt();
-    adapt(cells_per_dim_vec, assignRefinedLevel, lgr_name_vec, true, startIJK_vec, endIJK_vec);
-    postAdapt();
-
-    // Only for parallel runs
+     // Only for parallel runs
```
blattms, line 2407: Somehow the indentation is changed here.

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1904,12 +1921,37 @@ bool CpGrid::adapt()
...
+        // GR stands for GLOBAL REFINEMET
+        lgr_name_vec = { "GR" + std::to_string(preAdaptMaxLevel +1) };
+        is_global_refine = true; // parallel
+    }
+#endif
+    if (local_marked_elem_count == current_view_data_-> size(0)) {
+        // GR stands for GLOBAL REFINEMET
+        lgr_name_vec = { "GR" + std::to_string(preAdaptMaxLevel +1) };
+        is_global_refine = true; // sequential
+    }
```
blattms, line 1949: Why do we need this check? If the user does call CpGrid::globalRefine() then it is a global refine. Otherwise I would not care.

tests/cpgrid/addLgrsOnDistributedGrid_test.cpp
```diff
@@ -677,20 +676,66 @@ BOOST_AUTO_TEST_CASE(throw_not_fully_interior_lgr)
...
+
 BOOST_AUTO_TEST_CASE(globalRefine2)
 {
     // Create a grid
     Dune::CpGrid grid;
     const std::array<double, 3> cell_sizes = {1.0, 1.0, 1.0};
     const std::array<int, 3> grid_dim = {4,3,3};
     grid.createCartesian(grid_dim, cell_sizes);
+
+     std::vector<int> parts(36);
```
blattms, line 718: Please fix indentation

tests/cpgrid/addLgrsOnDistributedGrid_test.cpp
```diff
@@ -677,20 +676,66 @@ BOOST_AUTO_TEST_CASE(throw_not_fully_interior_lgr)
...
+            parts[elemIdx] = rank;
+        }
+    }
     // Distribute the grid
     if(grid.comm().size()>1)
     {
         grid.loadBalance();
 
-        BOOST_CHECK_THROW(grid.globalRefine(1), std::logic_error);
+        grid.globalRefine(1);
```
blattms, line 733: How often can we now call globalRefine?

tests/cpgrid/addLgrsOnDistributedGrid_test.cpp
```diff
@@ -677,20 +676,66 @@ BOOST_AUTO_TEST_CASE(throw_not_fully_interior_lgr)
...
+        for (const auto& elemIdx : cells_per_rank[rank]) {
+            parts[elemIdx] = rank;
+        }
+    }
+    // Distribute the grid
+    if(grid.comm().size()>1)
+    {
+        grid.loadBalance();
+
+        grid.globalRefine(1);
```
blattms, line 700: We should test other arguments, too.

blattms (changes requested):

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1260,6 +1255,28 @@ void CpGrid::markElemAssignLevelDetectActiveLgrs(const std::vector<std::array<in
             if(belongsToLevel) {
                 this-> mark(1, element);
                 assignRefinedLevel[element.index()] = level+1; // shifted since starting grid is level 0, and refined grids levels are >= 1.
+            }
+        }
+    }
+}
+
+void CpGrid::detectActiveLgrs(const std::vector<std::array<int,3>>& startIJK_vec,
```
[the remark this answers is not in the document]
aritorto: Good point! The reason is that in adapt(/* with arguments */)  we only call detectActiveLgrs(...) since the elements get marked somewhere else. In addLgrsUpdateLeafView(...), indeed we call both methods, one after the other. 

I'll incorporate the idea of a std::function/lambda to reduce code duplication. Many thanks!


To do: change the name of adapt(/* with arguments */) to something else, not to confuse with DUNE's grid interface adapt().

opm/grid/cpgrid/CpGrid.cpp
aritorto: Good point! Somehow I thought it was needed to check that all cells - in all processes - got marked for refinement. It makes sense that if global refine has been invoked, then that's the case and the check is indeed unnecessary. Thanks!

tests/cpgrid/addLgrsOnDistributedGrid_test.cpp
aritorto: I added a call with argument equal 2, that - for now - throws.

aritorto:

aritorto:
@blattms Thanks again for your review, all comments have been addressed!

opm/grid/cpgrid/CpGrid.cpp
blattms: > Good point! But this piece of code is in adapt() definition, which is not necessary a global refinement. That's why the check is done.

It still seems like this could be done with an optional parameter to adapt that is only true when called from globalRefine.

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1914,12 +1932,37 @@ bool CpGrid::adapt()
...
+        // GR stands for GLOBAL REFINEMET
+        lgr_name_vec = { "GR" + std::to_string(preAdaptMaxLevel +1) };
+        is_global_refine = true; // parallel
+    }
+#endif
+    if (local_marked_elem_count == current_view_data_-> size(0)) {
+        // GR stands for GLOBAL REFINEMET
+        lgr_name_vec = { "GR" + std::to_string(preAdaptMaxLevel +1) };
+        is_global_refine = true; // sequential
+    }
```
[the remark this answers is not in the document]
aritorto: The old version had a previous check for MPI above these lines. I see that it's not enough only to check this after that. In any case, the first suggestion was to remove this check from here and guarantee that the "GR" is set when someone calls global refinement.

In my humble opinion, I would prefer to fix the check here. When the check is correct, then calling adapt having marked all the elements will also end up being called "GR" instead of "LGR". 

It might be nice that the label of the refinement/adaptation reflects if it's a local or a global one, independently of having called globalRefine(). What do you think?

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1941,7 +1984,16 @@ bool CpGrid::adapt(const std::vector<std::array<int,3>>& cells_per_dim_vec,
...
 
     auto& data = currentData(); // data pointed by current_view_data_ (data_ or distributed_data_[if loadBalance() has been invoked before adapt()]).
-      
+
+    // To determine if an LGR is not empty in a given process, we set
+    // lgr_with_at_least_one_active_cell[in that level] to 1 if it contains
+    // at least one active cell, and to 0 otherwise.
+    // "rubissh" denotes an empty vector. The method 'markElemAssignLevelDetectActiveLgrs' skips marking elements and the assignment of their
+    // level when the passed boolean is 'true' ("onlyDetectLgrs").
+    const auto& [lgr_with_at_least_one_active_cell, rubissh] = markElemAssignLevelDetectActiveLgrs(startIJK_vec,
```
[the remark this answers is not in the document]
aritorto: Thanks for your feedback. I thought we agreed on merging your suggestion which was much better and clear. Maybe these comments where there before and I missed them. Sorry for that! 

Now we used https://github.com/aritorto/opm-grid/pull/17  So no rubbish around :)

aritorto:

aritorto:

opm/grid/cpgrid/CpGrid.cpp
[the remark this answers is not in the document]
blattms: Thanks.

blattms (approved):

PR #808, opened 2024-11-28
On loadBalance(...) "addCornerCells" should be called "addEdgeCells"?

aritorto:
In an attempt to assign unique point ids on a distributed grid with LGRs, I invoked loadBalance, enabling the inclusion of cells that share corners with interior cells on each process by setting addCornerCells = true (one of the function arguments in loadBalance). However, instead of adding corner-sharing cells, this configuration primarily adds cells that share edges, in addition to the overlap layer of size 1. As a result, cells that share only a corner are not included, which can impact the uniqueness of point ids in certain cases.

In the test addLgrsOnDistributedGrid_test.cpp, both escenarios are included:

cases where setting addCornerCells=true fix the issue of unique point ids
cases where setting addCornerCells=true is not enough to achieve unique point ids (cells that share only corners - and not edges - are not included).

Replacement for OPM/opm-grid#806

### DIFF
--- a/tests/cpgrid/addLgrsOnDistributedGrid_test.cpp
+++ b/tests/cpgrid/addLgrsOnDistributedGrid_test.cpp
@@ -397,9 +397,12 @@ void refinePatch_and_check(Dune::CpGrid& coarse_grid,
     BOOST_CHECK( static_cast<int>(allGlobalIds_cells.size()) == global_cells_count);
     BOOST_CHECK( allGlobalIds_cells.size() == allGlobalIds_cells_set.size() );
 
-    /** [Bug] Uniqueness of point global ids cannot be checked in general since current code sets overlap layer size equal to 1,
+    /** Uniqueness of point global ids cannot be checked in general since current code sets overlap layer size equal to 1,
         which in particular means that cells that share corners or edges (and not faces) with interior cells are not considered/
-        seen by the process. Therefore, depending how the LGRs are distributed, there may be "multiple ids" for the same points.*/
+        seen by the process. Therefore, depending how the LGRs are distributed, there may be "multiple ids" for the same points.
+        To achieve unique point global ids, IN SOME CASES, we invoke loanBalance( parts, false, true) where parts determines
+        the rank of each cell, false represents ownerFisrt, true represents addCornerCells. The last defaulted argument is
+        overlapLayerSize = 1. */
 
     // Local/Global id sets for level grids (level 0, 1, ..., maxLevel). For level grids, local might differ from global id.
     for (int level = 0; level < coarse_grid.maxLevel() +1; ++level)
@@ -456,6 +459,7 @@ BOOST_AUTO_TEST_CASE(threeLgrs)
     const std::array<int, 3> grid_dim = {10,8,8};
     grid.createCartesian(grid_dim, cell_sizes);
 
+    // Fully interior LGRs
     std::vector<int> parts(640);
     for (int k = 0; k < 8; ++k) {
         for (int j = 0; j < 8; ++j) {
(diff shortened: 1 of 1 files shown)


aritorto:

aritorto:

aritorto:

aritorto:

tests/cpgrid/addLgrsOnDistributedGrid_test.cpp
```diff
@@ -397,9 +397,12 @@ void refinePatch_and_check(Dune::CpGrid& coarse_grid,
     BOOST_CHECK( static_cast<int>(allGlobalIds_cells.size()) == global_cells_count);
     BOOST_CHECK( allGlobalIds_cells.size() == allGlobalIds_cells_set.size() );
 
-    /** [Bug] Uniqueness of point global ids cannot be checked in general since current code sets overlap layer size equal to 1,
+    /** Uniqueness of point global ids cannot be checked in general since current code sets overlap layer size equal to 1,
         which in particular means that cells that share corners or edges (and not faces) with interior cells are not considered/
-        seen by the process. Therefore, depending how the LGRs are distributed, there may be "multiple ids" for the same points.*/
+        seen by the process. Therefore, depending how the LGRs are distributed, there may be "multiple ids" for the same points.
+        To achieve unique point global ids, IN SOME CASES, we invoke loanBalance( parts, false, true) where parts determines
```
lisajulia, line 403: typo: loadBalance

tests/cpgrid/addLgrsOnDistributedGrid_test.cpp
```diff
@@ -545,14 +550,16 @@ BOOST_AUTO_TEST_CASE(atLeastOneLgr_per_process_attempt)
...
     for (int rank = 0; rank < 4; ++rank) {
         for (const auto& elemIdx : cells_per_rank[rank]) {
             parts[elemIdx] = rank;
         }
     }
     if(grid.comm().size()>1)
     {
-        grid.loadBalance(parts);
+        grid.loadBalance(parts); // ownerFirst = false, addCornerCells = false, overlapLayerSize =1
+        // IT's not necessary to change the default values since the LGRs are fully interior.
```
lisajulia, line 562: typo: It

tests/cpgrid/addLgrsOnDistributedGrid_test.cpp
```diff
@@ -397,9 +397,12 @@ void refinePatch_and_check(Dune::CpGrid& coarse_grid,
     BOOST_CHECK( static_cast<int>(allGlobalIds_cells.size()) == global_cells_count);
     BOOST_CHECK( allGlobalIds_cells.size() == allGlobalIds_cells_set.size() );
 
-    /** [Bug] Uniqueness of point global ids cannot be checked in general since current code sets overlap layer size equal to 1,
```
lisajulia, line 400: Did I get that right?
We call grid.loadBalance(parts, false, true) with addCornerCells = true to get unique IDs, correct?
Can you add that to the explanation here?
Otherwise I'd merge this PR :)

lisajulia:
Only minor changes in the documentation :+1:

tests/cpgrid/addLgrsOnDistributedGrid_test.cpp
aritorto: We call grid.loadBalance(parts, false, true) with addCornerCells = true to get, when it's possible, unique IDs for point entities. Unfortunately, 

either the name "addCornerCells" is not describing correctly what it actually does, i.e., add only cells that share edges with interior cells and neglect cells that share only corners - not edges ("addEdgeCells" would be a better name);

or there is a bug on how loadBalance with addCornerCells = true actually adds cells.  

I'll add a comment on this description. Thanks for your feedback! (Delayed reply due to Winter holidays : ) )

aritorto:

PR #809, opened 2024-11-29
Use transmissibilities as edge weights

michal-toth:
With these changes, GraphOfGrid can use transmissibilities for the edge weights if the vector of transm. is available upon the graph's construction. Without transmissibilities, the graph will use uniform edge weights equal to 1.

### DIFF
--- a/opm/grid/GraphOfGrid.cpp
+++ b/opm/grid/GraphOfGrid.cpp
@@ -186,7 +186,7 @@ void GraphOfGrid<Grid>::addWell (const std::set<int>& well, bool checkIntersecti
 }
 
 template<typename Grid>
-void GraphOfGrid<Grid>::addWellBuffer ()
+void GraphOfGrid<Grid>::addNeighboringCellsToWells ()
 {
     // mark all cells that will be added to wells (addding them one
     // by one would require recursive checks for neighboring wells)
--- a/opm/grid/GraphOfGrid.cpp
+++ b/opm/grid/GraphOfGrid.cpp
@@ -29,7 +29,7 @@
 namespace Opm{
 
 template<typename Grid>
-void GraphOfGrid<Grid>::createGraph ()
+void GraphOfGrid<Grid>::createGraph (const double* transmissibilities)
 {
     const auto& rank = grid.comm().rank();
     // load vertices (grid cells) into graph
@@ -57,7 +57,7 @@ void GraphOfGrid<Grid>::createGraph ()
             {
                 continue;
             }
-            WeightType weight = 1; // default edge weight
+            WeightType weight = transmissibilities ? transmissibilities[face] : 1; // default edge weight is 1
             vertex.edges.try_emplace(otherCell,weight);
         }
 
--- a/opm/grid/GraphOfGrid.cpp
+++ b/opm/grid/GraphOfGrid.cpp
@@ -58,10 +58,10 @@ void GraphOfGrid<Grid>::createGraph (const double* transmissibilities)
                 continue;
             }
             WeightType weight = transmissibilities ? transmissibilities[face] : 1; // default edge weight is 1
-            vertex.edges.try_emplace(otherCell,weight);
+            vertex.edges.try_emplace(otherCell, weight);
         }
 
-        graph.try_emplace(gID,vertex);
+        graph.try_emplace(gID, vertex);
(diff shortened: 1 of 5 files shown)


michal-toth:
jenkins build this please

opm/grid/GraphOfGrid.cpp
```diff
@@ -185,5 +185,30 @@ void GraphOfGrid<Grid>::addWell (const std::set<int>& well, bool checkIntersecti
...
+template<typename Grid>
+void GraphOfGrid<Grid>::addWellBuffer ()
+{
+    // mark all cells that will be added to wells (addding them one
+    // by one would require recursive checks for neighboring wells)
+    std::vector<std::set<int>> buffer(wells.size());
+    int i=0;
+    for (auto& w : wells)
+    {
+        buffer[i].insert(*w.begin()); // intersects with its well
```
blattms, line 197: Coud `w` be empty? Then we need to cater for this.

opm/grid/GraphOfGrid.hpp
```diff
@@ -162,9 +162,24 @@ class GraphOfGrid{
         return wells;
     }
 
+    /// \brief Contract a layer of verices around each wells into it
+    ///
+    /// Representing a well by one node guarantees that the well won't
+    /// be split over several processes. Giving the well an extra layer
+    /// of cells distances that well from the subdomain boundary.
+    void addWellBuffer ();
+    void addWellBuffer (int layers)
```
blattms, line 171: It would be nice if the function name would give a better hint about what the functions do.
Please document the other function, too.

opm/grid/GraphOfGrid.hpp
```diff
@@ -162,9 +162,24 @@ class GraphOfGrid{
...
+        for (int i=0; i<layers; ++i)
+        {
+            addWellBuffer();
+        }
+    }
+
 private:
     /// \brief Create a graph representation of the grid
-    void createGraph (); // edge weight=1
+    /// If transmissibilities are not supplied, edge weight=1
```
blattms, line 181: `\brief` went missing in action. Without that there will be no short description in doxygen. Is this intentional.

opm/grid/GraphOfGridWrappers.cpp
```diff
@@ -591,7 +591,7 @@ zoltanPartitioningWithGraphOfGrid(const Dune::CpGrid& grid,
 
     // prepare graph and contract well cells
     // non-root processes have empty grid and no wells
-    GraphOfGrid gog(grid);
+    GraphOfGrid gog(grid,transmissibilities);
```
blattms, line 594: Nitpick: space after the comma please. Also elsewhere. This will will make it a lot easier to read.

opm/grid/GraphOfGrid.cpp
akva2: deref'ing begin() on an empty container invokes undefined behavior.

opm/grid/GraphOfGrid.hpp
michal-toth: `\brief` did not get deleted. I added a second line to the function description.

opm/grid/GraphOfGrid.cpp
michal-toth: begin() can not be dereferenced on an empty container because the for-cycle has zero length.

akva2: you are talking about ```wells```, we are talking about ```w```.

michal-toth: Each well has `size()>1`.
Their function is to keep track of vertices that got contracted and are no longer present in the graph, so each well has a one node (smallest ID) that is still present in the graph and others that got removed.

akva2: so iow; the answer to the initial question is 'no', not 'yes'. ```w``` cannot be empty.

blattms:

opm/grid/GraphOfGrid.hpp
blattms: then the comment is now less brief as all lines are used until the first empty.

michal-toth: I will add an empty line there.

PR #812, opened 2024-12-04
Debug in progress. CARFIN case

aritorto:
Related to OPM/opm-simulators#5772


aritorto:
Replaced by OPM/opm-grid#816

PR #813, opened 2024-12-16
Changes for more resonable partitioning:

bska:
- option to not remove anything from partitioning graph
- added possibility for setting overlap
- changed export list: need testing in particular in parallel (if it is intended to work there)

### DIFF
--- a/opm/grid/CpGrid.hpp
+++ b/opm/grid/CpGrid.hpp
@@ -725,7 +725,8 @@ namespace Dune
         bool loadBalance(int overlapLayers=1,
                          int partitionMethod = Dune::PartitionMethod::zoltan,
                          double imbalanceTol = 1.1,
-                         int level =-1)
+                         int level =-1,
+                         bool addCornerCells = true)
         {
             using std::get;
             return get<0>(scatterGrid(/* edgeWeightMethod = */ defaultTransEdgeWgt,
@@ -734,7 +735,7 @@ namespace Dune
                                       /* possibleFutureConnections = */ {},
                                       /* serialPartitioning = */ false,
                                       /* transmissibilities = */ nullptr,
-                                      /* addCornerCells = */ true,
+                                      addCornerCells,
                                       overlapLayers,
                                       partitionMethod,
                                       imbalanceTol,
--- a/opm/grid/common/GridPartitioning.cpp
+++ b/opm/grid/common/GridPartitioning.cpp
@@ -508,7 +508,8 @@ void addOverlapLayerNoZeroTrans(const CpGrid& grid,
                 {
                     // Note: multiple adds for same process are possible
                     exportList.emplace_back(nb_index, owner, AttributeSet::copy);
-                    exportList.emplace_back(index, cell_part[nb_index],  AttributeSet::copy);
+                    // not need or check if
+                    //exportList.emplace_back(index, cell_part[nb_index],  AttributeSet::copy);
                     if ( recursion_deps>0 )
                     {
                         // Add another layer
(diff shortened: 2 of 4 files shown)


opm/grid/CpGrid.hpp
```diff
@@ -1219,10 +1219,10 @@ namespace Dune
         /// \param overlapLayers The number of layers of cells of the overlap region (default: 1).
         /// \param partitionMethod The method used to partition the grid, one of Dune::PartitionMethod
         /// \warning May only be called once.
-        bool loadBalance(int overlapLayers=1, int partitionMethod = Dune::PartitionMethod::zoltan, double imbalanceTol = 1.1)
+        bool loadBalance(int overlapLayers=1, int partitionMethod = Dune::PartitionMethod::zoltan, double imbalanceTol = 1.1, bool addCorners = true)
```
atgeirr, line 1222: Maybe this is an opportunity to get rid of the default arguments?

opm/grid/common/GridPartitioning.cpp
```diff
@@ -391,7 +391,8 @@ void addOverlapLayer(const CpGrid& grid, int index, const CpGrid::Codim<0>::Enti
                 {
                     // Note: multiple adds for same process are possible
                     exportList.emplace_back(nb_index, owner, AttributeSet::copy);
-                    exportList.emplace_back(index, cell_part[nb_index],  AttributeSet::copy);
+                    // not need or check if
+                    //exportList.emplace_back(index, cell_part[nb_index],  AttributeSet::copy);
```
atgeirr, line 395: Why is this not needed? Should it not be symmetrical?

opm/grid/common/ZoltanPartition.cpp
```diff
@@ -286,7 +286,7 @@ void setDefaultZoltanParameters(Zoltan_Struct* zz) {
     Zoltan_Set_Param(zz, "CHECK_GRAPH", "2");
     Zoltan_Set_Param(zz,"EDGE_WEIGHT_DIM","0");
     Zoltan_Set_Param(zz, "OBJ_WEIGHT_DIM", "0");
-    Zoltan_Set_Param(zz, "PHG_EDGE_SIZE_THRESHOLD", ".35");  /* 0-remove all, 1-remove none */
+    Zoltan_Set_Param(zz, "PHG_EDGE_SIZE_THRESHOLD", "1");  /* 0-remove all, 1-remove none */
```
atgeirr, line 289: What are the observations that favor this change? Have you observed problems with too many of the graph edges being cut? I must admit I do not know how to interpret the threshold: will 0.35 cut 35% of the edges?

atgeirr (commented):
My curiosity was triggered... These changes are likely a good idea but I would like some more information.

opm/grid/common/ZoltanPartition.cpp
akva2: the effect of setting this to 1 is that no (hyper-)edges will be omitted from the graph, even if the number of vertices on the edge is a large amount of the total amount of vertices in the graph (used to be 35%). I believe this is intended to detect all-to-all-ish connections that would be a problem for the partitioning algorithm, since all partitions would be penalized for this cut.

opm/grid/CpGrid.hpp
hnil: Yes. But probably that can be done separately.

opm/grid/common/GridPartitioning.cpp
hnil: I do not understand this fully, but this will be done for all processes so this will be added when the other partition is processed anyway.

blattms:
This is a very brief description. It is a bit hard to understand those bullet points. Please make it more detailed and explain the improvements.

PR #817, opened 2024-12-20
Fix type bug int->double.

atgeirr:
Value used is just 1 with the current code.

### DIFF
--- a/opm/grid/common/ZoltanPartition.cpp
+++ b/opm/grid/common/ZoltanPartition.cpp
@@ -66,7 +66,7 @@ makeImportAndExportLists(const Dune::CpGrid& cpgrid,
     std::vector<std::tuple<int,int,char,int>> myImportList;
     assert(rank==root || numExport==0);
     assert(rank!=root || numImport==0);
-    int buffer = 1.05;
+    constexpr double buffer = 1.05;
     // all cells on root are added to its export and its import list
     // myImportList will be expanded if distributed wells are not allowed and well cells are moved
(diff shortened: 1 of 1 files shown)


atgeirr:
jenkins build this please

aritorto:
It seems I cannot approve this PR, so I just left a thumbs up

opm/grid/common/ZoltanPartition.cpp
```diff
@@ -66,7 +66,7 @@ makeImportAndExportLists(const Dune::CpGrid& cpgrid,
     std::vector<std::tuple<int,int,char,int>> myImportList;
     assert(rank==root || numExport==0);
     assert(rank!=root || numImport==0);
-    int buffer = 1.05;
+    const double buffer = 1.05;
```
akva2, line 69: can be constexpr, feel free to ignore.

bska:
> Value used is just 1 with the current code.

"Almost always `auto`"

akva2:
@aritorto you should be able to approve any pr, even on stuff you have not reason to approve, so it's probably just the github interface to do so that confuses? if you go to files changed tab on top, do you not have a review button?

aritorto:
@akva2 I see it now, thanks for guiding to its location!

opm/grid/common/ZoltanPartition.cpp
atgeirr: Changed to constexpr.

atgeirr:
jenkins build this please

PR #797, opened 2024-11-14
cpgrid preprocess: use unsigned for face_ptrs

akva2:
should allow slightly larger models. this is very drafty

PR #818, opened 2025-01-11
Preprocess unsigned in grid processing for larger grid sizes.

kjetilly:
Draft pull request [based on PR 797 from akva2](https://github.com/OPM/opm-grid/pull/797). This is currently draft to iron out all new warnings.

### DIFF
--- a/opm/grid/CellQuadrature.cpp
+++ b/opm/grid/CellQuadrature.cpp
@@ -91,7 +91,7 @@ int CellQuadrature::numQuadPts() const
     }
     assert(grid_.dimensions == 3);
     int sumnodes = 0;
-    for (int hf = grid_.cell_facepos[cell_]; hf < grid_.cell_facepos[cell_ + 1]; ++hf) {
+    for (unsigned hf = grid_.cell_facepos[cell_]; hf < grid_.cell_facepos[cell_ + 1]; ++hf) {
         const int face = grid_.cell_faces[hf];
         sumnodes += grid_.face_nodepos[face + 1] - grid_.face_nodepos[face];
     }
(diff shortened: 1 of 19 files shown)


kjetilly:
Jenkins build this please

bska:
jenkins build this please

akva2 (approved):
Looks fine to me. The grid_size_t looks a bit half baked but it doesn't harm.

GitPaean:
@kjetilly  is `grid_size_t` intended for more usage? it is not used much in the current form of the PR.

kjetilly:
> @kjetilly is `grid_size_t` intended for more usage? it is not used much in the current form of the PR.

This is used mainly in the ops-upscaling pr: https://github.com/OPM/opm-upscaling/pull/396

Sorry, this was a bit unclear in the communication in this PR.

atgeirr:
While I do not think it is a bad idea to merge this PR (and downstream) maybe it is a better idea to go all the way once here? Why not just replace all occurrences of `int` with `std::int64_t` (and include `cstdint`), and we are done with this problem for the lifetime of everyone? Sure, we may use a bit more memory in the corner point grid processing, but it will be (mostly) thrown away, and if that is not where the peak is it should be ok?

akva2:
my main reasoning for not doing so was we wouldn't be done with the problem in any sensible definition of done. there are a lot of other spots such as CpGrid class and simulators which cannot in any way handle 64bit indices.

atgeirr:
> cannot in any way handle 64bit indices.

Do you mean "cannot handle it without significantly worse memory peaks"?

akva2:
CpGrid is using int internally to represent things, while things are mostly using unsigned in simulator codes. Even if we could handle the memory cost it would not help much.

PR #819, opened 2025-01-13
Preprocess int64

kjetilly:
Just a draft pull-request to measure the effects of moving the whole grid processing to int64


kjetilly:
This probably will not go through, the unsigned and int32-usage really manifests itself throughout the code, but keeping this PR for now...

PR #822, opened 2025-01-20
Add ElementChunks helper for thread-parallel grid iteration

atgeirr:
This is mostly to see if we can have a helper for grid iteration that makes chunked multi-threaded iteration easy.

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -160,6 +160,7 @@ list (APPEND TEST_DATA_FILES
 list (APPEND EXAMPLE_SOURCE_FILES
   examples/finitevolume/finitevolume.cc
   examples/mirror_grid.cpp
+  examples/griditer.cpp
   )
 
 # programs listed here will not only be compiled, but also marked for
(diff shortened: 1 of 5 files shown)


atgeirr:
Updated with a somewhat hacky ChunkIterator class, but was able to get rid of the extra storage of the iterator pairs, and also allowing direct range-for over the helper object.

atgeirr:
This should be ready for review now. It contains:
 - A new class ElementChunks to make it very easy to do thread-parallel loops over grids.
 - Improvements to createThreadIterators(), making it more general, while also keeping an overload with the GridView-specific interface.
 - A new function createChunkIterators() that is even more general, and handles some corner cases better.
 - An example executable demonstrating and comparing these facilities.

We may want to remove the original createThreadIterators() and replace it with using ElementChunks, but for now I kept both so as to not require downstream changes.

atgeirr:
jenkins build this please

atgeirr:
I see I did change the API (int& to std::size_t&), changing it back.

atgeirr:
jenkins build this please

CMakeLists_files.cmake
```diff
@@ -160,6 +160,7 @@ list (APPEND TEST_DATA_FILES
 list (APPEND EXAMPLE_SOURCE_FILES
   examples/finitevolume/finitevolume.cc
   examples/mirror_grid.cpp
+  examples/griditer.cpp
```
akva2, line 163: unconditional openmp usage. either make it conditional in build system or adjust the test.

akva2 (approved):
Looks good, just one small issue

CMakeLists_files.cmake
atgeirr: Adjusted the test so I did not have to mess with the build system.

atgeirr:
jenkins build this please

atgeirr:
I will also add a unit test to this.

atgeirr:
Needs one more round due to a warning in the added test...

atgeirr:
Approved, green and no new warnings: merging.

PR #823, opened 2025-01-20
Changing IdType to int64_t

kjetilly:
This pull request changes the `IdType` to `int64_t` to avoid some overflow errors seen previously.

No change in opm-simulators is needed.

### DIFF
--- a/opm/grid/LookUpData.hh
+++ b/opm/grid/LookUpData.hh
@@ -83,7 +83,11 @@ public:
     ///
     /// \param [in] elementOrIdx  Element or index of the element in the leaf grid view.
     /// \param [in] fieldProp     Vector (indexable collection) of field properties.
-    auto operator()(const auto& elementOrIdx, const auto& fieldProp) const;
+    ///
+    /// \tparam ElementOrIndex the type of the element or index passed in.
+    /// \tparam FieldProperties the type of the field properties vector. Should be an \c std::vector like type.
+    template<class ElementOrIndex, class FieldProperties>
+    auto operator()(const ElementOrIndex& elementOrIdx, const FieldProperties& fieldProp) const;
 
     /// \brief: Get field property of type double from field properties manager by name.
     std::vector<double> assignFieldPropsDoubleOnLeaf(const FieldPropsManager& fieldPropsManager,
@@ -108,12 +112,45 @@ public:
                      const std::string& propString,
                      const ElemOrIndex& elemOrIndex) const;
 
-    /// \brief: Return the same element index for all grids different from CpGrid.
+    /// @brief calls \c getFieldPropIdx<Grid, ElementType>(elem) 
+    template<typename ElementType>
+    auto getFieldPropIdx(const ElementType& elem) const;
+
+    /// \brief Return the index used for retrieving field properties, depending on whether
+    ///        the grid is a CpGrid or a general grid, and whether the element is in an LGR.
     ///
-    /// \tparam     GridType    Auxiliary type to overload the method, distinguishing general grids from CpGrid, with std::enable_if.
-    ///                         Default: GridType = Grid.
-    template<typename GridType = Grid>
-    auto getFieldPropIdx(const auto& elem) const;
+    /// This method determines the appropriate index under two main conditions:
+    ///
+    /// - **Non-CpGrid**: Returns the same element index that was passed in. If an entity is
+    ///   provided, it uses \c elemMapper_ to retrieve the index. The function asserts
+    ///   that \c maxLevel() == 0 (i.e., no LGR support) for non-CpGrid grids.
+    ///
+    /// - **CpGrid**: Depending on whether \c isFieldPropInLgr_ is \c true and the element
+    ///   (or index) is in a refined level (> 0), the returned index is for the equivalent
+    ///   LGR cell; otherwise, the index of the origin (level 0) cell is returned.
+    ///
+    /// \tparam GridType
+    ///     Auxiliary type used to specialize the method for CpGrid
+    ///     vs. other grids. If \c GridType = \c Dune::CpGrid, local grid refinements (LGR)
+    ///     are considered; otherwise, LGR is not supported.
+    /// 
+    /// \tparam ElementType
+    ///     The type of the element or index passed in. 
+    ///
+    /// \param elementOrIndex
+    ///     An integral cell index or a grid entity (e.g., \c Dune::cpgrid::Entity<0>).
+    ///
+    /// \return
+    ///     The integer index to be used for looking up the relevant field properties. For
+    ///     non-CpGrid grids, this is the same index or one retrieved from \c elemMapper_.
+    ///     For CpGrid grids, it is either the LGR-based index or the origin index,
+    ///     depending on \c isFieldPropInLgr_ and whether the cell is refined.
+    ///
+    /// \note
+    ///     If \c GridType is not CpGrid, the method will assert that \c maxLevel() == 0,
+    ///     since local grid refinements are only supported for CpGrid.
+    template<typename GridType, typename ElementType>
+    auto getFieldPropIdx(const ElementType& elem) const;
 
 protected:
     const GridView& gridView_;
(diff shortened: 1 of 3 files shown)


atgeirr:
The extra template argument that causes the need for a downstream PR: is there a way to avoid needing to specify the argument, having it deduced instead?

Otherwise I think this looks like a reasonable direction, but I have not done a thorough review as this is still in draft mode. I think replacing `is_same_v<EntityType, unsigned int>` with `is_integral_v<EntityType>` is a good thing, this was already noticed as something to fix.

kjetilly:
> The extra template argument that causes the need for a downstream PR: is there a way to avoid needing to specify the argument, having it deduced instead?

@atgeirr I think this can be avoided. One solution would be to exchange the order of the template arguments, ie. having `IdType` the last argument to allow for automatic deduction, but that would not make it compatible with the overloaded version taking non-integral types (which already is `<EntityType, GridType>`).

Another solution would be to make the functions have a version taking `auto` arguments that then dispatches the function with the correct template arguments, or using said `auto` function and just using a `if constexpr` to deal with the different cases. The latter would probably entail less code and might be easier to read. It would go something like:

```C++
template<typename Grid, typename GridView>
template<class GridType>
int Opm::LookUpData<Grid,GridView>::getFieldPropIdx(const auto& elem) const
{ // ignoring the check for GridType to keep it simple, but that could either be handled as is or with a constexpr if.
    if constexpr (!std::is_integral_v(decltype(elem)) {
        static_assert(std::is_same_v<Grid,GridType>);
        static_assert(std::is_same_v<EntityType,Dune::cpgrid::Entity<0>>);
        if (isFieldPropInLgr_ && elem.level()) { // level > 0 == true ; level == 0 == false
            // In case some LGRs do not have refined field properties, the next line need to be modified.
            return elem.getLevelElem().index();
        }
        else {
            return elem.getOrigin().index();
        }
    } else {
      static_assert(std::is_same_v<Grid,GridType>);
      // Check there are no LGRs. LGRs (level>0) only supported for CpGrid.
      assert(gridView_.grid().maxLevel() == 0);
      return elem;
  }
}
```

Now, this would also introduce change in `opm-simulators` I think, since it would remove the template argument from the non-integer version. 

In terms of minimisation of change, I think exchanging the order of the template arguments would be the winner, but that would lead to an inconsistency.

atgeirr:
Now I think that maybe changing the interface would be better in the end, if we could avoid having to specify (more than one) template arguments, despite a few downstream changes. I think there are confusingly many variants of these functions, which I think is because there are several dimensions here: CpGrid or not, Entity or not. Could we do with just a single one using `if constexpr`?

kjetilly:
@atgeirr I've added a sketch with a simplified function signature. Note that I still have to add documentation to the new functions to properly document this. The function signatures are a bit cleaner, though I would argue that their definitions might be slightly more cumbersome to read (a lot of `if else`)

kjetilly:
So it seems a fortunate side effect of this edition is that no change in opm-simulators is needed.

kjetilly:
Jenkins build this please

atgeirr:
I think this looks like a good reduction of complexity. @aritorto as the original author of this code, do you agree?

aritorto:
Indeed, nice simplification/improvement. Why is it still in draft mode? Now running a simulation with CARFIN keyword has been incorporated into master (OPM/opm-simulators#5903). This is (one test) where lookupdata would do the not trivial thing. So, just in case, we could check - one more time - that nothing breaks

aritorto:
jenkins build this please

kjetilly:
@aritorto the main reason it's still in draft is that I haven't completed the new documentation of the functions. I will do that now and take it out of draft.

akva2:
this requires c++-20.

atgeirr:
> this requires c++-20.

Which part of it?

akva2:
see jenkins console output and gcc tells you all you need to know.
```
/var/lib/jenkins/workspace/opm-grid-PR-builder/mpi/install/include/opm/grid/LookUpData.hh:86:27: warning: use of ‘auto’ in parameter declaration only available with ‘-std=c++20’ or ‘-fconcepts’
   86 |     auto operator()(const auto& elementOrIdx, const auto& fieldProp) const;
```

kjetilly:
Jenkins build this please

kjetilly:
Jenkins build this please

opm/grid/LookUpData.hh
```diff
@@ -345,13 +330,14 @@ std::vector<IntType> Opm::LookUpData<Grid,GridView>::assignFieldPropsIntOnLeaf(c
                                                                                const bool& needsTranslation,
                                                                                std::function<void(IntType, int)> valueCheck) const
 {
+    using IndexType = typename Dune::MultipleCodimMultipleGeomTypeMapper<GridView>::Index;
```
aritorto, line 333: according to a warning, this seems not to be used. If that's the case, it could be removed

opm/grid/LookUpData.hh
```diff
@@ -379,85 +365,66 @@ int Opm::LookUpData<Grid,GridView>::fieldPropInt(const FieldPropsManager& fieldP
...
-    if (isFieldPropInLgr_ && elem.level()) { // level > 0 == true ; level == 0 == false
-        // In case some LGRs do not have refined field properties, the next line need to be modified.
-        return elem.getLevelElem().index();
-    }
-    else {
-        return elem.getOrigin().index();
+    constexpr static bool isIntegral = std::is_integral_v<IndexType>;
+    if constexpr (std::is_same_v<GridType, Dune::CpGrid>) {
+        if constexpr (isIntegral) {
+            static_assert(std::is_same_v<Grid,GridType>);
```
aritorto, line 380: the check std::is_same_v<Grid, GridType> is present in all the if/else branches, shall we move it outside? e.g. below isIntegral definition. Or is there any reason (to have it like it is now)  I'm missing?

opm/grid/LookUpData.hh
```diff
@@ -513,47 +480,41 @@ int Opm::LookUpCartesianData<Grid,GridView>::fieldPropInt(const FieldPropsManage
...
-    static_assert(std::is_same_v<Grid,GridType>);
-    if (isFieldPropInLgr_ && elem.level()) { // level == 0 false; level > 0 true
-        return elem.getLevelCartesianIdx();
-    }
-    else {
-        return cartMapper_-> cartesianIndex(this->elemMapper_.index(elem));
+    constexpr static bool isIntegral = std::is_integral_v<ElementType>;
+    if constexpr (std::is_same_v<GridType, Dune::CpGrid>) {
+        if constexpr (isIntegral) {
+            static_assert(std::is_same_v<Grid,GridType>);
```
aritorto, line 495: same comment/question here, if it can be moved before the first if constexpr, below is Integral definiiton

aritorto (commented):
It's a nice simplification. I had only a couple of comments/questions

opm/grid/LookUpData.hh
kjetilly: You are right, this can be removed and is now removed.

opm/grid/LookUpData.hh
kjetilly: Absolutely correct, collected them one level higher up in the if-hierarchy.

opm/grid/LookUpData.hh
kjetilly: Also correct here, moved to higher level if.

kjetilly:
Jenkins build this please

aritorto (approved):
Nice improvements!

blattms:
The companion PR in opm-simulators is still WIP. I guess that needs to be merged with one?

blattms:
Jenkins build this please

kjetilly:
> The companion PR in opm-simulators is still WIP. I guess that needs to be merged with one?

This was a bit unclear since I did not close the opm-simulators PR, but in its current version, this PR does not need any change outside of opm-grid. I close the opm-simulators PR now since it is no longer needed.

opm/grid/LookUpData.hh
```diff
@@ -75,22 +75,19 @@ public:
...
-    template<typename EntityType, typename FieldPropType>
-    typename std::enable_if_t<!std::is_same_v<EntityType, unsigned int>, FieldPropType>
-    operator()(const EntityType& elem, const std::vector<FieldPropType>& fieldProp) const;
+    /// \param [in] elementOrIdx  Element or index of the element in the leaf grid view.
+    /// \param [in] fieldProp     Vector (indexable collection) of field properties.
+    ///
+    /// \tparam ElementOrIndex the type of the element or index passed in.
+    /// \tparam FieldProperties the type of the field properties vector. Should be an \c std::vector like type.
+    template<class ElementOrIndex, class FieldProperties>
+    auto operator()(const ElementOrIndex& elementOrIdx, const FieldProperties& fieldProp) const;
```
blattms, line 90: Still C++20, it seems...

blattms (changes requested):

opm/grid/LookUpData.hh
blattms: Or maybe not, sorry.

PR #826, opened 2025-01-24
Janitoring: quell some static analyzer issues

akva2:
Mostly trivialities.

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -74,17 +74,6 @@ if (opm-common_FOUND)
 endif()
 
 # originally generated with the command:
-# find tests/not-unit/ -name \*.cpp -o \*.cc
-list (APPEND ATTIC_FILES
-  attic/partition_test.cpp
-# attic/dumux_test.cpp
-  attic/mapper_test.cpp
-  attic/buildcpgrid_test.cpp
-# attic/cpgrid_test.cpp
-  attic/check_grid_normals.cpp
-  )
-
-# originally generated with the command:
 # find tests -name '*.cpp' -a ! -wholename '*/not-unit/*' -printf '\t%p\n' | sort
 list (APPEND TEST_SOURCE_FILES
   tests/p2pcommunicator_test.cc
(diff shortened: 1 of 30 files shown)


akva2:
jenkins build this please

tests/cpgrid/zoltan_test.cpp
```diff
@@ -89,7 +89,7 @@ BOOST_AUTO_TEST_CASE(zoltan)
     int m_argc = boost::unit_test::framework::master_test_suite().argc;
     char** m_argv = boost::unit_test::framework::master_test_suite().argv;
     {
-        auto& inst = Dune::MPIHelper::instance(m_argc, m_argv);
+        const auto& inst = Dune::MPIHelper::instance(m_argc, m_argv);
```
bska, line 92: If you're changing this anyway, maybe we could just tag this as `[[maybe_unused]]` to remove the cast below?

CMakeLists_files.cmake
```diff
@@ -76,12 +76,12 @@ endif()
...
+#  attic/partition_test.cpp
 # attic/dumux_test.cpp
-  attic/mapper_test.cpp
-  attic/buildcpgrid_test.cpp
+#  attic/mapper_test.cpp
+#  attic/buildcpgrid_test.cpp
 # attic/cpgrid_test.cpp
-  attic/check_grid_normals.cpp
+#  attic/check_grid_normals.cpp
   )
```
bska, line 85: Maybe we could just remove this list altogether?

bska (commented):
I'll defer to others to do a proper review though the C++ parts here look good to me.  Just a couple of minor things that we might also do while here.

CMakeLists_files.cmake
akva2: we can.

bska (approved):
Thanks! The C++ parts look good to me now, but as I said I'll defer to others to comment on whether the changes interfere with any ongoing work.

@aritorto, @blattms : Please have a look here.

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -3427,7 +3451,7 @@ void CpGrid::populateLeafGridFaces(Dune::cpgrid::EntityVariableBase<cpgrid::Geom
                         // Get the index of this corner with respect to the greatest marked element index, using find instead of count.
                         lastAppearanceLgr_lgrEquivCorner = vanishedRefinedCorner_to_itsLastAppearance.at({elemLgr, elemLgrCorn});
                         while (elemLgrAndElemLgrCorner_to_adaptedCorner.find( lastAppearanceLgr_lgrEquivCorner ) == elemLgrAndElemLgrCorner_to_adaptedCorner.end()) {
-                            auto& tempLgr_lgrCorner = lastAppearanceLgr_lgrEquivCorner;
+                            const auto& tempLgr_lgrCorner = lastAppearanceLgr_lgrEquivCorner;
```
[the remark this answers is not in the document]
akva2: i respectfully disagree. the underlying value might change, but we are not mutating the data through the reference. if we need the reference, we can move it below the assignment though to avoid confusion.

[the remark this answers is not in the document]
aritorto: You're right. I just didn't/don't remember if there was a reason why I didn't add const back then, or it was just "an accident". It seems to be the latter. Thanks for the improvements!

bska:
Given the most recent discussion here, I take it that the functional parts are approved as well.  Is that right, @aritorto?

aritorto:
Yes, all good here

PR #824, opened 2025-01-22
Use only transmissibilities when allow-distributed-wells=true

lisajulia:

PR #827, opened 2025-01-28
AllowDistributedWells with (default) partitioning method 3

michal-toth:
Switch off the merging of the well cells when `--allowDistributedWells==true`.

This goes against the nature of the partitioner that was designed to allow cell merging. However, with this flag set to true the partitioning results are still different<sup>1</sup> to `partitioning-method=1` (normal Zoltan). For this reason I think it is better to switch off the main feature of this partitioner instead of delegating the work to the partitioning method 1 when the `--allowDistributedWells==true`.

<sup>1</sup> From my limited experimenting, the partitioning of method 3 has fewer overlap cells than method 1. It is also slower on a small grid but gets faster on a big grid (limited testing! one grid with >1.5 million cells) and needs a negligible amount of extra memory (less than 1% extra, and there are procedures before and after partitioning that demand more memory - it is far from peak requirements).


michal-toth:
First wait how the PR #824 will be resolved.

PR #829, opened 2025-02-04
Compute COORD values for LGRs

aritorto:
Taking into account also inactive cells. 

It needs OPM/opm-grid#828.

Not relevant for the manual.

### DIFF
--- a/opm/grid/cpgrid/CpGridUtilities.cpp
+++ b/opm/grid/cpgrid/CpGridUtilities.cpp
@@ -20,10 +20,14 @@
 #include <opm/grid/cpgrid/CpGridUtilities.hpp>
 #include <opm/grid/cpgrid/LevelCartesianIndexMapper.hpp>
 
+#include <algorithm>
 #include <array>
+#include <limits>
+#include <map>
 #include <string>
 #include <tuple>
 #include <unordered_map>
+#include <utility>
 #include <vector>
 
 namespace Opm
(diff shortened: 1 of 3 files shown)


aritorto:
jenkins build this please

opm/grid/cpgrid/CpGridUtilities.cpp
```diff
@@ -65,4 +69,100 @@ lgrIJK(const Dune::CpGrid& grid, const std::string& lgr_name)
...
+}
+
+void processPillars(int i, int j, int nx,
+                    const Dune::cpgrid::Entity<0>& topElem,
+                    const Dune::cpgrid::Entity<0>& bottomElem,
+                    std::vector<std::array<double, 6>>& lgrCOORD)
+{
+    setPillarCoordinates(i, j, nx, 4, 0,  0, topElem, bottomElem, lgrCOORD);
+    setPillarCoordinates(i, j, nx, 5, 1,  1, topElem, bottomElem, lgrCOORD);
+    setPillarCoordinates(i, j, nx, 6, 2,  2, topElem, bottomElem, lgrCOORD);
```
blattms, line 164: Please add a comment why we need a few of these twice

opm/grid/cpgrid/CpGridUtilities.cpp
```diff
@@ -65,4 +69,100 @@ lgrIJK(const Dune::CpGrid& grid, const std::string& lgr_name)
...
+            const auto& bottomElem = Dune::cpgrid::Entity<0>(levelGrid, bottomElemIdx, true);
+            const auto& topElem = Dune::cpgrid::Entity<0>(levelGrid, topElemIdx, true);
+
+            // Recall that a cell has 8 corners:
+            //        6 --- 7
+            //       /     /   TOP FACE
+            //      4 --- 5
+            //        2 --- 3
+            //       /     /   BOTTOM FACE
+            //      0 --- 1
```
blattms, line 127: Maybe move this comment to the next function?

blattms (changes requested):
It took me some time to understand this. It is complex but that is actually needed.
This is good work.

Besides the comments in the text, I would suggest two things (not sure they are actually needed though):
- Check that not all cells are inactive
- We will have rather degenerated lines (they are actually points at (max_double, max_double, max_double)) if there are neighboring columns of inactive cells. If that creates problems we could just replace these with a copy of a neighboring line or any other one.
Those changes could be postponed, but we should keep them in mind.

opm/grid/cpgrid/CpGridUtilities.cpp
aritorto: Added!

opm/grid/cpgrid/CpGridUtilities.cpp
aritorto: Moved!

aritorto:
Thanks for your feedback @blattms ! All your comments have been addressed, except from:

> * We will have rather degenerated lines (they are actually points at (max_double, max_double, max_double)) if there are neighboring columns of inactive cells. If that creates problems we could just replace these with a copy of a neighboring line or any other one.

I'll postpone that for now

aritorto:
jenkins build this please

opm/grid/cpgrid/CpGridUtilities.cpp
```diff
@@ -20,10 +20,15 @@
 #include <opm/grid/cpgrid/CpGridUtilities.hpp>
 #include <opm/grid/cpgrid/LevelCartesianIndexMapper.hpp>
 
+#include <algorithm>
 #include <array>
+#include <limits>
+#include <map>
```
blattms, line 26:
```suggestion
```

aritorto:
jenkins build this please

PR #830, opened 2025-02-05
Do not throw in getLevelElem() and remove getEquivLevelElem()

aritorto:
getLevelElem() throws an exception if the grid view is not a leaf, whereas getEquivLevelElem() always returns an entity regardless of the grid view state.

This PR merges both methods into a single one that no longer throws an exception.


aritorto:
jenkins build this please

blattms:
thanks a lot for the simplification.

PR #834, opened 2025-02-10
Some static analyzer fixes

akva2:
Mostly trivial but maybe @aritorto  wants to have a look at some of the variable removals.

### DIFF
--- a/opm/grid/GraphOfGridWrappers.cpp
+++ b/opm/grid/GraphOfGridWrappers.cpp
@@ -517,7 +517,6 @@ zoltanPartitioningWithGraphOfGrid(const Dune::CpGrid& grid,
                                   const double zoltanImbalanceTol,
                                   const std::map<std::string, std::string>& params)
 {
-    int rc = ZOLTAN_OK - 1;
     float ver = 0;
     struct Zoltan_Struct *zz;
     int changes, numGidEntries, numLidEntries, numImport, numExport;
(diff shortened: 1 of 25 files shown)


akva2:
jenkins build this please

opm/grid/GraphOfGridWrappers.cpp
```diff
@@ -517,20 +517,21 @@ zoltanPartitioningWithGraphOfGrid(const Dune::CpGrid& grid,
...
-    if ( rc != ZOLTAN_OK )
-    {
+    int rc = Zoltan_Initialize(argc, argv, &ver);
+    if (rc != ZOLTAN_OK) {
         OPM_THROW(std::runtime_error, "Could not initialize Zoltan!");
     }
+    zz = Zoltan_Create(cc);
+    if (rc != ZOLTAN_OK) {
+        OPM_THROW(std::runtime_error, "Could not create Zoltan data structures!");
+    }
```
bska, line 534: The `rc` here is the one from the `Zoltan_Initialize()` call above, at least if there's no global reference anywhere, so I don't see how it could be anything other than `ZOLTAN_OK` in this instance.  Maybe you wanted `zz == nullptr` instead?

opm/grid/common/ZoltanPartition.cpp
```diff
@@ -307,20 +307,21 @@ zoltanGraphPartitionGridOnRoot(const CpGrid& cpgrid,
...
-    if ( rc != ZOLTAN_OK )
-    {
+    int rc = Zoltan_Initialize(argc, argv, &ver);
+    if (rc != ZOLTAN_OK) {
         OPM_THROW(std::runtime_error, "Could not initialize Zoltan!");
     }
+    zz = Zoltan_Create(cc);
+    if (rc != ZOLTAN_OK) {
+        OPM_THROW(std::runtime_error, "Could not create Zoltan data structures!");
+    }
```
bska, line 324: This `rc` is the one from `Zoltan_Initialize()` above and won't be anything other than `ZOLTAN_OK` if we get here.  Did you mean `zz == nullptr` instead?

bska (commented):
I'll defer to @blattms and @aritorto to do the full review here.  I just noticed a couple of potential issues in the Zoltan checks.

opm/grid/GraphOfGridWrappers.cpp
akva2: yes! thanks, that was stupid half-work.

aritorto:
The removed variables were not needed, indeed. (For tag/normal of refined faces, the mutable variables were used. How the partition is done - via a parts vector - was not relevant for global refinement). Great!

aritorto (approved):

PR #840, opened 2025-02-19
Make zoltanGoG default partitioner if grid has wells

michal-toth:
Recently (about 1 month ago?), `zoltanGoG` partitioner replaced `zoltan` as the default partitioner in opm-simulators. This commit reflects this change and makes `zoltanGoG` the default partitioner when grid has wells. To my knowledge, this commit has no effect on running flow (where opm-simulators' settings are used).

However, we noticed that changing the default partitioner to `zoltanGoG` in loadbalancing methods which do not have wells (which makes `zoltanGoG` basically equivalent to `zoltan` partitioner) makes the test `addLgrsOnDistributedGrid_test_parallel` fail. Investigation is ongoing.


michal-toth:
jenkins build this please

aritorto (approved):

aritorto:
Thanks a lot!

PR #842, opened 2025-02-21
Add and require partition support to ElementChunks.

atgeirr:
Existing code would always include all cells, i.e. Dune::Partitions::all, the new code enables you to only do Dune::Partitions::interior for example. It also requires this argument (no default) to force the user to think through what is the correct partition to iterate over.

Draft PR for now, as it will require downstream changes.

### DIFF
--- a/opm/grid/utility/ElementChunks.hpp
+++ b/opm/grid/utility/ElementChunks.hpp
@@ -22,6 +22,7 @@
 #include <opm/grid/utility/createThreadIterators.hpp>
 
 #include <cstddef>
+#include <iterator>
 #include <utility>
 #include <vector>
 
@@ -71,18 +72,21 @@ namespace Opm
 ///         // Do something else with elem
 ///     }
 /// }
-template <class GridView>
+template <class GridView, class PartitionSet>
 class ElementChunks
 {
 private:
-    using Iter = typename GridView::template Codim<0>::Iterator;
+    using Iter = decltype(std::begin(elements(std::declval<const GridView&>(), PartitionSet())));
     using Storage = std::vector<Iter>;
     using StorageIter = decltype(Storage().cbegin());
 
 public:
-    ElementChunks(const GridView& gv, const std::size_t num_chunks)
+    ElementChunks(const GridView& gv,
+                  const PartitionSet included_partition,
+                  const std::size_t num_chunks)
     {
-        grid_chunk_iterators_ = Opm::createChunkIterators(elements(gv), gv.size(0), num_chunks);
+        grid_chunk_iterators_
+            = Opm::createChunkIterators(elements(gv, included_partition), gv.size(0), num_chunks);
     }
 
     struct Chunk
--- a/tests/test_elementchunks.cpp
+++ b/tests/test_elementchunks.cpp
@@ -32,6 +32,8 @@
 
 #include <opm/grid/CpGrid.hpp>
 
+#include <dune/grid/common/partitionset.hh>
+
 struct Fixture
 {
     Fixture()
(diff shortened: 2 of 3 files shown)


atgeirr:
jenkins build this please

atgeirr:
jenkins build this opm-simulators=6027 please

opm/grid/utility/ElementChunks.hpp
```diff
@@ -71,18 +71,18 @@ namespace Opm
...
 {
 private:
-    using Iter = typename GridView::template Codim<0>::Iterator;
+    using Iter = decltype(std::begin(elements(std::declval<GridView>(), PartitionSet())));
     using Storage = std::vector<Iter>;
     using StorageIter = decltype(Storage().cbegin());
 
 public:
-    ElementChunks(const GridView& gv, const std::size_t num_chunks)
+    ElementChunks(const GridView& gv, const PartitionSet included_partition, const std::size_t num_chunks)
```
akva2, line 83: long line

opm/grid/utility/ElementChunks.hpp
```diff
@@ -71,18 +71,18 @@ namespace Opm
...
     {
-        grid_chunk_iterators_ = Opm::createChunkIterators(elements(gv), gv.size(0), num_chunks);
+        grid_chunk_iterators_ = Opm::createChunkIterators(elements(gv, included_partition), gv.size(0), num_chunks);
```
akva2, line 85: long line

tests/test_elementchunks.cpp
```diff
@@ -60,9 +60,12 @@ std::vector<int> countChunks(const EC& chunks)
     return counts;
 }
 
-void testCase(const GV& gv, const std::size_t num_chunks, const std::vector<int>& expected)
+using GV = Dune::CpGrid::LeafGridView;
+
+template <class PartitionSet>
+void testCase(const GV& gv, const PartitionSet part, const std::size_t num_chunks, const std::vector<int>& expected)
```
akva2, line 66: long line

akva2 (approved):
Looks good, some cosmetic suggestions

opm/grid/utility/ElementChunks.hpp
```diff
@@ -71,18 +71,21 @@ namespace Opm
 ///         // Do something else with elem
 ///     }
 /// }
-template <class GridView>
+template <class GridView, class PartitionSet>
 class ElementChunks
 {
 private:
-    using Iter = typename GridView::template Codim<0>::Iterator;
+    using Iter = decltype(std::begin(elements(std::declval<GridView>(), PartitionSet())));
```
bska, line 78: You need to be a little more careful here.  `Declval<>()` will (pretend to) create the **literal** type of its template argument&ndash;in this case `GridView`&ndash;whereas the `Iter` type in the `Chunk`s is the result of traversing a `const GridView&`.  The two iterator types may or may not be directly interchangeable, so I'd recommend writing this as
```C++
std::declval<const GridView&>()
```
instead.  As an aside, you should probably include `<iterator>` to bring in a declaration of `std::begin()` even though we **technically** get that from `<vector>` as well.

atgeirr:
Long lines shortened, force pushed.

Maybe we should reduce the ColumnLimit in .clang-format from 120 to 100?

atgeirr:
jenkins build this opm-simulators=6027 please

bska (commented):
Nice!

Just one small subtlety regarding how we use `std::declval<>()`.

opm/grid/utility/ElementChunks.hpp
atgeirr: Done.

atgeirr:
jenkins build this opm-simulators=6027 please

PR #843, opened 2025-02-21
Distribute level zero grid even though leaf grid view has LGRs 

aritorto:
To be cleaned-up soon


aritorto:
Replaced by OPM/opm-grid#845

PR #837, opened 2025-02-13
Faltten lgr COORD to facilitate writing output 

aritorto:
To easier invoke some methods developed in OPM/opm-common#4478 related to output files for LGRs, this PR flattens lgr COORD into a vector<double>. 
Additionally, it separates the tests for lgrIJK() and lgrCOORDZCORN() methods. 

Not relevant for the Reference Manual

PR #844, opened 2025-02-21
Bugfix: case with no wells

michal-toth:
Fix the partitioning for the case without any wells. The problem was dereferencing of the null pointer in the call of the constructor of WellConnections. Moreover, the sorting of the lists was skipped.

I noticed this bug while working on the small PR #840 but a partial fix already appeared in the PR #837.

The PR #840 made the `zoltanGoG` default partitioner of the `CpGrid` for the cases when wells are used, following the changes in opm-simulators. The `CpGrid`'s `loadBalance` method without wells keep `zoltan` as the default partitioner. (There is no need to construct the graph representation of the grid to give the partitioner the same information. There is one small difference, namely that `zoltanGoG` stores global IDs in an unordered map so the partitioning tool loads them in a different order. That often leads to a different partitioning of the same quality.)

### DIFF
--- a/opm/grid/GraphOfGrid.cpp
+++ b/opm/grid/GraphOfGrid.cpp
@@ -39,7 +39,7 @@ void GraphOfGrid<Grid>::createGraph (const double* transmissibilities,
     WeightType logMinTransm = std::numeric_limits<WeightType>::max();
     if (transmissibilities && edgeWeightMethod==Dune::EdgeWeightMethod::logTransEdgeWgt)
     {
-        for (int face = 0; face < getGrid().numFaces(); ++face)
+        for (int face = 0; face < grid.numFaces(); ++face)
         {
             WeightType transm = transmissibilities[face];
             if (transm > 0 && transm < logMinTransm)
--- a/opm/grid/GraphOfGridWrappers.cpp
+++ b/opm/grid/GraphOfGridWrappers.cpp
@@ -475,6 +475,11 @@ makeImportAndExportLists(const GraphOfGrid<Dune::CpGrid>& gog,
         auto wellRanks = getWellRanks(gIDtoRank, wellConnections);
         parallel_wells = wellsOnThisRank(*wells, wellRanks, cc, root);
     }
+    else
+    {
+        std::sort(myExportList.begin(), myExportList.end());
+        std::sort(myImportList.begin(), myImportList.end());
+    }
     return std::make_tuple( std::move(gIDtoRank),
                             std::move(parallel_wells),
                             std::move(myExportList),
@@ -550,8 +555,8 @@ zoltanPartitioningWithGraphOfGrid(const Dune::CpGrid& grid,
     // non-root processes have empty grid and no wells
     GraphOfGrid gog(grid, transmissibilities, edgeWeightMethod);
     assert(gog.size()==0 || !partitionIsEmpty);
-    auto wellConnections=partitionIsEmpty ? Dune::cpgrid::WellConnections()
-                                          : Dune::cpgrid::WellConnections(*wells, possibleFutureConnections, grid);
+    auto wellConnections = partitionIsEmpty || !wells ? Dune::cpgrid::WellConnections()
+                                                      : Dune::cpgrid::WellConnections(*wells, possibleFutureConnections, grid);
     addWellConnections(gog, wellConnections);
     gog.addNeighboringCellsToWells(layers);
 
@@ -723,7 +728,7 @@ zoltanSerialPartitioningWithGraphOfGrid(const Dune::CpGrid& grid,
     std::vector<std::tuple<int, int, char, int>> myImportList;
     std::vector<std::vector<int>> exportedCells;
     auto wellConnections = partitionIsEmpty || !wells ? Dune::cpgrid::WellConnections()
-                                            : Dune::cpgrid::WellConnections(*wells, possibleFutureConnections, grid);
+                                                      : Dune::cpgrid::WellConnections(*wells, possibleFutureConnections, grid);
 
     if (cc.rank() == root) {
         std::tie(rc, gIDtoRank) = applySerialZoltan(grid,


blattms (approved):

michal-toth:
I compared the performance of `map` and `unordered_map` and I will add the table here once I find it. The `map` took two times more time than `unordered_map` (in combination with `map` for `EdgeList` - the list of neighbors for each cell).

I would also add that the stability you mention is related to the implementation of Zoltan graph partitioner, which is a third party software. Methods `zoltan` and `zoltanGoG` give it the same information albeit permuted differently. The graph algorithm thus starts from a different point.
Changing `unordered_map` to `map` would indeed remove the differences. The ordering of cell neighbors would be different regardless of using `map` or `unordered_map` in `zoltanGoG`'s `EdgeList` but that seemingly did not affect results.

atgeirr:
> Please add an indication to the version used in the table.

Here are the corresponding versions:
```
Flow Version     =  2025.04-pre (b6f23f6fe)
Flow Version     =  2025.04-pre (b6f23f6fe)
Flow Version     =  2025.04-pre (b6f23f6fe)
Flow Version     =  2025.04-pre (b6f23f6fe)
Flow Version     =  2025.04-pre (01784904d*)
Flow Version     =  2025.04-pre (2477dd98d*)
Flow Version     =  2025.04-pre (2477dd98d*)
Flow Version     =  2025.04-pre (a0614bcd9)
Flow Version     =  2025.04-pre (a0614bcd9)
Flow Version     =  2025.04-pre (a0614bcd9)
Flow Version     =  2025.04-pre (a0614bcd9)
Flow Version     =  2025.04-pre (a0614bcd9)
Flow Version     =  2025.04-pre (1e426a6de)
Flow Version     =  2025.04-pre (9843ca8)
Flow Version     =  2025.04-pre (9843ca8*)
Flow Version     =  2025.04-pre (9843ca8*)
Flow Version     =  2025.04-pre (9843ca8*)
Flow Version     =  2025.04-pre (9843ca8*)
Flow Version     =  2025.04-pre (9843ca8*)
Flow Version     =  2025.04-pre (9843ca8*)
Flow Version     =  2025.04-pre (9843ca8*)
Flow Version     =  2025.04-pre (1e426a6de*)
Flow Version     =  2025.04-pre (8c204b217*)
Flow Version     =  2025.04-pre (c26f07abd)
Flow Version     =  2025.04-pre (25cd79c04)
Flow Version     =  2025.04-pre (dec5aab87)
Flow Version     =  2025.04-pre (4b688e594)
Flow Version     =  2025.04-pre (3c5b0ae07)
Flow Version     =  2025.04-pre (3c5b0ae07)
Flow Version     =  2025.04-pre (3c5b0ae07*)
Flow Version     =  2025.04-pre (36928f87c2)
```
Not sure if it is very useful though!

michal-toth:
This table compares the performance of `map` and `unordered_map` in `zoltanGoG` from PR https://github.com/OPM/opm-grid/pull/811. We use them at two places: for storing the grid cells ("graph") and the list of neighboring edges for every cell.

| graph | edges | Test: | Easy | Medium | Hard 1 | Hard 2 |
| --- | --- | --- | --- | --- | --- | --- |
| un_map | un_map | | 15162 | 14196684 | 169561690 | 185451570 |
| un_map | map | | 13910 | 12138575 | 163632397 | 165122661 |
| map | un_map | | 16852 | 15992087 | 191048511 | 189642649 |
| map | map | | 13460 | 28522892 | 331562668 | 315050777|

atgeirr:
> This table compares the performance

Thanks! But what are the numbers?

michal-toth:
> > This table compares the performance
> 
> Thanks! But what are the numbers?

It is the execution time of the partitioning process. It includes both the creation of the graph representation of the grid in the `zoltanGoG` and the partitioning.

atgeirr:
> In microseconds but each grid is processed 100 times during this time.

So the hard2 case goes from 1.85 to 3.15 seconds time spent for a single processing. It is a little surprising to me that the difference is so large, but as this is a one-time cost I think maybe it is acceptable. Approximately how many cells are in the hard  cases, and how many partitions are created?

blattms:
@atgeirr it is a bit hard to work with your provided information. Some of the runs are with commits that are not part of master.  Below I am trying to merge your information on the relevant part:
```
   sum         44431           14179         58610 2025.04-pre (25cd79c04) Tue Jan 14 09:45:26 2025 +0100
   sum         44431           14780         59211 2025.04-pre (dec5aab87) NOT FOUND in master!
   sum         44431           14780         59211 2025.04-pre (4b688e594) Fri Feb 14 20:40:31 2025 +0100
   sum         44431           13974         58405 2025.04-pre (3c5b0ae07) NOT FOUND in master!
   sum         44431           13974         58405 2025.04-pre (3c5b0ae07) NOT FOUND in master!
   sum         44431           13974         58405 2025.04-pre (3c5b0ae07) NOT FOUND in master!
   sum         44431           13974         58405 2025.04-pre (36928f87c2) Fri Feb 21 12:34:05 2025 +0100
```

BTW 25cd79c04 already has the new partitioning as default (Merge commit f59262d4036 on Mon Jan 13 13:40:12 2025 +0100), but the changes that you are seeing are after that.

I actually don't follow your thought here.

> It seems likely that this may lead to different partitioning between runs, maybe even immediate repeat runs, but certainly between computers, and changing over time as the simulator is changed.

What am I missing here? This should only change if the global id of a cell changes, the order of the grid traversal changes (no such change), or the type of the global id. The latter has been changed in #823

michal-toth:
Both Hard 1 and Hard 2 have about 1.5 million cells. I think I used 10 partitions, but it could be 8.
The extra time is still negligible. The partitioning algorithm takes much longer that is still a fraction of the time spent on parsing the input files.

atgeirr:
> Both Hard 1 and Hard 2 have about 1.5 million cells. I think I used 10 partitions, but it could be 8.
> The extra time is still negligible. The partitioning algorithm takes much longer that is still a fraction of the time spent on parsing the input files.

In that case I think we should make the change, thanks for the data. If we later find we need to optimize this part more, I would expect that the usage pattern is "build the thing once, then access it many times" which would let us use a sorted vector for fast access.

michal-toth:
> > It seems likely that this may lead to different partitioning between runs, maybe even immediate repeat runs, but certainly between computers, and changing over time as the simulator is changed.
> 
> What am I missing here? This should only change if the global id of a cell changes, the order of the grid traversal changes (no such change), or the type of the global id. The latter has been changed in #823

The partitioning will remain the same. Neither the Zoltan graph algorithm nor the `unordered_map` hash function use randomization. (The hash function (at least in gcc) uses bits of keys which gives consistent results.)

atgeirr:
Thanks for your thoughts. I also discussed this with Markus yesterday, and came to the conclusion that you are probably right: unordered_map is likely not to blame for the changes in partitioning. It is likely that the more recent change was instead due to #823 as suggested above.

My main point is that we need to ensure repeat runs produce exactly identical results, as I am sure you agree. But I no longer believe the current partitioning endangers this. Thanks for the discussion!

PR #847, opened 2025-03-03
[bug fix]  Grid global id set should see all level cell ids

aritorto:
During the review of OPM/opm-grid#845, a bug was identified in the behavior of global_id_set_ptr_ in CpGrid. 

Without this PR, calling grid.globalIdSet().id(element) was possible only for elements on the leaf grid view or a refined level grid (excluding the level zero grid). This issue was observed in both serial and parallel settings.
This PR fixes that (inserting all the level id sets before adding LGRs).


aritorto:
jenkins build this please

aritorto:
jenkins build this please

blattms (approved):
Thanks a lot for the fix.

PR #848, opened 2025-03-04
[fix] Use maybe_unused instead of wrong static_cast

aritorto:
correction from OPM/opm-grid#841

PR #849, opened 2025-03-04
[Bugfix] work around buggy cpgrid copy

blattms:
Continuation of #848 to really fix the tests.

:flushed: We have a rather buggy copy constructor in CpGrid it seems. Seems like the pointer current_data_ is just copied nd if the object that we copy from ist destroyed that pointer is dangling.

Another proper fix is upcoming.


blattms:

blattms:
Shameless self-merge :grimacing:

PR #850, opened 2025-03-04
[bugfix] Remove raw pointers for the views in CpGrid

blattms:
This pointers were dangling afte a grid was copied and the original one removed. Fortunatley, we did not do this in productive code.

This approach uses a boolean to see whether the current view is distributed or not. This way we can use default copy constructors.

Unfortunately. there are a lot of branches now. Let's see what that costs.

### DIFF
--- a/opm/grid/CpGrid.hpp
+++ b/opm/grid/CpGrid.hpp
@@ -316,6 +316,11 @@ namespace Dune
         /// @brief Returns either data_ or distributed_data_(if non empty).
         std::vector<std::shared_ptr<Dune::cpgrid::CpGridData>>& currentData();
 
+        /// @brief Returns either data_ or distributed_data_(if non empty).
+        const Dune::cpgrid::CpGridData& currentLeafData() const;
+
+        /// @brief Returns either data_ or distributed_data_(if non empty).
+        Dune::cpgrid::CpGridData& currentLeafData();
         /// @brief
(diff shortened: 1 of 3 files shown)


blattms:
benchmark please

blattms:

bska:
Please use your own fork instead of the main upstream repository.

blattms:
Sorry that this happened again. Somehow I do have different remotes but two of the are pointing here.

bska:
> Somehow I do have different remotes but two of the are pointing here.

I see.  For what it's worth, I typically equip my local repositories with something like
```sh
git remote set-url --push upstream PUSH-DISALLOWED
```
for my "upstream" repository (local name of the main OPM repository).  That way way I get a diagnostic of the form
```
fatal: 'PUSH-DISALLOWED' does not appear to be a git repository
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```
if I inadvertently try to run
```sh
git push upstream some-branch
```

Of course, we sometimes need to push directly to the main repository but in that case I try to do such work in a dedicated local repository to keep a bit of separation.

blattms:
Thanks, I usually do the same thing (ssh only for the local one and only for pushing. But somehow I accidentally did use the OPM URL instead of my blattms one. I blame copy&paste (with missing the changes in the URL) and sloppy scripting on my side. :disappointed:

PR #852, opened 2025-03-05
Avoid copying CpGrid in LgrCheck methods

aritorto:
This PR refactors createGridAndAddLgrs(...) to eliminate unnecessary CpGrid copies, preventing potential issues in parallel settings.

While this change does not introduce direct improvements, it addresses a potential risk similar to the one reported in [OPM/opm-grid#845](https://github.com/OPM/opm-grid/issues/845), where using copied CpGrid instances in a parallel context could lead to unexpected behavior.

This refactor reduces the risk of such issues in the future and allows reusing existing code.


blattms:
jenkins build this please

PR #853, opened 2025-03-06
Fixes for NVCC

kjetilly:
It seems like NVCC complains about `Entity` being defined twice, this PR fixes that issue.


kjetilly:
Jenkins build this please

bska (approved):

PR #854, opened 2025-03-10
Synchronize cell ids 

aritorto:
Synchronizes cell global ids across processes after load balancing and refinement.
        
LGRs (Local Grid Refinements) can be added either in the undistributed view first and then in the distributed view, or vice versa. The method CpGrid::syncCellIds()  ensures consistency by rewriting the global cell ids in the distributed view using the corresponding ids from the undistributed view.

Note: vertex ids are nor rewritten in this PR. 


aritorto:

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -2485,143 +2485,258 @@ bool CpGrid::adapt(const std::vector<std::array<int,3>>& cells_per_dim_vec,
...
+                         lgr_with_at_least_one_active_cell);
+    }
+
+
+    // Print total amount of cells on the adapted grid
+    Opm::OpmLog::info(std::to_string(markedElem_count) + " elements have been marked (in " + std::to_string(comm().rank()) + " rank).\n");
+    Opm::OpmLog::info(std::to_string(levels)  + " (new) refined level grid(s) (in " + std::to_string(comm().rank()) + " rank).\n");
+    Opm::OpmLog::info(std::to_string(cell_count)  + " total cells on the leaf grid view (in " + std::to_string(comm().rank()) + " rank).\n");
+
+    return preAdapt();
```
blattms, line 2499: Are you sure this should be here?
This is the adapt method.

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -2485,143 +2485,258 @@ bool CpGrid::adapt(const std::vector<std::array<int,3>>& cells_per_dim_vec,
...
+
+    auto size = comm().max(parentToFirstChildGlobalIds.size());
+
+    switchToDistributedView();
+
+    std::vector<int> firstChildGlobalIds(size);
+    if (comm().rank() == 0) {
+        firstChildGlobalIds = parentToFirstChildGlobalIds;
+    }
+    comm().broadcast(firstChildGlobalIds.data(), firstChildGlobalIds.size(), 0);
```
blattms, line 2676: Can't we omit this container just use parentToFirstChildGlobalIds directly?

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -2485,143 +2485,258 @@ bool CpGrid::adapt(const std::vector<std::array<int,3>>& cells_per_dim_vec,
...
+        }
+    }
+
+    // Syncronize ids via data Handle
+    const auto& parent_to_children = currentData().front()->getParentToChildren();
+    ParentToChildrenCellGlobalIdHandle parentToChildrenCellGlobalId_handle(parent_to_children,
+                                                                           syncCellIds);
+    currentData().front()->communicate(parentToChildrenCellGlobalId_handle,
+                                       Dune::InteriorBorder_All_Interface,
+                                       Dune::ForwardCommunication );
```
blattms, line 2707: This would not be needed if iterate over all elements in the loop above (and not only interior). I think we should do it like this.

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -2485,143 +2485,258 @@ bool CpGrid::adapt(const std::vector<std::array<int,3>>& cells_per_dim_vec,
...
+                                                       vertexIds);
+        }
+    }
+
+    for (int level = 1; level <=maxLevel; ++level) {
+        populateCellIndexSetRefinedGrid(level);
+        currentData()[level]->computeCellPartitionType();
+
+        currentData()[level]->computePointPartitionType();
+        // currentData()[level]->computeCommunicationInterfaces(currentData()[level]->size(3));
```
blattms, line 2725: This does not needed as we do not change partitions.

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -2485,143 +2485,258 @@ bool CpGrid::adapt(const std::vector<std::array<int,3>>& cells_per_dim_vec,
...
+    }
+
+    // Insert the new id sets into the grid global_id_set_ptr_
+    for (int level = 1; level <=maxLevel; ++level) {
+        this->global_id_set_ptr_->insertIdSet(*(*current_data_)[level]);
+    }
```
blattms, line 2731: Please merge this with the previous loop.

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -2485,143 +2485,258 @@ bool CpGrid::adapt(const std::vector<std::array<int,3>>& cells_per_dim_vec,
...
+    populateLeafGlobalIdSet();
+    this->global_id_set_ptr_->insertIdSet(*currentData().back());
+
+    currentData().back()->computeCellPartitionType();
+    currentData().back()->computePointPartitionType();
```
blattms, line 2736: Not needed?

blattms (changes requested):
Thanks. This is  good first version but we probably will need to refactor it to a version without any broadcasts of large containers later.

opm/grid/cpgrid/CpGrid.cpp
aritorto: Nice change!

opm/grid/cpgrid/CpGrid.cpp
aritorto: Thanks for the improvement. parentToFristChildGlobalIds does the job too

opm/grid/cpgrid/CpGrid.cpp
aritorto: Fixed now!

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -2485,143 +2485,258 @@ bool CpGrid::adapt(const std::vector<std::array<int,3>>& cells_per_dim_vec,
...
+    current_data_->back()->computeCommunicationInterfaces(current_data_->back()->size(3));
+    assert(static_cast<std::size_t>(current_data_->back()->cellIndexSet().size()) == static_cast<std::size_t>(current_data_->back()->size(0)) );
 #endif
-    } // end-if-comm().size()>1
+}
 
+void CpGrid::getFirstChildGlobalIds([[maybe_unused]] std::vector<int>& parentToFirstChildGlobalIds)
+{
+#if HAVE_MPI
+    switchToGlobalView();
```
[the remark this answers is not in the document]
aritorto: I actually use the fact that changes currentData() , to make sure in the lines below that currentDAta points at data

aritorto:
> We should concentrate on one direction (data_ -> distributed_data_) for now and make that clear from the function names and not call currentData() in the added code.

Except from this point, all comments have been addressed. In my understanding, switching views gives the code more flexibility in the sense that synchronization of cell ids (from data_ to distributed_data_) is possible for both situations:

(A) 1. add LGRs in  global grid, 2. switch to distributed  view, 3. add LGRs in distributed view,

or

(B) 1. distribute grid, 2. add LGRs in distributed grid, 3. switch to global view, 4. add LGRs in global view.

Maybe there is a way of refactoring the added code that keeps this flexibility and remove the use of switching views. 

Thanks for the review, @blattms!

opm/grid/cpgrid/CpGrid.cpp
blattms: This was for the next iteration. Please ignore.

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -2485,143 +2485,235 @@ bool CpGrid::adapt(const std::vector<std::array<int,3>>& cells_per_dim_vec,
...
+    // Re-assign new cell global ids for all refined level grids
+    std::vector<int> faceIds; // empty for all
+    for (int level = 1; level <= maxLevel; ++level) {
+        if(currentData()[level]->size(0)) { // Check if LGR is active in currect process.
+            auto vertexIds = currentData()[level]->global_id_set_-> getMapping<3>();
+            currentData()[level]->global_id_set_->swap(syncCellIds[level-1],
+                                                       faceIds,
+                                                       vertexIds);
+
+            populateCellIndexSetRefinedGrid(level);
```
blattms, line 2706: Sorry, I missed this one during the first review. It is a bit unclear to me how this behaves:

- The method does not clear the interface before adding. I am not sure what happens with our approach, might be one of these
  - We have multiple entries with the same global id
  - We only have one, but cannot be sure which one of the two wins.
- RemoteIndices using the ParallelIndexSet will be broken. Hence we will need to rebuild/update them. Not sure whether there some and whether we use those later.

I guess the easy fix is to clear the index set in the function before adding and then maybe rebuild remote indices.
In the long run this should have its own way of doing things as we actually need to do less then in populateCellIndexSetRefinedGrid.

aritorto: Thanks for noticing, indeed, the size of cell index set for levels was not correct (since there was no clearing of it before the assignment of new sync ids). This [commit](https://github.com/OPM/opm-grid/pull/854/commits/0f88e5666b0cd014176db09c616f4074e91ec1a2) fixes it.

aritorto:

aritorto:

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1491,10 +1491,15 @@ void CpGrid::selectWinnerPointIds([[maybe_unused]] std::vector<std::vector<int>>
...
+    const auto& level_data = currentData()[level];
+    const auto& level_global_id_set =  level_data->global_id_set_;
+    auto& level_index_set =  level_data->cellIndexSet();
 
-    level_index_set.beginResize();
+    // Redefine the level cell index set by creating a new ParallelIndexSet 
+    // and then swapping it with the existing one.
+    ParallelIndexSet new_level_index_set; 
+
+    new_level_index_set.beginResize();
```
blattms, line 1502: Too bad that ParallelIndexSet is such an old fashioned and incomplete container (no clear method, no swap, no move ...). Hence I think swapping at the end of this function is probably more costly than
```
level_index_set = ParallelIndexSet();
level_index_set.beginResize();
```
without creating the additional new_level_indexset.

I might be missing something here and maybe you already tried that?

blattms: Wait a minute. Now I am sceptical about my own hypothesis..

blattms: So I need to correct myself. As we only use std containers or custome containers that use standard containers and do not implement copy or assignment. The containers should even have autogenerated move constructors and swap should be efficient.

We still need extra space because of two containers.

aritorto:
Thanks again for the review!

aritorto:

PR #858, opened 2025-03-27
Changes to support 2.10

blattms:
This is  a rebased version of #758 but on my local fork. I will close the other one.


akva2:
Here we seem to be keeping some guards (first commit) that should not be necessary right?

akva2:
Doesn't look like it from the total diff.

blattms:
Sorry,. I was too fast with triggering merge in opm-common and then this got needed to fix the build.

PR #860, opened 2025-03-27
Refactor adapt_cpgrid_test, add global_refine_test

aritorto:
This PR refactors and improves adapt_cpgrid_test.cpp. Previously, the test took ~13 seconds and included checks for globalRefine(). Now, adapt_cpgrid_test.cpp takes less than a second and a separate test has been introduced specifically for globalRefine().

Additionally, redundant test cases that exhibited similar behavior to existing tests have been removed. The test names have also been cleaned up and improved, with emphasis on describing behavior clearly.

### DIFF
--- a/opm/grid/cpgrid/CpGrid.cpp
+++ b/opm/grid/cpgrid/CpGrid.cpp
@@ -706,8 +706,7 @@ void CpGrid::computeGlobalCellLgr(const int& level, const std::array<int,3>& sta
 
 void CpGrid::computeGlobalCellLeafGridViewWithLgrs(std::vector<int>& global_cell_leaf)
 {
-    for (const auto& element: elements(leafGridView()))
-    {
+    for (const auto& element: elements(leafGridView())) {
         // When refine via CpGrid::addLgrsUpdateGridView(/*...*/), level-grid to lookup global_cell_ is equal to level-zero-grid
         // In the context of allowed nested refinement, we lookup for the oldest ancestor, also belonging to level-zero-grid.
         auto ancestor = element.getOrigin();
@@ -1040,9 +1039,9 @@ void CpGrid::globalRefine (int refCount)
     // This is the case when a coarse cell has not been marked for refinement, but at least one of its neighboring cells
     // got refined. Therefore, the coarse face that they share got replaced by refined-faces. In this case, we do not
     // support yet global refinement.
-    if(currentData().size() >1) {
+    if(this->maxLevel()) {
         bool isOnlyGlobalRefined = true;
-        for (std::size_t level = 0; level < currentData().size(); ++level) {
+        for (int level = 0; level < this->maxLevel(); ++level) {
             // When the grid has been refined only via global refinement, i.e., each cell has been refined into 2x2x2 children cells,
             // then the quotient between the total amount of two consecutive refined level grids is equal to 8 = 2x2x2.
             isOnlyGlobalRefined = isOnlyGlobalRefined && ( (currentData()[level+1]->size(0)) / (currentData()[level]->size(0)) == 8 );
@@ -1052,13 +1051,11 @@ void CpGrid::globalRefine (int refCount)
         }
     }
     if (refCount>0) {
-        const std::vector<std::array<int,3>>& cells_per_dim_vec = {{2,2,2}}; // Arbitrary chosen values.
         for (int refinedLevel = 0; refinedLevel < refCount; ++refinedLevel) {
 
             std::vector<int> assignRefinedLevel(current_view_data_-> size(0));
             const auto& preAdaptMaxLevel = this ->maxLevel();
             std::vector<std::string> lgr_name_vec = { "GR" + std::to_string(preAdaptMaxLevel +1) };
-            bool isCARFIN = true;
             const std::array<int,3>& endIJK = currentData().back()->logicalCartesianSize();
 
             for(const auto& element: elements(this-> leafGridView())) {
@@ -1068,7 +1065,7 @@ void CpGrid::globalRefine (int refCount)
             }
 
             preAdapt();
-            adapt(cells_per_dim_vec, assignRefinedLevel, lgr_name_vec, isCARFIN, {{0,0,0}}, {endIJK});
+            adapt(/* cells_per_dim_vec = */ {{2,2,2}}, assignRefinedLevel, lgr_name_vec, {{0,0,0}}, {endIJK});
             postAdapt();
         }
     }
@@ -1505,8 +1502,7 @@ void CpGrid::populateCellIndexSetRefinedGrid([[maybe_unused]] int level)
     //
     // For refined level grids, we check if the parent cell is fully interior. Then, its children won't be seen
     // by any other process. Therefore, the boolean is set to false.
-    const auto& elements =  Dune::elements(levelGridView(level));
-    for(const auto& element : elements) {
+    for(const auto& element : elements(levelGridView(level))) {
         if ( element.partitionType() == InteriorEntity) {
 
             // Check if it has an overlap neighbor
@@ -1959,6 +1955,15 @@ template cpgrid::Entity<1> createEntity(const CpGrid&, int, bool); // needed in 
 
 bool CpGrid::mark(int refCount, const cpgrid::Entity<0>& element)
 {
+    // Throw if element has a neighboring cell from a different level.
+    // E.g., a coarse cell touching the boundary of an LGR, or
+    // a refined cell with a coarser/finner neighboring cell. 
+    const auto& intersections = Dune::intersections(leafGridView(), element);
+    for (const auto& intersection : intersections){
+        if (intersection.neighbor() && (intersection.outside().level() != element.level()) && (element.level()==0))
+             OPM_THROW(std::logic_error, "Refinement of cells at LGR boundaries is not supported, yet.");
+    }
+    
     // For serial run, mark elements also in the level they were born.
     if(currentData().size()>1) {
         // Mark element in its level
@@ -1989,6 +1994,10 @@ bool CpGrid::preAdapt()
 
 bool CpGrid::adapt()
 {
+    if(!preAdapt()) { // marked cells set can be empty
+        return false; // the grid does not change at all.
+    }
+
     const std::vector<std::array<int,3>>& cells_per_dim_vec = {{2,2,2}}; // Arbitrary chosen values.
     std::vector<int> assignRefinedLevel(current_view_data_-> size(0));
     const auto& preAdaptMaxLevel = this ->maxLevel();
@@ -2022,7 +2031,7 @@ bool CpGrid::adapt()
     }
     if (is_global_refine) { // parallel or sequential
         const std::array<int,3>& endIJK = currentData().back()->logicalCartesianSize();
-        return this->adapt(cells_per_dim_vec, assignRefinedLevel, lgr_name_vec, is_global_refine, {{0,0,0}}, {endIJK});
+        return this->adapt(cells_per_dim_vec, assignRefinedLevel, lgr_name_vec, {{0,0,0}}, {endIJK});
     }
     return this-> adapt(cells_per_dim_vec, assignRefinedLevel, lgr_name_vec);
 }
@@ -2030,7 +2039,6 @@ bool CpGrid::adapt()
 bool CpGrid::adapt(const std::vector<std::array<int,3>>& cells_per_dim_vec,
                    const std::vector<int>& assignRefinedLevel,
                    const std::vector<std::string>& lgr_name_vec,
-                   bool isCARFIN,
                    const std::vector<std::array<int,3>>& startIJK_vec,
                    const std::vector<std::array<int,3>>& endIJK_vec)
 {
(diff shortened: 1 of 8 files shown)


blattms:

aritorto:

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -2466,7 +2514,7 @@ bool CpGrid::adapt(const std::vector<std::array<int,3>>& cells_per_dim_vec,
     }
 
     std::vector<int> global_cell_leaf( data[levels + preAdaptMaxLevel +1]->size(0));
-    computeGlobalCellLeafGridViewWithLgrs(global_cell_leaf);
+    computeGlobalCellLeafGridViewWithLgrs(global_cell_leaf);
```
blattms, line 2517: Please, remove the trailing whitespace.

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -1959,6 +1959,15 @@ template cpgrid::Entity<1> createEntity(const CpGrid&, int, bool); // needed in
 
 bool CpGrid::mark(int refCount, const cpgrid::Entity<0>& element)
 {
+    // Throw if element has a neighboring cell from a different level.
```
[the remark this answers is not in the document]
aritorto: If cell0 belongs to level 0 (coarse cell) and cell1 to level1 (refined cell), the coarse element has more than 6 faces and marking it for refinement is "okay" , but refine code only knows - for now - how to refine cells with 8 corners and 6 faces.

I see that this throw does not represent exactly this, since it's also throwing when refined cells from different levels share a face. Neighboring LGRs with consistent number of subdivisions on the boundary that they share are allowed. So I adjust this throw for the first case I mentioned (level zero vs refined cell, i.e., cells with more than 6 faces)

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -706,8 +706,8 @@ void CpGrid::computeGlobalCellLgr(const int& level, const std::array<int,3>& sta
 
 void CpGrid::computeGlobalCellLeafGridViewWithLgrs(std::vector<int>& global_cell_leaf)
 {
-    for (const auto& element: elements(leafGridView()))
-    {
+    const auto& elements = Dune::elements(leafGridView());
+    for (const auto& element: elements) {
```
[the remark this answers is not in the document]
aritorto: We don't need it, but this avoids calling elements(leafGridView()) once per iteration. I might not be a significant improvement though...

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -2022,7 +2035,7 @@ bool CpGrid::adapt()
     }
     if (is_global_refine) { // parallel or sequential
         const std::array<int,3>& endIJK = currentData().back()->logicalCartesianSize();
-        return this->adapt(cells_per_dim_vec, assignRefinedLevel, lgr_name_vec, is_global_refine, {{0,0,0}}, {endIJK});
+        return this->adapt(cells_per_dim_vec, assignRefinedLevel, lgr_name_vec, /*isCARFIN =*/ true, {{0,0,0}}, {endIJK});
```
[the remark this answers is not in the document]
aritorto: Yes! I'll remove it.

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -2038,15 +2051,31 @@ bool CpGrid::adapt(const std::vector<std::array<int,3>>& cells_per_dim_vec,
     assert( static_cast<int>(assignRefinedLevel.size()) == current_view_data_->size(0));
     assert(cells_per_dim_vec.size() == lgr_name_vec.size());
 
+    bool all_elements_marked = std::all_of(assignRefinedLevel.begin(), assignRefinedLevel.end(),
+                                  [&](const auto& other_level) {
+                                      return other_level == assignRefinedLevel.front();
+                                  });
+    auto all_elements_marked_global = comm().min(all_elements_marked);
```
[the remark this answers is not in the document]
aritorto: I see your point. The intention of this check  (and the one below that takes care of startIJK and endIJK being "begging  and end of the grid") is to identify when the entire (global) grid is about to be refined, even though globalRefine() might have not been invoked. Knowing this makes a difference when defining logical_cartesian_size_ of the leaf grid view, for example. 

Maybe we can avoid communication here with a bool param isGlobal that ill be true when globalRefine is called.

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -2446,7 +2486,15 @@ bool CpGrid::adapt(const std::vector<std::array<int,3>>& cells_per_dim_vec,
     (*data[levels + preAdaptMaxLevel +1]).leaf_to_level_cells_ =  leaf_to_level_cells;
     (*data[levels + preAdaptMaxLevel +1]).index_set_ = std::make_unique<cpgrid::IndexSet>(data[levels + preAdaptMaxLevel +1]->size(0),
                                                                                           data[levels + preAdaptMaxLevel +1]->size(3));
-    (*data[levels + preAdaptMaxLevel +1]).logical_cartesian_size_ =  (*data[0]).logical_cartesian_size_;
+    if (all_elements_marked_global || isGlobalRefine) {
+        assert(cells_per_dim_vec.size() == 1);
```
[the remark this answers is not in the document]
aritorto: Luckily, it does not. But thanks to this comment, I added some tests in global_refine_test to check if that multiple calls were supported. There was a bug (not related to this assert) related to how it was determined if the grid contains only global refined levels or not. Now that is fixed. 

This assert should help when all the elements have been marked for refinement (all assigned to the same level grid), or start/end ijk represent the entire gri. Therefore, the size of the vector cells_per_dim_vec must be only 1. The difference with calling directly globalRefine is that the default number of subdivisions in each direction is 2x2x2, and here we do not have that restriction.

tests/cpgrid/LgrChecks.hpp
```diff
@@ -168,61 +216,94 @@ void Opm::checkEqMinMaxGlobalCellLevelZeroAndLeaf(const std::vector<int>& global
...
-    const auto endItFather = element.father().hend(grid.maxLevel());
-    // If itFather != endItFather and !element.father().isLeaf() (if dristibuted_data_ is empty).
-    BOOST_CHECK( itFather != endItFather );
-    Opm::checkAverageCenterAndVolume(itFather, endItFather, expected_total_children);
+    if (preAdaptMaxLevel) {
+        BOOST_CHECK( father.level() <= preAdaptMaxLevel );
+    }
+    else {
+        BOOST_CHECK_EQUAL( father.level(), 0 ); // If there is no nested refinement
+    }
```
blattms, line 231: Do we need if-else? If preAdaptMaxLevel is false then it is 0 anyway and the first check should suceed.

tests/cpgrid/LgrChecks.hpp
```diff
@@ -100,66 +109,105 @@ void checkExpectedVertexGlobalIdsCount(const Dune::CpGrid& grid,
...
 
-    for (; it != endIt; ++it)
-    {
-        referenceElemOneParent_volume_it += it-> geometryInFather().volume();
-        for (int c = 0; c < 3; ++c)
-        {
-            referenceElem_entity_center_it[c] += (it-> geometryInFather().center())[c];
+void Opm::checkReferenceElemParentCellCenter(Dune::cpgrid::HierarchicIterator it,
+                                             const Dune::cpgrid::HierarchicIterator& endIt,
+                                             double total_children)
```
blattms, line 161: This could be determined by counting the iterations.

tests/cpgrid/LgrChecks.hpp
```diff
@@ -100,66 +109,105 @@ void checkExpectedVertexGlobalIdsCount(const Dune::CpGrid& grid,
...
+{
+    assert(total_children);
+    std::array<double,3> reference_elem_parent_cell_center = {0.0, 0.0, 0.0}; // Expected {.5,.5,.5}
+    for (; it != endIt; ++it) {
+        for (int c = 0; c < 3; ++c) {
```
blattms, line 166: Maybe should add an assert that all elements are on the same level as this is assumed by the algorithm?

aritorto:
Thanks for your first feedback, @blattms. All comments have been addressed. From the two new cases added in global_refine_test.cpp, we can decide which one stays, to reduce the time it takes (~5 seconds)

aritorto:

tests/cpgrid/LgrChecks.hpp
```diff
@@ -168,61 +216,94 @@ void Opm::checkEqMinMaxGlobalCellLevelZeroAndLeaf(const std::vector<int>& global
...
-
-                const auto expected_total_children = cells_per_dim_vec[element.level()-1][0]*cells_per_dim_vec[element.level()-1][1]*cells_per_dim_vec[element.level()-1][2];
-                BOOST_CHECK_CLOSE(element.geometryInFather().volume(), 1./expected_total_children, 1e-6);
-
-                checkFatherAndSiblings(element.getLevelElem(), element.level(), expected_total_children, grid);
+            if (element.hasFather() && (element.level()>preAdaptMaxLevel)) {
+                // second bool discards leaf element born in refined level grids that had not been marked for refinement in the last adapt call.
+                int subdivisionsIdx = element.level()-1-preAdaptMaxLevel;
+                const auto expected_total_children = cells_per_dim_vec[subdivisionsIdx][0]*cells_per_dim_vec[subdivisionsIdx][1]*cells_per_dim_vec[subdivisionsIdx][2];
+                BOOST_CHECK_CLOSE(element.geometryInFather().volume(), 1./expected_total_children, 1e-6);
```
blattms, line 280: Not part of this PR, but we seem to check a rather special case here (cell is a cube).

tests/cpgrid/LgrChecks.hpp
```diff
@@ -492,6 +578,165 @@ void Opm::checkConsecutiveChildGlobalIdsPerParent(const Dune::CpGrid& grid)
...
+                BOOST_CHECK(intersection_match.neighbor() == intersection.neighbor());
+
+                if(intersection.neighbor()) {
+                    BOOST_CHECK(intersection_match.indexInOutside() == intersection.indexInOutside());
+                }
+
+                BOOST_CHECK( Opm::areClose(intersection_match.centerUnitOuterNormal(), intersection.centerUnitOuterNormal()) );
+
+                const auto& geom_match = intersection_match.geometry();
+                BOOST_TEST(0.0 == 1e-11, boost::test_tools::tolerance(1e-8));
```
blattms, line 674: Is this debugging stuff?

tests/cpgrid/adapt_cpgrid_test.cpp
```diff
@@ -33,24 +33,14 @@
 
 #define BOOST_TEST_MODULE AdaptTests
 #include <boost/test/unit_test.hpp>
-#include <boost/version.hpp>
-#if BOOST_VERSION / 100000 == 1 && BOOST_VERSION / 100 % 1000 < 71
-#include <boost/test/floating_point_comparison.hpp>
-#else
-#include <boost/test/tools/floating_point_comparison.hpp>
-#endif
```
blattms, line 41: I could need see where this was moved. Is it not needed  anymore. What are we using instead of
```
BOOST_TEST(c1[c] == c2[c], boost::test_tools::tolerance(1e-12));
```
then?

blattms (commented):
Indeed the test files are much easier to read now.

Still missing the review for one file.

opm/grid/cpgrid/CpGrid.cpp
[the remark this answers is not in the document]
aritorto: I didn't express myself correctly. I meant that we call leafGridView() once per iteration in the old code (maybe I'm also wrong here). In any case, I'll go back to the old code or at least extract the leafGridView call outside the for loop (as in the provided link)

opm/grid/cpgrid/CpGrid.cpp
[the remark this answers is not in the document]
blattms: I see, thanks for the explanation

tests/cpgrid/LgrChecks.hpp
aritorto: Good point! Thanks!

tests/cpgrid/LgrChecks.hpp
aritorto: Indeed. I kept the param renaming it to expected_total_children, then comparing it with the total_dhildren (counting with the iterations)

tests/cpgrid/LgrChecks.hpp
aritorto: It could be a wrong interpretation of what geometryInFather should return. Currently, returns the geometry in the reference element, which is indeed the unit cube.

tests/cpgrid/LgrChecks.hpp
aritorto: It seems so. Removed now!

aritorto:

blattms (approved):
Looks good now. If you want, then please squash the commits and tell me when it is ready.

aritorto:

PR #861, opened 2025-03-28
Fix compilation with 64bit metis bindings of scotch.

blattms:
IDXTYPEWIDTH might be an expression and then compilation fails with
```
opm/grid/common/MetisPartition.cpp:151:30: error: missing binary operator before token "("
```

For newer scotch versions IDXTYPEWIDTH is `sizeof(::idx_t) * 8`


blattms:
jenkins build this please

lisajulia:
jenkins build this please

PR #862, opened 2025-04-04
Distribute level zero (with wells) of CpGrid with LGRs

aritorto:
In OPM/opm-grid#845, only two of the many loabBalance() in CpGrid were modified in such a way that it's possible to distribute level zero grid from a CpGrid with LGRs. Neither of those loadBalnce() had wells as a parameter. 

This PR 
1. modifies the loadBalance() methods that get called in opm-simulators, in such a way that distributing the level zero grid of a CpGrid with LGRs is supported.
2. tests (1.), invokes loadBalance with wells (either created or extracted from the schedule) as one of its arguments. Note: wells are added in the level zero grid of a CpGrid with LGRs. 
3. additionally, refactors GraphOfGrid::addWell(). 

Not relevant for the manual


aritorto:

opm/grid/GraphOfGrid.cpp
```diff
@@ -269,49 +269,97 @@ int GraphOfGrid<Grid>::wellID (int gID) const
...
+        return;
+    int wID = *(well.begin());
+
+    bool gridHasLgrs = (grid.maxLevel()>0);
+
+    std::set<int> origin_well{};
+    if (gridHasLgrs) {
+        for (const auto& w :well) {
+            const auto& leaf_elem = Dune::cpgrid::Entity<0>(*(grid.currentData().back()), w, true);
+            origin_well.insert( leaf_elem.getOrigin().index() );
```
michal-toth, line 339: I assume that the function `getOrigin().index()` returns the index of the cell from before the local grid refinement. This begs the question: How does the well know the ID of the locally refined cell?
AFAIK well cells are given through their cartesian coordinates, which are related to level 0. I do not know if there is another function processing well cell IDs to accommodate LGR but I would expect wells to be completely unaware of the local refinement.

tests/cpgrid/distribute_level_zero_from_grid_with_lgrs_and_wells_test.cpp
michal-toth, line 1: I do not know how could wells have IDs of locally refined cells. There might be functions outside this branch that has such functionality. If not, these tests are too detailed. (If user can not define wells with IDs of locally refined cells, what is tested here? Implementation details!)

michal-toth (commented):
With some small changes it looks fine. However, I have one big concern: How would it be possible to obtain a well with IDs of locally refined cells? I am familiar with one way of defining a well and that uses the cartesian coordinates of grid cells. Wells can then have only IDs of the parent cells and can not contain IDs of cells created by local refinement.
The troubles with handling a well with locally refined cells seems to be the major part of this PR but I do not see a way for this to happen. I have thus quite mixed feelings about this PR.

aritorto:
@michal-toth Thanks for your feedback. Indeed, this PR was still in draft mode so I understand your mixed feelings about it. 

After a discussion with @‌blattms, we realized that the addWell() function from master should also support the test cases I added involving LGRs, since the wells are constructed purely from ijk values of the level-zero grid. So I first fixed that.

The issue was resolved with a small change in the WellConnections constructor. (To make it more general, I should introduce an int level parameter, probably with a default value.)

aritorto:

aritorto:

opm/grid/GraphOfGrid.cpp
```diff
@@ -269,49 +269,97 @@ int GraphOfGrid<Grid>::wellID (int gID) const
...
+     wells.emplace_front(well);
+}
+
+
+template<typename Grid>
+void GraphOfGrid<Grid>::addWell (const std::set<int>& well, bool checkIntersection)
+{
+    if (well.size()<2)
+        return;
+    int wID = *(well.begin());
```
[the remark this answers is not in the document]
aritorto: I missed that in the refactor step, thanks!

opm/grid/GraphOfGrid.cpp
```diff
@@ -269,49 +269,97 @@ int GraphOfGrid<Grid>::wellID (int gID) const
...
+        for (const auto& w :well) {
+            const auto& leaf_elem = Dune::cpgrid::Entity<0>(*(grid.currentData().back()), w, true);
+            origin_well.insert( leaf_elem.getOrigin().index() );
+            // leaf_elem.getOrigin() returns father cell in level zero, if leaf_elem.level()>0,
+            // geometrically equivalent cell in level zero otherwise.
+        }
+    }
+    int origin_well_idx = *(origin_well.begin());
+
+    if (!gridHasLgrs) {
```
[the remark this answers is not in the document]
aritorto: Thanks, in the end, this approach was deleted, so nothing to do here.

opm/grid/GraphOfGrid.cpp
aritorto: Entity<0>::getOrigin() for an element on the leaf grid view, of a grid with LGRs, returns the parent cell, if the leaf element "was born" in an LGR (so it's a refined cell), or the geometrically equivalent cell from level zero otherwise. 

If the LGR has been created via CARFIN keyword, we do have an underlying Cartesian structure (therefore, local ijk's to potentially defined wells locally). However, the goal is of PR was not to define "local wells" in each refined level grid, but define them in level zero grid - as usual - instead. 

If the refinement has been triggered via adapt(), the local underlying Cartesian structure for each refined level grid is not guarantee - for now. 

To summarize, this PR intents to prove that we can add wells in level zero grid and distribute it "as usual", even though the grid has been refined.

tests/cpgrid/distribute_level_zero_from_grid_with_lgrs_and_wells_test.cpp
aritorto: I see your point. I deleted the extra "documentation" lines about local refinement indices. Also deleted some cases since there is no change in the behavior.

aritorto:
> "Is it possible to obtain a well with IDs of locally refined cells?". My concern is that you are now solving a problem that can not be created yet.

Sorry if I missed answering that. I'm not familiar with well code yet, but I would say obtaining a well with IDs of locally refined cells is not supported now. For this, I would assume that the well code need to be modified to be aware of refined level grids. As I said, not an expert in wells - yet, so that's what I can say for now.

In case, somehow a well is provided with IDs of locally refined cells, I would say that a first step is to replace those IDs for parent cells IDs and add the wells in level zero grid instead (which is supported!).

aritorto:

aritorto:
@blattms, @michal-toth ready for further review.

michal-toth (commented):
I have only two nitpicks.
I did not think that WellConnections could be adjusted to level 0 so simply/nicely.

CMakeLists_files.cmake
```diff
@@ -84,6 +84,7 @@ list (APPEND TEST_SOURCE_FILES
   tests/test_column_extract.cpp
   tests/cpgrid/addLgrsOnDistributedGrid_test.cpp
   tests/cpgrid/distribute_level_zero_from_grid_with_lgrs_test.cpp
+  tests/cpgrid/distribute_level_zero_from_grid_with_lgrs_and_wells_test.cpp
```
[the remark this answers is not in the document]
aritorto: Thanks, I missed that analogous correction from previous review (CMakeLists.txt)

opm/grid/CpGrid.hpp
```diff
@@ -1230,44 +1230,67 @@ namespace Dune
         // loadbalance is not part of the grid interface therefore we skip it.
 
         /// \brief Distributes this grid over the available nodes in a distributed machine
+        ///
         /// \param overlapLayers The number of layers of cells of the overlap region (default: 1).
         /// \param partitionMethod The method used to partition the grid, one of Dune::PartitionMethod
+        /// \param imbalanceTol
+        /// \param level Level grid to be distributed. Integer between 0,..., maxLevel().
+        ///              Defualt value set to -1, representing the leaf grid view.
+        /// \warning Throw if level>0. Currently, only distributing level zero grid is supported.
```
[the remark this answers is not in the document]
aritorto: Indeed, for CpGrid without LGRs. I corrected the documentation here with "Currently, for CpGrid with LGRs, only ..."

aritorto:
> I have only two nitpicks. I liked the WellConnections adjustment to allow level 0, I did not think it could be done so simply/nicely.

Thanks! : ) The two comments were addressed. I run jenkins one last time

aritorto:

opm/grid/common/WellConnections.cpp
```diff
@@ -82,13 +82,23 @@ WellConnections::WellConnections(const std::vector<OpmWellType>& wells,
...
                                  const Dune::CpGrid& cpGrid)
 {
-    const auto& cpgdim = cpGrid.logicalCartesianSize();
+    // - If the grid with dimensions {NX, NY, NZ} has been globally refined 'count' times,
+    //   then cpgrid.logicalCartesianSize() returns {2^count * NX, 2^count * NY, 2^count * NZ}.
+    //   By default, global refinement subdivides each cell into 2 in each direction (x, y, and z).
+    //
+    // - If the grid has undergone local (not global) refinement, cpgrid.logicalCartesianSize()
+    //   returns the logical Cartesian size of the original (level-zero) grid.
+    const auto& cpgdim = cpGrid.maxLevel()? cpGrid.currentData().front()->logicalCartesianSize() : cpGrid.logicalCartesianSize();
```
blattms, line 91: I think both branches should do the same, right? Maybe skip the decision and just always use the first?

opm/grid/common/WellConnections.cpp
```diff
@@ -82,13 +82,23 @@ WellConnections::WellConnections(const std::vector<OpmWellType>& wells,
...
+
     // create compressed lookup from cartesian.
     std::vector<int> cartesian_to_compressed(cpgdim[0]*cpgdim[1]*cpgdim[2], -1);
 
-    for( int i=0; i < cpGrid.numCells(); ++i )
+    // If the grid has LGRs, use globalCell from level zero grid.
+    const auto& globalCell = cpGrid.maxLevel()? cpGrid.currentData().front()->globalCell() : cpGrid.globalCell();
```
blattms, line 97: Same here: we seem to do the same in both branches. If there is only one level  `cpGrid.globalCell()` and `cpGrid.currentData().front()->globalCell()` yield the same.

tests/cpgrid/distribute_level_zero_from_grid_with_lgrs_and_wells_test.cpp
```diff
@@ -0,0 +1,291 @@
...
+
+        const auto& data = grid.currentData();
+        Opm::checkCellGlobalIdUniquenessForInteriorCells(grid, data);
+        Opm::checkConsecutiveChildGlobalIdsPerParent(grid);
+
+    }
+}
+
+BOOST_AUTO_TEST_CASE(extract_wells_from_deck_and_loadBalance_level_zero_of_cpgrid_with_lgrs_created_from_deck){
+    const std::string deck_string = R"(
```
blattms, line 144: Nice!

blattms (commented):
Looks good.

Just two nitpicks about code simplification (If there is only one level then that is what is returned by front())

opm/grid/common/WellConnections.cpp
aritorto: There is an inconsistency in how logical Cartesian size behaves in the following cases:
1. global refinement (well-defined for the leaf grid view)
2. local refinement via CARFIN (startIJK and endIJK) -> each refined level grid gets a well-defined logical Cartesian size. However, the leaf grid view gets a copy of level zero grid's logical Cartesian size.
3. local refinement via adapt() -> refined level grids and leaf grid get a copy of level zero grid's logical Cartesian size. 

I added a global refine case in the test, and deleted the comment regarding logical Cartesian size different behavior, since I was thinking to add a separated test (in a separated PR) describing this inconsistency. 

In any case, we can just take level zero logical Cartesian size here, omitting the decision.

opm/grid/common/WellConnections.cpp
aritorto: Thanks, both comments have been addressed.

aritorto:

PR #867, opened 2025-04-28
Check iterator is not at end, before dereferencing

aritorto:


aritorto:

opm/grid/GraphOfGrid.cpp
```diff
@@ -271,6 +271,9 @@ int GraphOfGrid<Grid>::wellID (int gID) const
 template<typename Grid>
 void GraphOfGrid<Grid>::mergeWellIndices(const std::set<int>& well)
 {
+    if (!well.empty())
```
akva2, line 274: i doubt you want to return if we are not empty.

aritorto: It should be when it is empty, I'll correct that

aritorto:
jenkins build this please

akva2:
last fix does the job, at least in terms of no debug iterator failures and test passing.

aritorto:

aritorto:
@akva2 Thanks for the gdb.txt. Please let me know if there is something else to improve/correct.

opm/grid/GraphOfGrid.cpp
```diff
@@ -279,17 +282,20 @@ void GraphOfGrid<Grid>::mergeWellIndices(const std::set<int>& well)
...
+        for (auto w = wells.begin(); w != wells.end();) {
             if (w->find(idx) != w->end()) {
                 // idx is in another well => remap it and join wells
                 if ( wellIdx == idx ) {
                     wellIdx = *(w->begin());
                 }
                 idx = *(w->begin());
                 newWell.insert(w->begin(), w->end());
-                wells.erase(w);
+                w = wells.erase(w);
```
akva2, line 293: unnecessary assignment as you're immediately breaking

akva2 (approved):
With the minor pointed out above.

opm/grid/GraphOfGrid.cpp
aritorto: done!

aritorto:

PR #869, opened 2025-05-02
Now formatting struct with braces on newline

kjetilly:
There seems to be a consensus that `struct` definitions should have their opening brace on a newline. The current `.clang-format` file places the brace on the same line as the struct (derived from `BreakBeforeBraces: Linux,`). This PR makes the brace come on a newline.

That is, the old version format as

```C++
struct Somestruct {
```

whereas this format as 

```C++
struct Somestruct
{
```

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -20,6 +20,7 @@
 	BraceWrapping: {
 		       AfterStruct: true,
 		       AfterClass: true,
-		       AfterFunction: true
+		       AfterFunction: true,
+		       AfterNamespace: true,
 	}
 }
--- a/.clang-format
+++ b/.clang-format
@@ -9,7 +9,7 @@
         AlwaysBreakTemplateDeclarations: Yes,
         BinPackArguments: false,
         BinPackParameters: false,
-        BreakBeforeBraces: Linux,
+        BreakBeforeBraces: Custom,
         BreakConstructorInitializers: BeforeComma,
         ColumnLimit: 120,
         Cpp11BracedListStyle: true,
@@ -17,4 +17,9 @@
         InsertNewlineAtEOF: true,
         MaxEmptyLinesToKeep: 5,
         NamespaceIndentation: Inner,
+	BraceWrapping: {
+		       AfterStruct: true,
+		       AfterClass: true,
+		       AfterFunction: true
+	}
 }


blattms:
For me it is always hard to know what the consensus is (I asked once what the clang-format file says in human words, but nobody knew).

So with this the formatting of class and struct is the same and it was not before?

kjetilly:
> For me it is always hard to know what the consensus is (I asked once what the clang-format file says in human words, but nobody knew).
> 
> So with this the formatting of class and struct is the same and it was not before?

So I have no idea about consensus, I just know that at least @atgeirr and @akva2 have requested newlines before the opening curly brackets for `struct`s.

But yes, with this PR, we would have the same behavior for `struct` and `class`, that is both with newline, whereas the master branch version currently only adds newlines for `class`.

I think the fact that it was not the samebefore, was due to us using the `Linux` style for `BreakBeforeBraces`, which [actually mandates no newline for `struct`s](https://www.kernel.org/doc/html/v4.10/process/coding-style.html), and Linux being pure C, probably does not have rules for `class`.

I can't summarize the full effect of the `.clang-format`, in part because we inherit from the WebKit style, but to give a concrete example, this is how the PR's `.clang-format` would format a generic C++ generated by ChatGPT, where the only formatting difference from the current `.clang-format` is the `struct` newline (I am personally annoyed by the newlines after return values, but I am also of the opinion that having a consistent formatting is better than a non-consistent...):

```C++
// main.cpp
#include <iostream>
#include <string>

#define ENABLE_LOGGING
#define MAX_SIZE 100

#ifdef ENABLE_LOGGING
#define LOG(msg) std::cout << "[LOG] " << msg << std::endl;
#else
#define LOG(msg)
#endif

namespace Core
{
struct Point
{
    int x;
    int y;

    void print() const
    {
        std::cout << "Point(" << x << ", " << y << ")" << std::endl;
    }
};

class Rectangle
{
public:
    Rectangle(int w, int h)
        : width(w)
        , height(h)
    {
    }
    int area() const
    {
        return width * height;
    }

private:
    int width, height;
};
} // namespace Core

namespace Utils
{
void
printMessage(const std::string& msg)
{
    std::cout << "Message: " << msg << std::endl;
}

template <typename T>
void
printValue(const T& value)
{
    std::cout << "Value: " << value << std::endl;
}

struct Config
{
    std::string name;
    int version;
};

template <typename T>
struct Container
{
    T data;
    void show() const
    {
        std::cout << "Container holding: " << data << std::endl;
    }
};

template <typename T>
class MathHelper
{
public:
    static T square(T x)
    {
        return x * x;
    }
};
} // namespace Utils

int
multiply(int a, int b)
{
    return a * b;
}

template <typename T>
T
add(T a, T b)
{
    return a + b;
}

int
main()
{
    LOG("Application started");

    Core::Point p {3, 4};
    p.print();

    Core::Rectangle rect(5, 6);
    std::cout << "Area: " << rect.area() << std::endl;

    Utils::printMessage("Hello, namespaces!");

    Utils::printValue<int>(42);
    Utils::printValue<std::string>("Template function!");

    Utils::Config cfg {"App", 1};
    std::cout << "Config: " << cfg.name << ", Version: " << cfg.version << std::endl;

    Utils::Container<std::string> c {"Hello container"};
    c.show();

    std::cout << "Square of 7: " << Utils::MathHelper<int>::square(7) << std::endl;
    std::cout << "Multiply 3 and 4: " << multiply(3, 4) << std::endl;
    std::cout << "Add 2.5 and 3.1: " << add(2.5, 3.1) << std::endl;

    LOG("Application finished");

    return 0;
}
```

kjetilly:
@blattms take this with a grain of salt, but this is how ChatGPT O3 would summarize the PR's `.clang-format`, just skimming it, it does not seem that far off:

**Evaluating clang-format settings**

I need to break down how the given .clang-format style will affect C++ file formatting. I’ll focus on details like indentations, line breaks, and the organization of code blocks, and I’ll try to gather a clear picture of specific rules for things like spaces around operators and curly braces. Let me update the information logically and concisely for easier understanding. Then, I'll summarize how this will impact code readability and consistency.


Thought for 14 seconds


Here’s what those options mean in plain English, so you know how `clang‑format` will shape your C++ files:

### Overall foundation

* **BasedOnStyle: WebKit** – Everything not overridden here follows WebKit’s rules (e.g., 4‑space indents, spaces before `*` and `&`, etc.).

### Alignment & spacing

* **No vertical alignment:**
  `AlignConsecutiveAssignments` and `AlignConsecutiveDeclarations` are **false**, so consecutive `=` signs or variable names will **not** be lined up into tidy columns.
* **AlignAfterOpenBracket: Align** – When a call, template, or initializer list wraps, the continued lines line up under the first token after the `(`, `[`, or `<`:

  ```cpp
  myFunction(long_parameter_name,
             another_parameter);
  ```
* **ColumnLimit 120** – Lines may wrap after 120 columns; longer lines are split according to the other rules.
* **MaxEmptyLinesToKeep 5** – Extra blank lines beyond five in a row are collapsed.
* **InsertNewlineAtEOF true** – Ensures the file ends with a single newline.

### Brace placement

* **BreakBeforeBraces: Custom** with
  `BraceWrapping` set to **AfterStruct/Class/Function/Namespace = true**
  ⇒ The opening brace goes **on its own line** for those constructs:

  ```cpp
  class Foo
  {
      ...
  };
  ```

### Line‑breaking strategy

* **AlwaysBreakAfterReturnType: TopLevelDefinitions** – For free‑standing functions and top‑level methods, the return type lives on its own line when breaking is needed:

  ```cpp
  int
  veryLongFunctionName(...)
  {
      ...
  }
  ```
* **AlwaysBreakTemplateDeclarations: Yes** – Template parameter lists are always broken after the `>`:

  ```cpp
  template <typename T>
  class Bar
  {
      ...
  };
  ```
* **BinPackArguments/Parameters: false** – When a call or declaration overflows the column limit, each argument/parameter gets its **own line** instead of being “packed”:

  ```cpp
  void foo(int first,
           long second,
           std::string third);
  ```
* **AllowShortBlocksOnASingleLine: false** and
  **AllowShortFunctionsOnASingleLine: None** – Even tiny `if`/`else` blocks or one‑liner functions are expanded to multi‑line form; no `{…}` on a single line.

### Constructor initializer lists

* **BreakConstructorInitializers: BeforeComma** – If an initializer list wraps, the break comes **before** each comma:

  ```cpp
  MyClass()
      : firstMember{1}
      , secondMember{2}
      , thirdMember{3}
  {}
  ```

### Other style tidbits

* **Cpp11BracedListStyle: true** – Enforces modern brace‑init spacing rules (`{1, 2}` rather than `{ 1, 2 }`).
* **NamespaceIndentation: Inner** – Code inside every namespace level is indented one extra level.
* **FixNamespaceComments: true** – Adds trailing comments after closing braces of a namespace:

  ```cpp
  } // namespace mylib
  ```

---

**In practice**
Expect a WebKit‑flavored style with **opening braces on their own lines**, **120‑column width**, **no column alignment grids**, each long function call or parameter list **broken one‑item‑per‑line**, and tidy namespace‑closing comments.

PR #868, opened 2025-04-29
Add logicalCartesianSize() test to show differences when refinement

aritorto:
This PR adds a test of CpGrid::logicalCartesianSize() and CpGridData::logicalCartesianSize(), showing the differences when the CpGrid has undergone local or (hidden) global refinement. 

grid.logicalCartesianSize() "makes sense" (is well-defined and coherent with the dimensions of the updated grid) when the CpGrid has undergone global refinement either by calling 
- explicitly grid.globalRefine( positive integer ),
- implicitly grid.addLgrsUpdateLeafView( /* params */) in such a way that the block of cells to be refined is the entire level zero grid. 
- implicitly grid.adapt(), if all elements have been marked for refinement. 

Otherwise, grid.logicalCartesianSize() returns a copy of the logicalCartesianSize of level zero grid (the grid dimensions before refinement). 

For refined level grids, CpGridData::logicalCartesianSize(), 
- if the refinement has been done via CpGrid::addLgrsUpdateLeafView(...), each refined level grid has a well-defined logicalCartesianSize. 
- if the refinement has been done via CpGrid::adapt(), all refined level grids have a copy of level zero grid logicalCartesianSize. 

Not relevant for the reference manual.

PR #875, opened 2025-05-20
Add cases with distributed grids in logicalCartesianSize_and_refinement_test.cpp

aritorto:
- Follow up of OPM/opm-grid#868 (see a detailed description there).

- Add the previously-failing case for serial: mark all elements and call adapt() should provide the expected number of refined cells in each direction, i.e. `NX*2`, `NY*2`, `NZ*2`, if coarse level zero grid has dimensions `{NX, NY, NZ}` and recalling that the number of subdivision per cell when refining via adapt() is $2\times 2\times 2$.
(bugfix `comm().size() == 0` :arrow_right: `comm().size() ==1`).

- Extend the test for distributed grid. 


aritorto:

aritorto:

aritorto:

opm/grid/cpgrid/CpGrid.cpp
```diff
@@ -2022,7 +2022,7 @@ bool CpGrid::adapt()
         is_global_refine = true; // parallel
     }
 #endif
-    if ( (comm().size() == 0) && (local_marked_elem_count == current_view_data_-> size(0)) ) {
+    if ( (comm().size() == 1) && (local_marked_elem_count == current_view_data_-> size(0)) ) {
```
blattms, line 2025: This part still looks a bit strange.

I am wondering whether we need the `#if HAVE_MPI ... #endif` above. If would hope that the code inside the #if statement would work even without MPI. In that case we can skip the if-statement and the code here.

blattms (commented):
This looks good.
My only suggestion is actually unrelated to this PR. I am sorry for not noticing it earlier.

Please try to get rid of the `#if HAVE_MPI ...` and the if-statement.

opm/grid/cpgrid/CpGrid.cpp
aritorto: Good point, it seems to work without if-MPI and reducing the if-statement.

aritorto:

PR #876, opened 2025-05-22
CartisianMapper returns logCartSize from level zero

aritorto:


aritorto:

PR #877, opened 2025-05-28
add SparseTable getter of pointer to underlying data

multitalentloes:
quite useful if we want to quickly copy a sparse table from the cpu to the gpu.


atgeirr (approved):

atgeirr:
jenkins build this please

PR #882, opened 2025-06-13
Use Volume type for volume and  integrationElement

SoilRos:
The dune interface requires Volume since a version 2.9 to possibly account for units on the grid. In this case (no units), is the same as the ctype.


aritorto:

PR #888, opened 2025-06-24
[tmp-fix] Inherit ParallelIndex flag from parent cell to fix communication for refined cells

aritorto:


aritorto:

PR #889, opened 2025-06-26
[tmp-fix] Ignore ParallelIndex flag when building cellIndexSet for CpGrid with LGRs to fix communication

aritorto:
When loadBalance() is called, overlapLayerSize defaults to 1. This defines how many faces must be crossed from an overlap cell to reach an interior cell. Conversely, any interior cell within that range of an overlap cell is also treated as shared.

For that reason, when defining the cellIndexSet for refined levels and leaf grids in a CpGrid with LGRs, the ParallelIndex boolean flag that indicates whether a cell may appear in multiple processes, was set to:

FALSE (these elements do not appear in any other process)
For fully interior cells, i.e., 
- interior cells that lie on the physical boundary of the grid with no neighbors,
- interior cells that are surrounded only by other interior cells.
(Neighbors are defined via shared intersections.)

TRUE (these elements might appear in other processes)
For 
- overlap cells
- interior cells with at least one overlap neighbor.

Problem:
Previously, this classification did not distinguish between coarse and refined cells. As a result:
- a refined cell (in a refined or leaf grid) might be tagged as false (fully interior),
- while its parent cell is not fully interior and thus is shared across processes.
This inconsistency leads to incorrect inter-process communication, since a child cell tagged as fully interior might actually be required on another process due to its parent’s interior-with-overlap-neighbor (NOT fully interior) status.

(Temporary) Solution (what this PR does):
Set the build flags to true to ignore the flags of Parallel indices.
Pros: 
- Ensures correctness by fully including all possibly needed cells.
Cons: 
- Increases communication overhead, reducing performance.

Alternative approach in closed OPM/opm-grid#888: Inherit ParallelIndex flag from parent cell.
Set the flags taking into account the parent cell situation. Precisely, if the parent cell is interior but has overlap neighboring cell, then all its children will have flag "true" since they all might appear in other processes. In this case, even if a child cell is interior and it's surrounding by other interior sibling cells, it gets a flag "true".

### DIFF
--- a/tests/cpgrid/communicate_distributed_grid_with_lgrs_test.cpp
+++ b/tests/cpgrid/communicate_distributed_grid_with_lgrs_test.cpp
@@ -194,7 +194,7 @@ void checkCommunicationWorks(const Dune::CpGrid& grid)
     const auto& gridView = grid.leafGridView();
     // initialize the volume of the finite volumes to zero
     const std::size_t numDof = gridView.size(0);
-    dofTotalVolume.resize(numDof, 0.0);
+    dofTotalVolume.resize(numDof);
 
     for (const auto& element : Dune::elements(gridView, Dune::Partitions::interior)) {
         // ignore everything which is not in the interior if the


tests/cpgrid/communicate_distributed_grid_with_lgrs_test.cpp
```diff
@@ -0,0 +1,297 @@
...
+    // of fvbasadiscretization finishInit()
+
+    double gridTotalVolume = 0.0;
+    std::vector<double> dofTotalVolume{};
+
+    const auto& gridView = grid.leafGridView();
+    // initialize the volume of the finite volumes to zero
+    const std::size_t numDof = gridView.size(0);
+    dofTotalVolume.resize(numDof);
+    std::fill(dofTotalVolume.begin(), dofTotalVolume.end(), 0.0);
```
blattms, line 198: This could be condensed to
```suggestion
    dofTotalVolume.resize(numDof, 0.0);
```
Or even do it in the constructor
```
std::vector<double> dofTotalColume(numDof, 0.0);
```
Actually the values will be default initialized anyway. Hence we can even skip specifying the zero above

tests/cpgrid/communicate_distributed_grid_with_lgrs_test.cpp
```diff
@@ -0,0 +1,297 @@
...
+
+    for (const auto& element : Dune::elements(gridView)) {
+        // ignore everything which is not in the interior if the
+        // current process' piece of the grid
+        if (element.partitionType() != Dune::InteriorEntity) {
+            continue;
+        }
```
blattms, line 205: Could be simplified to 
```suggestion
    for (const auto& element : Dune::elements(gridView, Dune::Partitions::interior)) {
        // ignore everything which is not in the interior if the
        // current process' piece of the grid
```

blattms (commented):
Nice just some very minor suggestions to simplify the code.

PR #891, opened 2025-07-04
Return geometry by value (alternative 3)

SoilRos:
In this PR, the idea is that the geometry holds all the corners of the cell. This because in most of the cases, the geometry is constructed in place and the compiler theoretically could see through the things that are actually needed and only generate code for that. In other words, it would avoid making the copy if those values are not used.

Here is a benchmark for that hypothesis: https://quick-bench.com/q/IPXA-MuhVrdQ_9DSbgbcAyRW9ek

* Indeed, this seems to be the case in GCC and there is no difference between storing the values or the pointers to the values.

8ek-EyAlGhyT2wGlEONCBPsEi3g

* Sadly, the picture changes for Clang and storing the values is 5 times more expensive even if only the center of the geometry is being used.

IPXA-MuhVrdQ_9DSbgbcAyRW9ek

Note that this PR is based on #885 which fixes all the lifetime issues of the geometry objects.

### DIFF
--- a/opm/grid/cpgrid/Entity.hpp
+++ b/opm/grid/cpgrid/Entity.hpp
@@ -528,35 +528,24 @@ Dune::cpgrid::Geometry<3,3> Dune::cpgrid::Entity<codim>::geometryInFather() cons
         OPM_THROW(std::logic_error, "Entity has no father.");
     }
 
-    // Indices of corners in entity's geometry in father reference element.
-    static constexpr std::array<int,8> in_father_reference_elem_corner_indices = {0,1,2,3,4,5,6,7};
-    // 'static': The returned object Geometry<3,3> stores a pointer to in_father_reference_elem_corner_indices. Therefore,
-    // this variable is declared static to prolongate its lifetime beyond this function (static storage duration).
-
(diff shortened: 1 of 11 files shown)


blattms:
jenkins build this please

opm/grid/cpgrid/Entity.hpp
```diff
@@ -523,35 +523,24 @@ Dune::cpgrid::Geometry<3,3> Dune::cpgrid::Entity<codim>::geometryInFather() cons
...
             for (int c = 0; c < 3; ++c)
             {
-                center_in_father_reference_elem[c] += corners_in_father_reference_elem_temp[corn][c]/8.;
+                center_in_father_reference_elem[c] += in_father_reference_elem_corners[corn][c]/8.;
             }
         }
         // Compute the volume of the 'local-entity'.
         double volume_in_father_reference_elem = double(1)/(cells_per_dim[0]*cells_per_dim[1]*cells_per_dim[2]);
         // Construct (and return) the Geometry<3,3> of 'child-cell in the reference element of its father (unit cube)'.
+        OPM_THROW(std::logic_error, "Entity has no father.");
```
aritorto, line 541: This throw before returning seems to be accidental

SoilRos:
Closing in favor of #1055.

PR #892, opened 2025-07-04
Improve interface compatibility with DUNE grid

SoilRos:
* ~~Add entity capability for codimension 1~~
* ~~Export Codim only for valid codimensions~~
* Implement the `id<codim>(e)` and `index<codim>(e)` methods
* Make iterators conform to forward iterator concept

### DIFF
--- a/opm/grid/CpGrid.hpp
+++ b/opm/grid/CpGrid.hpp
@@ -334,7 +334,7 @@ namespace Dune
         /// Set whether we want to have unique boundary ids.
         /// \param uids if true, each boundary intersection will have a unique boundary id.
         void setUniqueBoundaryIds(bool uids);
-       
+
 
         // --- Dune interface below ---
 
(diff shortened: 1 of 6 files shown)


SoilRos:

opm/grid/cpgrid/Intersection.hpp
```diff
@@ -300,6 +301,12 @@ namespace Dune
                 return *this;
             }
 
+            IntersectionIterator operator++(int)
+            {
+                IntersectionIterator tmp(*this);
```
atgeirr, line 306: This implementation is wrong. The correct is to increment this object (call increment() here) after you make the copy, and then return the copy in `tmp`. As is, you leave this object unchanged, but return an incremented object.

SoilRos: Good catch, thank you!

blattms:
jenkins build this please

opm/grid/CpGrid.hpp
```diff
@@ -2040,6 +2040,13 @@ namespace Dune
             static const bool v = true;
         };
 
+        /// \todo Please doc me !
+        template <>
+        struct hasEntity<CpGrid, 1>
+        {
+            static const bool v = true;
+        };
```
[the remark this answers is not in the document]
SoilRos: In either case, I removed it. It does not affect the checks anyways.

SoilRos:

blattms (approved):

PR #895, opened 2025-07-21
Entity::getOrigin returns oldest ancestor

aritorto:
Regarding nested refinement, it's convenient to modify Entity<0>::getOrigin() to return the oldest ancestor of an element, e.i. element.getOrigin() belongs to level zero grid. This PR implements that change. 


aritorto:
Approved (thanks @michal-toth!) and all checks passed, I'll merge it

PR #899, opened 2025-07-25
Add missing reference element method on entities

SoilRos:
The DUNE grid interface requires that entities are able to give a reference element directly from the entity and without necessarily going through the geometry.

### DIFF
--- a/opm/grid/cpgrid/Entity.hpp
+++ b/opm/grid/cpgrid/Entity.hpp
@@ -39,6 +39,7 @@
 
 #include <dune/common/version.hh>
 #include <dune/geometry/type.hh>
+#include <dune/geometry/referenceelements.hh>
 #include <dune/grid/common/gridenums.hh>
 
 #include "PartitionTypeIndicator.hpp"
@@ -292,6 +293,14 @@ public:
 protected:
     const CpGridData* pgrid_;
 };
+
+// Reference element for a given entity
+template <int codim>
+auto referenceElement(const Entity<codim>&)
+{
+    return Dune::referenceElement<double,3-codim>(Dune::GeometryTypes::cube(3-codim));
+}
+
 
 } // namespace cpgrid
 } // namespace Dune
--- a/tests/cpgrid/entity_test.cpp
+++ b/tests/cpgrid/entity_test.cpp
@@ -70,12 +70,22 @@ BOOST_AUTO_TEST_CASE(entity)
     BOOST_CHECK(e2 != e3);
     BOOST_CHECK(e2 != e4);
     BOOST_CHECK(e3 != e4);
+
+
     BOOST_CHECK_EQUAL(e1.level(), 0);
     // BOOST_CHECK(e1.type().isSingular()); // Our new type
     BOOST_CHECK(e1.type().isCube());
+    BOOST_CHECK(e1.type().dim() == 3);
+    using Dune::referenceElement;
+    BOOST_CHECK_EQUAL(referenceElement(e4).type(), e4.type());
+    BOOST_CHECK_EQUAL(referenceElement(e4.geometry()).type(), e4.type());
     BOOST_CHECK_EQUAL(e1.partitionType(), InteriorEntity);
+
     cpgrid::Entity<3> e5(g, 0, true);
+    BOOST_CHECK_EQUAL(e5.level(), 0);
     BOOST_CHECK(e5.type().isCube());
+    BOOST_CHECK_EQUAL(referenceElement(e5).type(), e5.type());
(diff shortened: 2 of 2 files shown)


aritorto:
@SoilRos thanks for noticing the lack of this method in Entity interface :) It looks great, except for a warning that brought up the only question/comment I have

bska:
I'm a little on the fence about this.  The CpGrid does not really **have** reference elements in the traditional sense.  It is true that each cell nominally is a cube, but pinch-outs and faults mean that the number of (distinct) vertices and intersections cannot be predicted/classified ahead of time.

SoilRos:
> I'm not entirely sure what I'm proposing, but it's more of the latter.

Ok, then this needs to discussed with @blattms. I got the task of testing the grid against the DUNE grid checks (See https://github.com/OPM/opm-grid/pull/903). Without this, I do not know how we could achieve that.

> The corner-point grid geometry itself does not support a reference element and it never will.

What prevents the implementation of the 5 missing reference elements (out of the 9) that arise from collapsing vertices of a cube?

> Any method that depends on a reference element will never work on a corner-point geometry.

Out of interest, what makes this (general) statement true? Is it the lack of affine mappings, or the lack of elements that can be posed on the topology of the reference element?

SoilRos:
@bska Thanks for that reference, that was definitely a nice read!

To be fair, I am not sure I get your point. It seems to me that we are talking about different things here. I am talking about the implementation of reference elements (in the sense of the [DUNE paper](https://conan.iwr.uni-heidelberg.de/data/people/peter/pdf/BastianBlattDednerEngwerKloefkornOhlbergerSander_Computing_2008.pdf)), which is by definition a _convex_ polytope. I do not see how the barycenter of an entity, or the number of _runtime_ sub-entities comes to play here.

Indeed, my impression is that the definition of a geometric realization (in the sense of the DUNE paper, again) of the CPGrid is kind of loosely defined, on purpose and for very good reasons, but this has nothing to do with the reference elements of the deformed entities.

SoilRos:
> [...] the real geometry of corner-point grids does not support a reference element.

I guess this is precisely the part that I do not understand. Could you please extend why? The MRST book that you provided says the following:

> Since each face of a grid cell is specified by four (arbitrary) points, the cell interfaces in the
> grid will generally be bilinear and possibly strongly curved surfaces. Geometrically, this
> can lead to several complications. Cell faces on different sides of a fault may intersect each
> other so that cells overlap volumetrically, or they may leave void spaces in between them.
> There may be tiny overlap areas between cell faces on different sides a fault, and so on.
> All these factors contribute to make fault geometries hard to interpret in a geometrically
> consistent way: a subdivision into triangles is, for instance, not unique. Likewise, top and
> bottom surfaces may intersect for highly curved cells with high aspect ratios, cell centroids
> may be outside the cell volume, etc.

Which basically admits that the proposed interpolation for the faces (bilinear between corners of the entity, or triangular subdivision) does not provide a valid geometric realization, hence, not a proper grid. But this does not mean that there is not a suitable geometric realization of the grid. From what I can see for example, making a bi- or tri-linear interpolation based on the end points of the vertical pillars[^1] and not on the corners of the entity will yield a mapping that does not have such gaps and overlaps between geometries. Perhaps am I wrong here? Or, is there perhaps something beyond these gaps that I do not know?

[^1]: Assuming that the pillars are straight.

SoilRos:
@atgeirr thanks for the detailed explanation.

> We have done DG in the past with CpGrid for example, quite successfully.

Amazing! I was under the impression that one needed a proper geometric realization for this. Is there a paper, or a source where I can learn more about this?

> I do not think this leads to less stiff problems though: the MINPV feature for example is there specifically to make the problems less stiff, a true discretization of the complete space without MINPV would have worse condition numbers.

I was meaning this in the sense of having the same active cells for simulation (e.g., same MINPV criterion) but different geometry so that the discretization only differs by the cell geometries being used. But I get what you mean too, enabling/disabling MINPV is probably more important in the end.

> In the end, if you use CpGrid (without NNCs or volume/depth shenanigans) through the Dune interface, you should get something reasonable enough to do interesting numerics on, if the grid otherwise conforms to the requirements of the discretization method.

So, coming back to @bska's question. I think this settles that proper geometries and reference elements exist for CpGrids, they are just not entirely implemented yet. But his question is still valid: Is it worth having the `referenceElement` (and for the same effect `obj.type() -> Dune::GeometryType`)  interface on `Entity` and `Geometry` even if we know that they are technically wrong?

atgeirr:
> Amazing! I was under the impression that one needed a proper geometric realization for this. Is there a paper, or a source where I can learn more about this?

https://andreas.folk.ntnu.no/papers/diag-ecmor-xiv.pdf

PR #902, opened 2025-07-28
[WIP&Messy] try to limit usage of IdSet with EntityRep

blattms:
It should only be used internally if you know what you are doing.

Not finished and quite messy, but at least it compiles. Might serve as inspiration, though.

Main point is to rename the id method taklng an EntityRep to kind of prevent misusage. :grimacing: 

This might break the DUNE interface....


aritorto:

PR #903, opened 2025-07-31
Enable DUNE grid tests

SoilRos:
For this to work it is additionally needed that
* Entities return the correct number of faces (to count intersections in the tests)
* The refinement does not throw. Instead, it returns false when it cannot make a refinement.


SoilRos:
I convert this PR to be a draft because dune 2.11 requires the grid to fulfill the grid concept check. And this will not work until
* the geometry type is returned as a value #885,
* the instantiations of `template Codim<2>::Entity` are disabled (see [dune-grid!778](https://gitlab.dune-project.org/core/dune-grid/-/merge_requests/778)), and
* the grid correctly handles the entities that cannot be handled (this original PR has some things missing for the tests to pass).

PR #884, opened 2025-06-13
Return geometry by value (alternative 2)

SoilRos:
To avoid copying a shared pointer, which is rather costly, the geometry now stores a variant holding either the ownership or a non-owning view to the entity variable. In the case of entity variables owned by the grid, a non-owning storage is fine as the geometry should not outlive the grid. On grid refinement the grid points are only owned by the geometry, so a move into the geometry is preferred.

Note that this PR is based on #885 which fixes all the lifetime issues of the geometry objects.

PR #904, opened 2025-07-31
Revert entity test with geometry

SoilRos:
The entity has no knowledge of the grid, hence, it cannot create a geometry so it segfaults. This test may work when the compiler manages to figure out that the geometry object is not really needed and can create the geometry on the fly. This may explain why this was not trigger in the CI.

This produces failures in other PRs like #885 #884 and #891

### DIFF
--- a/tests/cpgrid/entity_test.cpp
+++ b/tests/cpgrid/entity_test.cpp
@@ -78,14 +78,12 @@ BOOST_AUTO_TEST_CASE(entity)
     BOOST_CHECK(e1.type().dim() == 3);
     using Dune::referenceElement;
     BOOST_CHECK_EQUAL(referenceElement(e4).type(), e4.type());
-    BOOST_CHECK_EQUAL(referenceElement(e4.geometry()).type(), e4.type());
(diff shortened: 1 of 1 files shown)


tests/cpgrid/entity_test.cpp
```diff
@@ -78,14 +78,12 @@ BOOST_AUTO_TEST_CASE(entity)
...
     BOOST_CHECK_EQUAL(e1.partitionType(), InteriorEntity);
 
     cpgrid::Entity<3> e5(g, 0, true);
     BOOST_CHECK_EQUAL(e5.level(), 0);
     BOOST_CHECK(e5.type().isCube());
     BOOST_CHECK_EQUAL(referenceElement(e5).type(), e5.type());
-    BOOST_CHECK_EQUAL(referenceElement(e5.geometry()).type(), e5.type());
 
     // Cannot check other members without a real grid.
     // Put in more checks when it is possible to construct
```
SoilRos, line 89: This is my bad, I didn't read this comment until now. But it pretty much explains why I should not have used the geometry here.

aritorto:

aritorto (approved):

PR #907, opened 2025-08-04
Remove unnecessary friend declarations

aritorto:


michal-toth:
The removed friend function is defined and used in two test files. Since the tests run through, declaring the function seems indeed unnecessary.

aritorto:
@michal-toth Thanks for taking a look at it!

PR #911, opened 2025-08-06
Remove a few obsolete CpGridData methods

aritorto:
This PR removes a few unused CpGridData methods originally introduced to enforce a restriction during the early implementation of cell block refinement, where specified IJK ranges were required to be fully disjoint, i.e., they could not share vertices or faces. However, this assumption has been removed some time ago, and the method is no longer relevant or used.

Note: This PR removes  also the call to refinePatch from geometry_test.cpp. The reason for this change is that refinePatch is no longer the appropriate method for triggering the refinement of a block of cells ("patch"). The correct and current approach is to use CpGrid::addLgrsUpdateLeafView().


aritorto:
jenkins build this please

michal-toth (approved):
Good catch! Trimming the unused parts of code is always welcome.

PR #912, opened 2025-08-06
Refactor and rename grid_lgr_test.cpp

aritorto:
This PR refactors grid_lgr_test.cpp to improve clarity, reduce duplication, and align with updated assumptions, e.g. it is no longer required that all cells are active.

- Replace repeated logic with check methods from LgrChecks.hpp.
- Delete redundant test cases with overlapping behavior.
- Rename the test file from grid_lgr_test.cpp to addLgrs_in_allActiveCartesianGrid_test.cpp to better reflect its content and purpose: testing CpGrid::addLgrsUpdateLeafView(...) on a Cartesian grid with all active cells. 

Fix: Refinement was incorrectly triggered when the desired number of refined cells per dimension was {1, 1, 1}, a case that should not result in any refinement.
This PR resolves the issue by filtering out such cases before invoking the refinement.
The test has also been updated to include these cases.


aritorto:

aritorto:

blattms (approved):
Thanks a lot for the cleanup. Looks good.

Issue #505, opened 2021-01-04
BuildFaceToCell Calling EntityRep::setValue() With Index of "-1"


bska:
I was debugging another issue when an `assert` triggered in [`EntityRep<>::setValue()`](https://github.com/OPM/opm-grid/blob/723e0d3591bcd2e4c9b6b4b2c9c5f5f3a6993e90/opm/grid/cpgrid/EntityRep.hpp#L120).  The direct cause of this assertion failure was function [`cpgrid::buildFaceToCell()`](https://github.com/OPM/opm-grid/blob/723e0d3591bcd2e4c9b6b4b2c9c5f5f3a6993e90/opm/grid/cpgrid/processEclipseFormat.cpp#L785) calling [`setValue()`](https://github.com/OPM/opm-grid/blob/723e0d3591bcd2e4c9b6b4b2c9c5f5f3a6993e90/opm/grid/cpgrid/processEclipseFormat.cpp#L823) with
```
other_cell = -1
```

There are a number of possibilities here
1. The assertion is wrong and it's okay to call the function with a negative index (unlikely)
2. The calling code in `buildFaceToCell()` needs a check for negative cell indices
3. The pinch processing code is wrong since it generates a negative cell index

At the top of my head I don't know which is which here so I'd like to discuss the matter.  For the record, this issue presented when running the [VOLVE_2020](https://github.com/dabiged/Volve2OPM/blob/c9021eb4de3037536feebeec5a6d7902bf14134e/VOLVE_2020.DATA) model.

blattms:
1 and 2 should not be needed. 3 seems not to be the case as actnum is checked for both above and below, when [inserting](/OPM/opm-grid/blob/723e0d3591bcd2e4c9b6b4b2c9c5f5f3a6993e90/opm/grid/MinpvProcessor.hpp#L208). There are issues with some of the logic expressions, though.

Maybe the active cells are changed somewhere in between such that `global_to_local[it->second]<0` despite `actnum[it->second]>0`?

PR #913, opened 2025-08-11
Fix bug in faceCell().

atgeirr:
The faceCell() method is expected to return cell indices (or -1 for the outside) in the order implied by the face orientation. The addition of using std::numeric_limits<int>::max() to represent nonexistent cells in some contexts did not properly address this, but could give the wrong order in the output (i.e. orientation) if the first stored connection was the nonexistent one (i.e. r[0]) as that would not have a proper orientation associated with it.

Manual: Fixed a small bug for grid topology that did not affect OPM Flow, but did affect geomechanical codes.

### DIFF
--- a/opm/grid/CpGrid.hpp
+++ b/opm/grid/CpGrid.hpp
@@ -1703,7 +1703,14 @@ namespace Dune
         ///
         /// Note that a face here is always oriented. If there are two
         /// neighboring cells then the orientation will be from local_index 0
-        /// to local_index 1
+        /// to local_index 1.
+        /// If there is only one neighbouring cell, meaning that the face is on
+        /// the grid boundary, or a process boundary, then that cell's index will
+        /// be returned for local_index == 0 if the face normal is pointing outwards
+        /// from the cell and for local_index == 1 if the face normal is pointing in.
+        /// The other index will in that case yield -1. Effectively, we return
+        /// element local_index of (c, -1) or (-1, c) depending on face orientation
+        /// with respect to c.
         /// \param face The index identifying the face.
         /// \param local_index The local_index of the cell.
         /// \param level Integer representing the level grid to be considered.
--- a/opm/grid/cpgrid/CpGrid.cpp
+++ b/opm/grid/cpgrid/CpGrid.cpp
@@ -1175,35 +1175,40 @@ const cpgrid::OrientedEntityTable<0,1>::row_type CpGrid::cellFaceRow(int cell) c
 
 int CpGrid::faceCell(int face, int local_index, int level) const
 {
-    // In the parallel case we store non-existent cells for faces along
-    // the front region. Theses marked with index std::numeric_limits<int>::max(),
-    // orientation might be arbitrary, though.
-    bool validLevel = (level>-1) && (level<= maxLevel());
-    cpgrid::OrientedEntityTable<1,0>::row_type r
-        = validLevel? data_[level]->face_to_cell_[cpgrid::EntityRep<1>(face, true)]
+    // Get the correct neigbour relation (row from face->cell table).
+    const bool validLevel = (level > -1) && (level <= maxLevel());
+    const cpgrid::OrientedEntityTable<1,0>::row_type r = validLevel
+        ? data_[level]->face_to_cell_[cpgrid::EntityRep<1>(face, true)]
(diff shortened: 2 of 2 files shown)


bska:
Could this in any way be related to #505?

atgeirr:
> Could this in any way be related to #505?

I do not think so. I believe that issue is older than code change that introduced the buggy behaviour which is from LGR work.

atgeirr:
jenkins build this please

aritorto:
> > Could this in any way be related to #505?
> 
> I do not think so. I believe that issue is older than code change that introduced the buggy behaviour which is from LGR work.

Sorry to hear that. I would need some time to check where it's used. In a quick look, std::numeric_limits in face_to_cell_ seems to be used to distinguish boundary of the process, if I get [it](https://github.com/OPM/opm-grid/blob/9d9787663b58ba7e752c64bd5ba9ca347d15c298/opm/grid/cpgrid/CpGridData.hpp#L753C3-L759C53) correctly.

atgeirr:
> In a quick look, std::numeric_limits in face_to_cell_ seems to be used to distinguish boundary of the process, if I get [it](https://github.com/OPM/opm-grid/blob/9d9787663b58ba7e752c64bd5ba9ca347d15c298/opm/grid/cpgrid/CpGridData.hpp#L753C3-L759C53) (or [here](https://github.com/OPM/opm-grid/blame/9d9787663b58ba7e752c64bd5ba9ca347d15c298/opm/grid/cpgrid/CpGridData.cpp#L1501)) correctly.

I was mistaken, I assumed this treatment was added in LGR work since git indicated that the lines were written by you, but the parts you linked to are older and show that was not the case. Sorry for implying wrongly that the LGR had something to do with this!

In light of that, we cannot rule out that this is related to #505.

blattms:
Does anybody know what holds for the orientation of the cells attached to an NNC?

blattms:
You are right that negating is safe (i actually owe @michal-toth for noticing that instantly. Kudos to him. Thanks).

I checked the code where we set these indices up and we actually should use the correct orientation. So I guess the bug is really here. I admit that I do not know what exactly you were seeing

Please update the doxygen documentation to tell us exactly what is returned at index 0 
- if there are two cells,
- if there is one with positive orientation
- if there is just one with negative orientation (might only happen in parallel due to the implementation?)

Assuming that a face with one cell should always return that cell for index zero, shouldn't this do the trick and be clearer?

```c++
....
    cpgrid::OrientedEntityTable<1,0>::row_type r
        = validLevel? data_[level]->face_to_cell_[cpgrid::EntityRep<1>(face, true)]
        : current_view_data_->face_to_cell_[cpgrid::EntityRep<1>(face, true)];
    // In the legacy interface the first cell is the one with positive orientation or the only one attached to the face
    // but the way we store things the positive oriented one can be at any index
    int legacy_first_index = !r[0].orientation(); // zero if orientation of first cell is positive
    // In parallel std::numeric_limits<int>::max() marks a cell that is not there. Hence we can only return one cell
    // and that is the other one.
    auto size = r.size();
    if(r[0].index() == std::numeric_limits<int>::max()){
      // only one cell
     legacy_first_index = 1; //index of only valid cell
     --size; // only one cell;
      assert(r.size() > 1 && r[0].index() != std::numeric_limits<int>::max());
    }
    if(r[1].index() == std::numeric_limits<int>::max()){
      assert( r[0].index() != std::numeric_limits<int>::max());
      legacy_first_index = 0; //index of only valid cell
     --size; // only one cell;
    }
    int return_index = (legacy_first_index + local_index) % 2;

    if( return_index >= size)
       return -1;
    else
       return r[return_index].index();
}
´´´

atgeirr:
@blattms can I assume that on the front faces (with one max index cell in the face->cell relation), we still have the invariant that the two entities in the relation will have one of each orientation? That simplifies the code, and should be ensured in any case.

atgeirr:
For a sanity check, I replaced the function body with `return -999;`, and it does cause the test distribute_level_zero_from_grid_with_lgrs_and_wells_test_parallel to fail, and several other tests such as addLgrsOnDistributedGrid_test_parallel to hang. Doing `return 0` instead I get no hangs, but distribution_test_parallel fails. It therefore looks like we have some test coverage at least.

With my brand new implementation, all test succeed. I think it is a lot easier to read and understand, hope you think the same.

atgeirr:
jenkins build this please

PR #916, opened 2025-08-19
Fixes issues for mult pv

hnil:
-- This fix some issues with mult pv when collapsed cells get a eps volume.

### DIFF
--- a/opm/grid/MinpvProcessor.cpp
+++ b/opm/grid/MinpvProcessor.cpp
@@ -123,7 +123,7 @@ MinpvProcessor::process(const std::vector<double>& thickness,
                 bool c_active = actnum.empty() || actnum[c];
                 bool c_thin = (thickness[c] <= z_tolerance);
                 bool c_thin_inactive = !c_active && c_thin;
-                bool c_low_pv_active = pv[c] < minpvv[c] && c_active;
+                bool c_low_pv_active = (pv[c] < minpvv[c] && c_active) || (c_thin && c_active);
 
                 if (c_low_pv_active || c_thin_inactive) {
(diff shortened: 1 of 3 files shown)


blattms (changes requested):
Thanks a lot. There is certainly a bugfix in here.
- Please elaborate a bit more about the fix in the description (What was the issue with MULTPV?)
- please extend test/test_minpvprocessor.cpp with tests covering the behvior (and failing with out)
- we might need a few small regression tests to use to check this changes against other simulators. Please reach out if somebody else should do that. We do have good people for that
- please remove unrelated changes in files other than MinpvProcessor

hnil:
I will go over it one more time: main issue is that pv >minpv can be true for fully collapsed cells do to round of error but the thickness is ok and is also used in the processing so one get an incosistency.

The false trans is often a problem if one work on other physics, temperature + mechanics.

The edgconformal part is kind of a hack but it is to ensure path in the code which make a grid which is geometrical. i.e. not having "false" connections. The motivation is mechancis.

The fix is a hack to try to get the code working for some examples. But I do not se a nice way of doing this in the current code.

hnil:
Closing: superseded by work now staged in the geomech upstreaming plan.

The thin-cell low-PV semantics and the edge-conformal -> mergeMinPVCells wiring in this PR are being resent as a self-contained, default-off change stacked on top of #1051, justified by the no-NNC requirement of edge-conformal processing. That version is test-first and does not carry the commented-out c_low_pv_active condition this branch still has.

Closing rather than reworking here so the new one can open cleanly against #1051. Branch retained on my fork.

PR #917, opened 2025-08-19
Give correct labeling of nodes in parallel

hnil:
This corrects several issues with labeling of nodes. It include correct addition of corners for main the call used in opm-flow and for proper use depend on edge conformal processing.

### DIFF
--- a/opm/grid/CpGrid.hpp
+++ b/opm/grid/CpGrid.hpp
@@ -1744,6 +1744,9 @@ namespace Dune
         ///  of the vertex.
         int faceVertex(int face, int local_index) const;
 
+
+        /// \brief Get maps from vertexToCell out[0] and cellToVertex out[1];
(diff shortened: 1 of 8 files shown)


blattms:
jenkins build this please

hnil:
Closing: superseded by the opm-grid parallel series being prepared for upstreaming.

The node-labeling and corner-addition work here is being resent as an ordered pair: a partition-of-unity test that fails on master and demonstrates the unreachable hanging nodes, then the all-points interface, then vertex-adjacency corner overlap on top. That split makes the ordering invariant reviewable -- specifically that cell_to_allpoint_ rows are traversal-order deduplicated and NOT sorted, since the attribute exchange pairs sender and receiver rows positionally across ranks.

Closing so the series can open test-first. Branch retained on my fork.

PR #918, opened 2025-08-27
Janitoring: quell some SCA warnings

akva2:
Various static analyzer and clang(-tidy) warnings addressed

### DIFF
--- a/opm/grid/cpgpreprocess/make_edge_conformal.cpp
+++ b/opm/grid/cpgpreprocess/make_edge_conformal.cpp
@@ -121,7 +121,9 @@ sorted_outer_boundary(const processed_grid&   grid,
     // Remove internal edges.
     auto iter = edges.begin();
     auto iternext = iter;
-    ++iternext;
+    if (iternext != edges.end()) {
+        ++iternext;
+    }
 
     const opposite is_opposite{};
     for (; iternext != edges.end();) {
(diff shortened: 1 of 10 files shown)


opm/grid/cpgpreprocess/make_edge_conformal.cpp
```diff
@@ -98,10 +98,10 @@ sorted_outer_boundary(const processed_grid&   grid,
             };
         }
 
-        face_edges.emplace_back() = std::array {
+        face_edges.emplace_back(std::array {
```
bska, line 101: If you write this, you should use `push_back()` instead.  The only reason to use `emplace_back()` is that you can construct the object in place, but for the very specific case of `std::array` this does not work (in C++17) since there are no constructors.  In C++20 you can just pass the elements to `emplace_back()` however.

The current statement is in preparation of C++20.

bska (commented):
For what it's worth, `emplace_back() = std::array {...}` construction was actually intentional.

opm/grid/cpgpreprocess/make_edge_conformal.cpp
akva2: ic. dropped the commit.

akva2:

bska (approved):
Thanks a lot for the updates.  This looks good to me now, so please merge when green.

PR #919, opened 2025-08-29
Janitoring: Fix some issues seen with LTO

akva2:
First is probably not a real problem, but memset the struct to shut the compiler up.
Second is definitely a problem, as we ended up with a mix of CpGridData instances with and without MPI, leading to ODR violations and consequent warnings when LTO'ing.

### DIFF
--- a/opm/grid/GridUtilities.cpp
+++ b/opm/grid/GridUtilities.cpp
@@ -17,6 +17,8 @@
   along with OPM.  If not, see <http://www.gnu.org/licenses/>.
 */
 
+#include <config.h>
+
 #include <opm/grid/GridUtilities.hpp>
 #include <opm/grid/GridHelpers.hpp>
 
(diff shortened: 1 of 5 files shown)


opm/grid/cornerpoint_grid.c
```diff
@@ -170,6 +171,7 @@ create_grid_cornerpoint(const struct grdecl *in,
     struct UnstructuredGrid *g;
     int                      ok;
     struct processed_grid    pg;
+    memset(&pg, 0, sizeof(struct processed_grid));
```
bska, line 174: I think you can get the same effect, and portably, by declaring `pg` as
```C
struct processed_grid pg {0};
```
instead.  That syntax, while strange if you've not seen it before, works very similarly to the pair of empty braces in C++.  It sets the struct's integers to zero, its floating-point members to 0.0, and its pointers to `NULL`.

Is there a quick way to test that in a static analysis context?

bska (commented):
Much appreciated!  If at all possible, I'd like to use something other than `memset(0)` though.

akva2:
jenkins build this please

bska (approved):

PR #922, opened 2025-09-02
Nested refinement for CARFIN keyword

aritorto:
This PR updates addLgrsUpdateLeafView to support nested refinement.

Previously, each LGR could only refine cells directly from the global grid. With this change, an LGR can now specify another LGR as its parent, enabling hierarchical/nested refinements.

Each LGR entry in the input file may now include an additional string indicating the parent grid from which its cells should be refined.

This allows definitions such as:
CARFIN
--   I1  I2   J1  J2   K1  K2   NX  NY  NZ   GRID NAME

LGR1  1   4   1   4   1   3   8   8   6   GLOBAL
LGR2  1   4   1   4   1   6   8   8  12   LGR1
ENDFIN

Here, LGR2 refines cells from LGR1, rather than from the global grid.

Added a test case (serial only for now) to validate nested refinement behavior.

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -38,6 +38,7 @@ list (APPEND MAIN_SOURCE_FILES
   opm/grid/cpgrid/Iterators.cpp
   opm/grid/cpgrid/Indexsets.cpp
   opm/grid/cpgrid/LgrHelpers.cpp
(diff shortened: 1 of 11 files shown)


aritorto:

aritorto:
Re-based, approved, and checks passed. I'll merge it!

PR #924, opened 2025-09-08
fix doxy errors

akva2:
Ref https://ci.opm-project.org/job/opm-grid-static-analysis/326/doxygen/


aritorto (approved):

doc/doxygen/Doxylocal
```diff
@@ -1,4 +1,3 @@
 INPUT                  += @abs_top_srcdir@/doc/doxygen/mainpage \
                          @abs_top_srcdir@/doc/doxygen/modules \
-                         @abs_top_srcdir@/dune/common/ \
-                         @abs_top_srcdir@/dune/grid/
+                         @abs_top_srcdir@/opm/grid /
```
bska, line 3: What does this&ndash;i.e., a single trailing `/` character&ndash;do to the configuration?  Does it mean we effectively search the full tree rooted at `/` or does it do something else (or nothing at all)?

bska (commented):
Most of this looks fine, but I don't understand all parts of the change to `Doxylocal`. Would you mind elaborating a little bit on that, please?

PR #929, opened 2025-09-30
GPU support SparseTable<INFO> for assembly

multitalentloes:


opm/grid/utility/SparseTable.hpp
```diff
@@ -288,5 +432,46 @@ namespace Opm
...
+}
+
+// Since we cannot have MatrixBlock in GPU code, we need to convert to MiniMatrix
+// For now I think this is an okay place to convert types
+template<class MatrixBlockType, class MiniMatrixType, class ResidualNBInfoType, template <typename, typename...> class TemplatedStruct>
+auto copy_to_gpu(const SparseTable<TemplatedStruct<ResidualNBInfoType, MatrixBlockType>>& cpu_matrix)
+{
+    std::vector<TemplatedStruct<ResidualNBInfoType, MiniMatrixType>> minimatrices(cpu_matrix.dataSize());
+    for (auto e : cpu_matrix.dataStorage()) {
+        minimatrices.push_back(e); // will this be converted correctly???
```
atgeirr, line 454: You will end up with minimatrices having twice the size it should, either give it size on construction or use push_back, not both.

opm/grid/utility/SparseTable.hpp
```diff
@@ -288,5 +432,46 @@ namespace Opm
...
+template <class T>
+auto copy_to_gpu(const SparseTable<T>& cpu_table)
+{
+    return SparseTable<T, GpuBuffer>(
+        GpuBuffer<T>(cpu_table.dataStorage()),
+        GpuBuffer<int>(cpu_table.rowStarts())
+    );
+}
+
+// Since we cannot have MatrixBlock in GPU code, we need to convert to MiniMatrix
```
atgeirr, line 447: The whole existence of this specialization is a bit of a hack... I suggest that we place it with the definition of the struct (where it can be concrete and need not be templatized) instead, it does not have to be placed here.

atgeirr (changes requested):
Should merge, but needs some fixes

opm/grid/utility/SparseTable.hpp
multitalentloes: This is a completely criminal error on my side...

opm/grid/utility/SparseTable.hpp
multitalentloes: Not really sure how to avoid this. I suppose we want to end up with this being a very straight-forward initialization where we can just use `SparseTable(GPUBuf1, GPUBuf2)` and have the conversion of the stuff in the struct be handled automatically. I have made a couple of attempts at this but I have not gotten anything to work smoothly...

opm/grid/utility/SparseTable.hpp
```diff
@@ -73,6 +140,15 @@ namespace Opm
 	    setRowStartsFromSizes(rowsize_beg, rowsize_end);
         }
 
+        SparseTable (Storage<T> data, Storage<int> row_starts)
+            : data_(data)
+            , row_start_(row_starts)
+        {
+            // removed for now because we cannot access the zero'th element if Storage is a GpuBuffer
+            // OPM_ERROR_IF(row_start_.size() == 0 || row_start_[0] != 0,
+            //              "Invalid row_start array");
```
kjetilly: I think storage can probably be passed as an rvalue reference:

```C++
Storage<T>&& data, Storage<int>&& row_starts 
```

it will save us an allocation and a copy on the gpu.

kjetilly: also, I would either remove the commented out code, or remove the comment and have the exception in an `if constexpr (std::is_same_v<Storage<int>, std::vector<int>>)`

kjetilly (changes requested):
I think the copy and make view functions have the wrong signature, but I am not sure. These are not actually tested, so a bit difficult to tell.

Also some minor issues with using an rvalue ref instead of copying

opm/grid/utility/SparseTable.hpp
```diff
@@ -288,5 +405,49 @@ namespace Opm
...
+    return SparseTable<T, GpuBuffer>(
+        GpuBuffer<T>(cpu_table.dataStorage()),
+        GpuBuffer<int>(cpu_table.rowStarts())
+    );
+}
+
+// Since we cannot have MatrixBlock in GPU code, we need to convert to MiniMatrix
+// For now I think this is an okay place to convert types
+template<class MatrixBlockType, class MiniMatrixType, class ResidualNBInfoType, template <typename, typename...> class TemplatedStruct>
+auto copy_to_gpu(const SparseTable<TemplatedStruct<ResidualNBInfoType, MatrixBlockType>>& cpu_matrix)
```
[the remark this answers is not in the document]
multitalentloes: done

bska:
jenkins build this opm-simulators=6508 serial rocm hipify please

multitalentloes:
jenkins build this opm-simulators=6508 serial rocm hipify please

multitalentloes:
jenkins build this opm-simulators=6508 serial rocm hipify please

PR #935, opened 2025-10-14
Skip NNCs created by PINCH if one of the cells is collapsed.

### LINKED ISSUE #505: BuildFaceToCell Calling EntityRep::setValue() With Index of "-1"

blattms:
Collapsed here means that the points of the top and lower plane on the same pillar coincide.
Later those cells will be treated as inactive in [preprocess.c](/OPM/opm-grid/blob/6a456765a4f2d55182738df4dd6cf690b357a42f/opm/grid/cpgpreprocess/preprocess.c#L380-L388) and will not be part of the grid. when we create the grid topology. They would present numerical problems later anyway.

Note that this means that these cells (collapsed but PORV larger than MINPVV) will now be present a barrier preventing creation of a PINCH NNC.

Fixes the assertion:
```
flow: ./opm/grid/cpgrid/EntityRep.hpp:121: void Dune::cpgrid::EntityRep<codim>::setValue(int, bool) [with int codim = 0]: Assertion `index_arg >= 0' failed.
```
Closes #505

### DIFF
--- a/opm/grid/MinpvProcessor.cpp
+++ b/opm/grid/MinpvProcessor.cpp
@@ -57,6 +57,13 @@ double MinpvProcessor::computeGap(const std::array<double,8>& coord_above,
     return min_val;
 }
 
+/// \brief Whether top and bottom plane coincide
+bool isCollapsed(const std::array<double,8>& coord)
+{
+    return coord[0] == coord[4] && coord[1] == coord[5] &&
+        coord[2] == coord[6] && coord[3] == coord[7];
+}
+
 MinpvProcessor::Result
 MinpvProcessor::process(const std::vector<double>& thickness,
                         const double z_tolerance,
@@ -293,8 +300,11 @@ MinpvProcessor::process(const std::vector<double>& thickness,
                         option4ALLZero = option4ALLZero || (!permz.empty() && permz[c_above] == 0.0) || multz(c_above) == 0.0;
                         nnc_allowed = nnc_allowed && (computeGap(cz_above, cz_below) < max_gap) && (!pinchOption4ALL || !option4ALLZero) ;
 
+                        // Note that collapsed cells become inactive in preprocess.c
+                        // We treat them as a barrier preventing NNCs here.
                         if ( nnc_allowed &&
                              (actnum.empty() || (actnum[c_above] && actnum[c_below])) &&
+                             !isCollapsed(cz_below) && !isCollapsed(cz_above) &&
                              pv[c_above] > minpvv[c_above] && pv[c_below] > minpvv[c_below]) {
                             result.add_nnc(c_above, c_below);
                         }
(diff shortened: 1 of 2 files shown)


blattms:
With this the collapsed cell  will prevent creation of an NNC. It might turn out that this is not what we want, but it is certainly way better than a model aborting in an assert. Hence I think this should be merged but we should compare behavior with others.

blattms:
jenkins build this please

blattms:
jenkins build this please

blattms:
Bugfix from October. Maybe this can make it into the release?

There is a test added in case you were wondering.

blattms:
jenkins build this please

plgbrts:
Thanks @blattms for this fix. With it, a big field case (18M cells) was able to run that otherwise gives the mentioned assert.  It would be great if this PR can be merged.

blattms:
Would do you think @bska? Finally ready to merge after a rebase?

blattms:
jenkins build this please

atgeirr:
I believe the code accomplishes what it sets out to do. What I do not understand yet is how we end up in the situation where we have cells that are

> ... collapsed but PORV larger than MINPVV

as that indicates a completely degenerate cell (collapsed along all pillars) still have a positive pore volume. I would be very surprised if this can happen in our normal calculations of pore volume, so my hypothesis is that this happens because the deck overrides the pore volume of the cell in question directly (i.e. setting PORV), but I would need to see some of the cases that made this fix necessary to verify that.

I will merge this fix now as it adresses observed problems, but think it is likely we will revisit the minpv code.

PR #1018, opened 2026-04-09
[LGR] Improvement in nested refinement, in serial

aritorto:
This PR extends the autoRefine(...) test by showing that the method can be invoked 
- multiple times,
- both before and after calls to globalRefine().

To support this, a set of auxiliary utilities has been introduced:
- detect whether the grid has already undergone global refinement
-  retrieve the grid name corresponding to the maximum refinement level

These additions represent an improvement on nested refinement. At this stage, limited to serial setting (undistributed grid).


aritorto:

PR #1022, opened 2026-04-10
[LGR] Fix wrong parent intersection lookup at LGR coarse/fine interfaces

arturcastiel:
## Fix

Ensure `getParentIntersectionFromLgrBoundaryFace` always returns the correct level-0 (parent/coarse) intersection at LGR coarse/fine interfaces, independent of whether the refined cell is `intersection.inside()` or `intersection.outside()`.

The implementation now maps the leaf-grid intersection to the level-0 grid using `getOrigin()` and performs comparisons using `indexInInside()` within the same grid level. This avoids mismatches caused by using indices from different cells with opposite local face orientations.

This guarantees:
- Correct parent intersection selection for both inside/outside refined configurations.
- Consistent face index comparisons within the level-0 grid.

## Impact

Restores correct geometric references at LGR boundaries and ensures symmetric and physically consistent transmissibility calculations across coarse/fine interfaces.

### DIFF
--- a/opm/grid/cpgrid/CpGrid.cpp
+++ b/opm/grid/cpgrid/CpGrid.cpp
@@ -1237,33 +1237,26 @@ int CpGrid::faceVertex(int face, int local_index) const
 Dune::cpgrid::Intersection CpGrid::getParentIntersectionFromLgrBoundaryFace(const Dune::cpgrid::Intersection& intersection) const
 {
     if ( intersection.neighbor()) {
-        if ((intersection.inside().level() != intersection.outside().level())) {
-            // one coarse and one refined neighboring cell
-            /** Now, it could also be two refined cells. In that case, any of them will fit to search for the parent face */
-            const auto& cellIn = intersection.inside();
-            const auto& cellOut = intersection.outside();
-
-            // Identify the coarse and the refined neighboring cell
-            const auto coarseCell =  (cellIn.level() == 0) ? cellIn : cellOut;
-            const auto refinedCell =  (coarseCell == cellIn) ? cellOut : cellIn;
-            assert(coarseCell.level() != refinedCell.level());
-
-            // Get parent cell of the refined cell
-            const auto& parentCell = refinedCell.father();
-
-            // Get the index inside and orientation from the leaf grid (refined) face
-            const auto& intersectionIdxInInside = intersection.indexInInside();
-
-            for(const auto& parentIntersection : intersections(this->levelGridView(0), parentCell)){
-                // Get the inInsideIdx and orientation from the parent intersection
-                const auto& parentIdxInInside = parentIntersection.indexInInside();
-                if (parentIdxInInside == intersectionIdxInInside) {
-                    return parentIntersection;
+        // Only handle intersections between cells at different levels where one of them
+        // is at level 0 (coarse/fine boundary). Skip same-level intersections and
+        // intersections between two different refined level grids (neither at level 0).
+        if (intersection.inside().level() != intersection.outside().level() &&
+            intersection.inside().level() * intersection.outside().level() == 0) {
+            // Get the equivalent level-0 cell if intersection.inside() is already at level 0,
+            // or the coarsest ancestor at level 0 if it is a refined cell.
+            // In both cases, intersection.indexInInside() is the correct face index to match:
+            // - coarse inside (level 0): indexInInside() is already the level-0 face index.
+            // - refined inside: fine cells inherit face directions from their parent, so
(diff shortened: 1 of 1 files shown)


arturcastiel:
jenkins build this please

aritorto:
This looks like a great catch. I would suggest slightly revising the PR description, for example:

"[...] intersection.indexInInside() is the local face index in the intersection.inside() cell, which may correspond to either the refined cell or the coarse cell. In the latter case, the comparison with parentIntersection.indexInInside()—the local face index in the parent cell—yields an incorrect conclusion. [...]"

arturcastiel:
@aritorto thank you for the suggestions. I will look them carefully.

aritorto:
@arturcastiel, I added [a test](https://github.com/OPM/opm-grid/pull/1023) that should fail based on these findings, but it currently passes. I’ll look into it and follow up!

One thing I forgot to mention is that this method is intended to handle intersections belonging to both coarse and refined cells, since both appear as cells in the leaf grid view. In both cases, it should be possible to return the corresponding original coarse intersection, regardless of whether the given intersection belongs to a coarse or a refined cell, as shown in the test in https://github.com/OPM/opm-grid/pull/1023

arturcastiel:
@aritorto should we talk about this in a quick meeting?

aritorto (changes requested):
This can be simplified by using getOrigin(), which returns either:

the coarsest ancestor in the level-0 grid if intersection.inside() is refined, or
the geometrically equivalent cell if it is a leaf cell that was never refined.

This avoids explicitly checking whether the cell is coarse or refined, and ensures that the comparison between face tag and orientation remains consistent

aritorto (changes requested):
I make the suggestion even though it's still in draft mode

opm/grid/cpgrid/CpGrid.cpp
aritorto:
```suggestion   
            // Skip case when both cells belong to different refined level grids
            if (intersection.inside().level()*intersection.outside().level() == 0) {

                // Get the equivalent level-0 cell if intersection.inside().level() == 0;
                // otherwise, get the coarsest ancestor at level 0. 
                const auto insideOrigin = intersection.inside().getOrigin();
                
                for(const auto& originIntersection : intersections(this->levelGridView(0), insideOrigin)){
                    if (originIntersection.indexInInside() == intersection.indexInInside()) {
                        return originIntersection;
                    }
                }
            }
```
```

Comment (aritorto) [reply opm/grid/cpgrid/CpGrid.cpp]: I couldn’t include all the lines I’d remove in this suggestion, but I hope the intended change is still clear. Let me know if anything is unclear—I’m happy to clarify!

**Review by aritorto — CHANGES_REQUESTED** [review]

**Comment by arturcastiel** [comment]
jenkins build this please

**Review by aritorto — CHANGES_REQUESTED** [review]
Thanks for incorporating the suggestion. We still need to include the line that verifies that the cells attached to the given intersection come from different grid levels, in addition to checking that one of those levels is zero.

**Comment by arturcastiel** [comment]
true, hold on

File: opm/grid/cpgrid/CpGrid.cpp [hunk 1]
```diff
@@ -1237,28 +1237,19 @@ int CpGrid::faceVertex(int face, int local_index) const
 Dune::cpgrid::Intersection CpGrid::getParentIntersectionFromLgrBoundaryFace(const Dune::cpgrid::Intersection& intersection) const
 {
     if ( intersection.neighbor()) {
-        if ((intersection.inside().level() != intersection.outside().level())) {
```
aritorto, line 1240: We also need this line, or somehow check both:
- inside and outside elements belong to different level grids
- one of those levels must be zero (we do not handle the case when cells attached to the intersection belong to two different refined level grids)

arturcastiel:
jenkins build this please

aritorto:
@arturcastiel would you mind updating the “Problem” and “Fix” sections in the PR description as well?

Key points:

- The method returns the parent (coarse) intersection, regardless of whether intersection.inside() refers to a coarse or refined cell.
- The comparison using intersection.indexInInside() was incorrect in cases where the refined cell corresponds to intersection.outside().
- The new approach, using getOrigin(), ensures a correct comparison of indexInInside between leaf-grid and level-zero-grid intersections.


Also, is there a reason the PR is still in draft mode?

aritorto (approved):

aritorto:

aritorto:
@arturcastiel thanks for incorporating the suggestion and updating the PR description, merging!

PR #1027, opened 2026-04-27
Remove face parent container in return refineSingleCell

aritorto:
The refinement strategy is implemented by constructing auxiliary single-cell refinement CpGridData objects. Geometries are then grouped and stored per grid level, avoiding duplication caused by shared vertices or intersections.

This PR removes an intermediate auxiliary container that mapped parent cell face indices to the corresponding refined faces generated during single-cell refinement.

### DIFF
--- a/opm/grid/cpgrid/CpGridData.cpp
+++ b/opm/grid/cpgrid/CpGridData.cpp
@@ -1782,9 +1782,10 @@ bool CpGridData::hasNNCs(const std::vector<int>& cellIndices) const
 }
 
 std::tuple< const std::shared_ptr<CpGridData>,
-            const std::vector<std::array<int,2>>,                 // parent_to_refined_corners(~boundary_old_to_new_corners)
-            const std::vector<std::tuple<int,std::vector<int>>> > // parent_to_children_faces (~boundary_old_to_new_faces)
-CpGridData::refineSingleCell(const std::array<int,3>& cells_per_dim, const int& parent_idx) const
+            const std::vector<std::array<int,2>>>               // parent_to_refined_corners(~boundary_old_to_new_corners)
+CpGridData::refineSingleCell(const std::array<int,3>& cells_per_dim,
+                             const int& parent_idx,
+                             std::vector<std::vector<std::pair<int, std::vector<int>>>>& faceInMarkedElemAndRefinedFaces) const
 {
     // To store the LGR/refined-grid.
     std::vector<std::shared_ptr<CpGridData>> refined_data;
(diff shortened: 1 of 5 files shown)


aritorto:

opm/grid/cpgrid/CpGridData.cpp
```diff
@@ -1797,15 +1798,17 @@ CpGridData::refineSingleCell(const std::array<int,3>& cells_per_dim, const int&
...
+   
     const auto& parent_to_point = this->cell_to_point_[parent_idx];
-    Opm::Lgr::containsEightDifferentCorners(parent_to_point);
+    Opm::Lgr::containsEightDifferentCorners(parent_to_point); // refinement supported only for hexahedron
+    
     // Refine parent cell
-    parent_cell.refineCellifiedPatch(cells_per_dim, refined_geometries, refined_cell_to_point, refined_cell_to_face,
-                                     refined_face_to_point, refined_face_to_cell, refined_face_tags, refined_face_normals,
-                                     {1,1,1}, /*widthX, lengthY, heightZ*/ {1.}, {1.}, {1.});
+    const auto& parentCellElem = Entity<0>(*this, parent_idx, true);
```
blattms: Let's keep it simple. Prolonging the lifetime of temporary is not needed if we use an object:
```suggestion
    const auto parentCellElem = Entity<0>(*this, parent_idx, true);
```

tests/cpgrid/lgr/replace_lgr1_corner_idx_by_lgr2_corner_idx_test.cpp
```diff
@@ -66,11 +66,13 @@ const std::shared_ptr<Dune::cpgrid::CpGridData> createSingleCellGridAndRefine(co
     const std::array<int,3>& coarse_grid_dim = {1,1,1};
     lgr.createCartesian(coarse_grid_dim, cell_sizes);
 
+    std::vector<std::vector<std::pair<int, std::vector<int>>>> faceInMarkedElemAndRefinedFaces{};
+    faceInMarkedElemAndRefinedFaces.resize(6);
```
blattms, line 70: This is a bit suspicious as we will populate the container using push_back. Is it really needed?

tests/cpgrid/lgr/replace_lgr1_face_idx_by_lgr2_face_idx_test.cpp
```diff
@@ -65,11 +65,13 @@ const std::shared_ptr<Dune::cpgrid::CpGridData> createSingleCellGridAndRefine(co
     const std::array<int,3>& coarse_grid_dim = {1,1,1};
     lgr.createCartesian(coarse_grid_dim, cell_sizes);
 
+    std::vector<std::vector<std::pair<int, std::vector<int>>>> faceInMarkedElemAndRefinedFaces{};
+    faceInMarkedElemAndRefinedFaces.resize(6);
```
blattms, line 69: Same here.

blattms (commented):
A few nitpicks.

Technically the temporary was there before. Now it is juts more explicit.

tests/cpgrid/lgr/replace_lgr1_corner_idx_by_lgr2_corner_idx_test.cpp
aritorto: Nice observation. The resize is done on the outer std::vector. In both modified tests, we have to do it ourselves because we’re only calling refineSingleCell() on its own, without later building the level grids or updating the leaf grid view.

The push_back() calls then fill the inner vectors. We don’t know their size ahead of time—each face can be connected to one or two cells.

When refining the whole grid, this outer std::vector is resized automatically using currentLeafData()->face_to_cell_.size(), as shown [here](https://github.com/OPM/opm-grid/blob/d205522693c6e5d58db10fd71401b6acd2f6f143/opm/grid/cpgrid/CpGrid.cpp#L1967).

I hope this helps

opm/grid/cpgrid/CpGridData.cpp
aritorto: Thanks for the correction!

aritorto: I messed up the commit suggestion when rebasing and pushing, but at least the reference is not there anymore. Sorry!

aritorto:
jenkins build this please

PR #1032, opened 2026-05-04
[LGR] First steps towards faults

aritorto:


aritorto:
replaced by OPM/opm-grid#1045

PR #1034, opened 2026-05-11
Register Number of Processors for "cpgrid_aquifer_parallel" Test

bska:
The test uses four processors (`mpirun -np 4`) and `ctest` should know that for its scheduling algorithm.


akva2 (approved):

bska:
On the one hand, this isn't the most important aspect in the world. On the other, every other parallel test records its number of CPUs, so on consistency grounds alone it's better to have this than to not have it.

bska:
jenkins build this please

bska:
PR approved and build check is green.  I'll merge into master.

PR #1035, opened 2026-05-11
Make well edge weight equal to sum of grid edges instead of limits::max

michal-toth:
The partitioners metis and zoltan (not the default zoltanwell) uses std::numeric_limits::max for the weights of well edges (plus considers all cells perforated by one well to be pairwise neighboring). This PR lowers the constant to the sum of all transmissibilities of the grid. This induces an extra cost, we have to iterate over all transmissibilities first.

Testing showed that the algorithm with new weights yields partitioning that is as good as previous version, and in several cases the improvement is significant. The best observed result was 50% faster. (The computation on 12 ranks was the same. Splitting onto 16 ranks hurt the computation and increased the time by 16% with the updated weights whereas older version increased the time by 168%.) The updated algorithm is more robust and reaches the breaking point later.

We explain the improvement by giving better information to the partitioning algorithm. Lowering the gap between the weights between well connections and weights of other edges reduces the risk of losing information to rounding errors.

### DIFF
--- a/opm/grid/common/ZoltanGraphFunctions.cpp
+++ b/opm/grid/common/ZoltanGraphFunctions.cpp
@@ -245,14 +245,14 @@ void getCpGridEdgeList(void *cpGridPointer, int sizeGID, int sizeLID,
 #endif
 }
 template<typename ID, typename weightType>
-void fillNBORGIDAndWeightsForSpecificCellAndIncrementNeighborCounterForGridWithWells(const CombinedGridWellGraph& graph, const int localCellId, ID globalID, int& neighborCounter, ID& nborGID, weightType *ewgts) {
+void fillNBORGIDAndWeightsForSpecificCellAndIncrementNeighborCounterForGridWithWells(const CombinedGridWellGraph& graph, const int localCellId, ID globalID, int& neighborCounter, ID& nborGID, weightType *ewgts, const weightType& weWeight) {
     const Dune::CpGrid&  grid = graph.getGrid();
     // First the strong edges of the well completions.
     auto wellEdges = graph.getWellsGraph()[localCellId];
     for( auto edge : wellEdges)
     {
         nborGID[neighborCounter] = edge;
-        ewgts[neighborCounter++] = std::numeric_limits<weightType>::max();
+        ewgts[neighborCounter++] = weWeight;
     }
 
     // Now the ones of the grid that are not handled by the well completions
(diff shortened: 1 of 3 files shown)


michal-toth:
This is a draft until it gets tested with Metis partitioner. Plus we need to discuss whether the improvement is significant enough to outweigh the cost of iterating over the edges.

michal-toth:
jenkins build this please

michal-toth:
PR closed in favor of #1036. It is possible to set the well weights without the extra iteration over the grid and without interface changes.

Edit: #1036 turned up to be wrong. It is not possible to skip iterating over the grid. (That PR used information from edges of a single cell, not whole grid.)

michal-toth:
jenkins build this please

michal-toth:
Testing revealed that lowering the well edge weights indeed helps. Partitioning has fewer split wells (even though this PR lowers the weight of well edges). Most of the time the difference in computation time and number of iterations is miniscule but also found the case where it halves the computation time. This persuaded me that the one extra iteration over the grid is a small cost for improvement in robustness.

michal-toth:
jenkins build this please

opm/grid/common/ZoltanGraphFunctions.cpp
```diff
@@ -308,9 +308,19 @@ void getCpGridWellsEdgeList(void *graphPointer, int sizeGID, int sizeLID,
 #endif
     int neighborCounter = 0;
 
+    // well edge weight for partitioning, big enough that wells should not get split
+    float weWeight = 0; // type of ewgts even though transmissibility is double
+    for (int edge=0; edge<grid.numFaces(); ++edge) {
+        // assumes that transmissibility is positive (too obvious to leave an assert)
+        weWeight += graph.transmissibility(edge);
+    }
+    if (weWeight == std::numeric_limits<float>::infinity()) {
```
blattms: We would need to check `has_infinity()` and work around the case where this is false.

Maybe there is even an algorithm that works for idx_t and float and we could get rid off some code duplication?

blattms:
I think you have actually checked that this change is better than before. Maybe update the description where we say that we hope this is better.

michal-toth:
jenkins build this please

opm/grid/common/ZoltanGraphFunctions.cpp
michal-toth: The algorithm for Metis works in both cases. It is unclear if a specialized loop for float is faster, so I kept just the one for Metis, i.e. the version without infinity().

blattms (approved):

PR #1036, opened 2026-05-12
Partitioning: lower the well edge weights to the sum of all grid edges

michal-toth:
The partitioners metis and zoltan (not the default zoltanwell) uses std::numeric_limits::max for the weights of well edges (plus considers all cells perforated by one well to be pairwise neighboring). This PR lowers the constant to the sum of all transmissibilities of the grid.

We hope that this will make the partitioning better for hard cases when some wells get split. Lowering the gap between the weights between well connections and weights of other edges reduces the risk of losing information to rounding errors.


michal-toth:
jenkins build this please

PR #1038, opened 2026-05-19
remove Doxylocal

akva2:
adds nothing and contained autoconf substitution variables. the mainpage and modules files are template doxygen files from dune, they also add no value


akva2:
jenkins build this please

bska (approved):
Excellent! Please merge when green.

PR #1040, opened 2026-05-21
adjust to installed binary changes

akva2:
Downstream of https://github.com/OPM/opm-common/pull/5174


bska (approved):

PR #1043, opened 2026-06-19
SparseTable initializer_list constructor

atgeirr:
Makes it easier to make quick tables for tests etc.

Also: fixing some weird indentation errors in SparseTable.hpp (separate commit).

### DIFF
--- a/opm/grid/utility/SparseTable.hpp
+++ b/opm/grid/utility/SparseTable.hpp
@@ -44,6 +44,7 @@
 
 #include <algorithm>
 #include <cassert>
+#include <initializer_list>
 #include <numeric>
 #include <ostream>
 #include <type_traits>
@@ -147,6 +148,16 @@ private:
             }
         }
 
+
+        /// Initializer list constructor for easy construction of small SparseTables.
+        SparseTable(std::initializer_list<std::initializer_list<T>> initlist) requires (std::is_same_v<Storage<T>, std::vector<T>>)
+        {
+            row_start_.push_back(0);
+            for (const auto& row : initlist) {
+                data_.insert(data_.end(), row);
+                row_start_.push_back(data_.size());
+            }
+        }
 
         /// Sets the table to contain the given data, organized into
         /// rows as indicated by the given row sizes.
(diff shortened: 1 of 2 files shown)


atgeirr:
jenkins build this please

akva2 (approved):
LGTM

opm/grid/utility/SparseTable.hpp
```diff
@@ -148,8 +148,18 @@ struct PoisonIterator {
         }
 
 
+        /// Initializer list constructor for easy construction of small SparseTables.
+        SparseTable(std::initializer_list<std::initializer_list<T>> initlist)
```
bska: This constructor assumes that [`Storage`](https://github.com/OPM/opm-grid/blob/0ff52dd4aab43f51d65a0a2c552ada366d9d4d41/opm/grid/utility/SparseTable.hpp#L115) supports `insert(iterator, initializer_list<T>)`. Now `std::vector<T>` (the default) [obviously has such an `insert()` member function](https://en.cppreference.com/cpp/container/vector/insert), but that need not be the case for all `SparseTable<T>` clients. I suggest you restrict this constructor to those `Storage` types that actually support this particular `insert()` member function overload.

bska (commented):
I don't see the test code advertised in the PR description, but more importantly the new constructor makes rather severe assumptions about the underlying container.

opm/grid/utility/SparseTable.hpp
atgeirr: A very good point, I will try to ensure this.

And I forgot to commit the test addition, thanks for noticing!

atgeirr: Restricted the constructor to `Storage` being `std::vector`, which would be the common use case.

Also included the test additions.

atgeirr:
jenkins build this please

opm/grid/utility/SparseTable.hpp
```diff
@@ -148,8 +148,18 @@ struct PoisonIterator {
         }
 
 
+        /// Initializer list constructor for easy construction of small SparseTables.
+        SparseTable(std::initializer_list<std::initializer_list<T>> initlist) requires (std::is_same_v<Storage<T>, std::vector<T>>)
+        {
```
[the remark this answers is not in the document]
atgeirr: Fixed.

atgeirr:
jenkins build this please

bska (approved):

PR #1046, opened 2026-07-01
Check CpGrid::maxLevel() on distributed grid

aritorto:
Add an explicit check for distributed level-zero grids after refinement to ensure that grid.maxLevel() is identical across all ranks and matches the total number of LGRs represented in the global view.


aritorto:

PR #1050, opened 2026-07-06
Zoltan partitioning with transmissibility coarsened graph

andrthu:
Implementation of the Zoltan coarse graph partitioning method presented at the OPM summit.

Adds another option for doing partitioning, zoltanCG. The new method coarsens the partitioning graph by merging nodes that are connected with a transmissibilty larger than a given threshold. The threshold is calculated in opm-simulators. There is also a limit to how many nodes that can be merged set by another parameter in opm-simulators. Coarsening is done by performing a depth first search of the partitioning graph.

zoltanCG supports distributed-wells=true/false and serial-partitioning=true/false.

### DIFF
--- a/opm/grid/CoarseGraphOfGrid.cpp
+++ b/opm/grid/CoarseGraphOfGrid.cpp
@@ -28,60 +28,7 @@
 
 #include <numeric>
 
-
 namespace Opm {
-
-template<typename Grid>
-void CoarseGraphOfGrid<Grid>::dfsq(Row row, std::priority_queue<WgtIdx> &q, int v, int master,
-                                   double w, int maxNode, std::vector<bool>& visited,
-                                   std::vector<int>& cnode, std::vector<std::tuple<int,int,double> >& edges)
-{
-    visited[v] = true;
-    map_to_coarse_[v] = master;
-    cnode.push_back(v);
-
-    // Add all neighboring vertices of v with transmissibility larger than w to the queue. 
-    auto col = row.begin();
-    for (; col != row.end(); ++col) {
-        int nab = col.index();
-        double wgt = (*transGraph)[v][nab];
-        if ( wgt > w) {
-            if (!visited[nab]) {
-                q.push({wgt, nab});
-            }
-        }
-    }
-
-    // Only merge more vertices if the current coarse node is smaller than maxNode.
-    if ( (int)cnode.size() < maxNode ) {
-        if (!q.empty()) {
-
-            // Find strongest connection in queue q not already merged to cnode.
-            auto strongCon = q.top();
-            int nab = strongCon.idx;
-            q.pop();
-            while (visited[nab] && !q.empty()) {
-                strongCon = q.top();
-                nab = strongCon.idx;
-                q.pop();
-            }
-            // Call dfsq reflexively on strongest connection in queue
-            if (!visited[nab])
-                dfsq((*transGraph)[nab],q,nab,master,w,maxNode,visited,cnode,edges);
-        }
-    } else {
-        q = std::priority_queue<WgtIdx>();
-    }
-
-    // Add connection between v and nab if v and nab are not merged
-    col = row.begin();
-    for (; col != row.end(); ++col) {
-        int nab = col.index();
-        if (map_to_coarse_[v]!=map_to_coarse_[nab]) {
-            edges.push_back({v,nab,(*transGraph)[v][nab]});
-        }
-    }
-}
 
 template<typename Grid>
 void CoarseGraphOfGrid<Grid>::mergeWellCellsForCoarseGraph(std::vector<int>& hasWell,
@@ -93,7 +40,7 @@ void CoarseGraphOfGrid<Grid>::mergeWellCellsForCoarseGraph(std::vector<int>& has
     for (const auto& well : wellConn) {
 
         bool cellInMultWells = false;
-        std::unordered_set<int> otherWell;
+        std::set<int> otherWell;
         std::vector<int> perfs;
 
         // Loop over all connections in the well. Create list perf of connections
@@ -117,7 +64,7 @@ void CoarseGraphOfGrid<Grid>::mergeWellCellsForCoarseGraph(std::vector<int>& has
         }
         // If current well intersects with other wells. Add the well connections together
         if (cellInMultWells) {
-            int minOtherWell = * std::min_element(otherWell.begin(),otherWell.end());
+            int minOtherWell = *otherWell.begin();
             for (int idx : well) {
                 if (hasWell[idx]!=minOtherWell) {
                     hasWell[idx] = minOtherWell;
@@ -145,16 +92,20 @@ void CoarseGraphOfGrid<Grid>::mergeWellCellsForCoarseGraph(std::vector<int>& has
 }
 
 template<typename Grid>
-void CoarseGraphOfGrid<Grid>::dfsqw(Row row, std::priority_queue<WgtIdx> &q, int v, int master,
-                              double w, int maxNode, std::vector<bool>& visited,
-                              std::vector<int>& cnode, std::vector<std::tuple<int,int,double> >& edges,
-                              const std::vector<int>& hasWell, const std::vector<std::vector<int>>& wellPerf)
+void CoarseGraphOfGrid<Grid>::dfsqw(const Row& row, std::priority_queue<WgtIdx> &q, int v, int master,
+                                    double w, int maxNode, std::vector<bool>& visited,
+                                    std::vector<std::vector<std::tuple<int,int,double>>>& gEdges,
+                                    const std::vector<int>& hasWell, const std::vector<std::vector<int>>& wellPerf)
 {
+
+    auto& current_cnode = coarseNodes.back();
+    auto& current_edges = gEdges.back();
+
     std::vector<int> wellIdxs;
     if (hasWell[v] == -1) {
         visited[v] = true;
         map_to_coarse_[v] = master;
-        cnode.push_back(v);
+        current_cnode.push_back(v);
 
         // Add all neighboring vertices of v with transmissibility larger than w to the queue.
         auto col = row.begin();
@@ -168,14 +119,14 @@ void CoarseGraphOfGrid<Grid>::dfsqw(Row row, std::priority_queue<WgtIdx> &q, int
             }
         }
     } else {
-        // If node has a well, murge all connections to  master node.
+        // If node has a well, merge all connections to  master node.
         int wellId = hasWell[v];
-        std::vector<int> perfs = wellPerf[wellId];
+        const std::vector<int>& perfs = wellPerf[wellId];
 
         for (const auto& idx : perfs) {
             visited[idx] = true;
             map_to_coarse_[idx] = master;
-            cnode.push_back(idx);
+            current_cnode.push_back(idx);
             wellIdxs.push_back(idx);
         }
         for (const auto& idx : perfs) {
@@ -194,10 +145,10 @@ void CoarseGraphOfGrid<Grid>::dfsqw(Row row, std::priority_queue<WgtIdx> &q, int
     }
 
     // Only merge more vertices if the current coarse node is smaller than maxNode.
-    if ( (int)cnode.size() < maxNode ) {
+    if ( (int)current_cnode.size() < maxNode ) {
         if (!q.empty()) {
 
-            // Find strongest connection in queue q not already merged to cnode.
+            // Find strongest connection in queue q not already merged to current_cnode.
             auto strongCon = q.top();
             int nab = strongCon.idx;
             q.pop();
@@ -208,7 +159,7 @@ void CoarseGraphOfGrid<Grid>::dfsqw(Row row, std::priority_queue<WgtIdx> &q, int
             }
             // Call dfsq reflexively on strongest connection in queue
             if (!visited[nab])
-                dfsqw((*transGraph)[nab],q,nab,master,w,maxNode,visited,cnode,edges,hasWell,wellPerf);
+                dfsqw((*transGraph)[nab],q,nab,master,w,maxNode,visited,gEdges,hasWell,wellPerf);
         }
     } else {
         q = std::priority_queue<WgtIdx>();
@@ -222,7 +173,7 @@ void CoarseGraphOfGrid<Grid>::dfsqw(Row row, std::priority_queue<WgtIdx> &q, int
             for (; col != wrow.end(); ++col) {
                 int nab = col.index();
                 if (map_to_coarse_[v]!=map_to_coarse_[nab]) {
-                    edges.push_back({v,nab,(*transGraph)[idx][nab]});
+                    current_edges.push_back({v,nab,(*transGraph)[idx][nab]});
                 }
             }
         }
@@ -232,7 +183,7 @@ void CoarseGraphOfGrid<Grid>::dfsqw(Row row, std::priority_queue<WgtIdx> &q, int
         for (; col != row.end(); ++col) {
             int nab = col.index();
             if (map_to_coarse_[v]!=map_to_coarse_[nab]) {
-                edges.push_back({v,nab,(*transGraph)[v][nab]});
+                current_edges.push_back({v,nab,(*transGraph)[v][nab]});
             }
         }
     }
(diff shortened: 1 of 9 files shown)


blattms:
jenkins build this opm-simulators=7216 please

michal-toth:
I am reviewing the PR and I will need more time to digest the algorithm but here are two little things that could be improved:

- The year in the license/preamble of the new files should be updated to 2026
- Check the input and output arguments. Many vectors are passed by a copy and constant reference would be better.

opm/grid/CoarseGraphOfGrid.cpp
```diff
@@ -0,0 +1,314 @@
...
+#include <config.h>
+#include "CoarseGraphOfGrid.hpp"
+
+#include <numeric>
+
+
+namespace Opm {
+
+template<typename Grid>
+void CoarseGraphOfGrid<Grid>::dfsq(Row row, std::priority_queue<WgtIdx> &q, int v, int master,
```
michal-toth: What is the type Row? I thought that it is a pointer but its methods are accessible via dot. Taking it by a const reference might be better.

opm/grid/CoarseGraphOfGrid.cpp
```diff
@@ -0,0 +1,314 @@
...
+            int nab = strongCon.idx;
+            q.pop();
+            while (visited[nab] && !q.empty()) {
+                strongCon = q.top();
+                nab = strongCon.idx;
+                q.pop();
+            }
+            // Call dfsq reflexively on strongest connection in queue
+            if (!visited[nab])
+                dfsq((*transGraph)[nab],q,nab,master,w,maxNode,visited,cnode,edges);
```
michal-toth: Try rewriting the recursion into an iterative algorithm. It seems like the recursion can run quite deep, and it manipulates the same objects. The hierarchy of recursion is not necessary so the iterative form can save a lot.

opm/grid/CoarseGraphOfGrid.cpp
```diff
@@ -0,0 +1,314 @@
...
+            }
+            else {
+                if (hasWell[idx]!=-1) {
+                    otherWell.insert( hasWell[idx] );
+                }
+            }
+        }
+        // If current well intersects with other wells. Add the well connections together
+        if (cellInMultWells) {
+            int minOtherWell = * std::min_element(otherWell.begin(),otherWell.end());
```
michal-toth: Use a normal (ordered) set for `otherWell`. Finding the minimum will be then trivial.

opm/grid/CoarseGraphOfGrid.cpp
```diff
@@ -0,0 +1,314 @@
...
+        }
+        else {
+            wellPerf.push_back(perfs);
+            wellId++;
+        }
+    }
+}
+
+template<typename Grid>
+void CoarseGraphOfGrid<Grid>::dfsqw(Row row, std::priority_queue<WgtIdx> &q, int v, int master,
```
michal-toth: `dfsqw` and `dfsq` are very similar and it would be great if it was possible to keep just one version. What happens when `dfsqw` gets input corresponding to zero wells?

opm/grid/CoarseGraphOfGrid.cpp
```diff
@@ -0,0 +1,314 @@
...
+            if ( wgt > w) {
+                if (!visited[nab]) {
+                    q.push({wgt, nab});
+                }
+            }
+        }
+    } else {
+        // If node has a well, murge all connections to  master node.
+        int wellId = hasWell[v];
+        std::vector<int> perfs = wellPerf[wellId];
```
michal-toth: const&

opm/grid/CoarseGraphOfGrid.cpp
```diff
@@ -0,0 +1,314 @@
...
+                newV++;
+                gEdges.push_back(edges);
+                coarseNodes.push_back(cnode);
+                if ((int)cnode.size() > biggest)
+                    biggest = cnode.size();
+            }
+        }
+        std::cout << "Coarse partitioning graph size: " << coarseNodes.size() <<" Largest node: "<< biggest << std::endl;
+
+        for (std::vector<std::tuple<int,int,double> > es : gEdges ) {
```
michal-toth: make `es` const reference and `fe` as well

opm/grid/CoarseGraphOfGrid.cpp
```diff
@@ -0,0 +1,314 @@
...
+    int biggest = 0;
+
+    if (rank == 0) {
+        for (int v = 0; v < N; ++v) {
+
+            if (!visited[v]) {
+
+                std::priority_queue<WgtIdx> q;
+                c2f.push_back(v);
+                std::vector<int> cnode;
```
michal-toth: Both `cnode` and `edges` are filled in one function and pushed to the `coarseNodes` and `gEdges`. We can just pass `coarseNodes` and `gEdges` to the function and avoid copying.

opm/grid/CoarseGraphOfGrid.hpp
```diff
@@ -0,0 +1,155 @@
...
+
+namespace Opm {
+
+struct WgtIdx {
+
+    double wgt;
+    int idx;
+
+    bool operator<(const WgtIdx& other) const {
+    return wgt < other.wgt;
```
michal-toth: indentation

opm/grid/CoarseGraphOfGrid.cpp
```diff
@@ -0,0 +1,314 @@
...
+            }
+        }
+        std::cout << "Coarse partitioning graph size: " << coarseNodes.size() <<" Largest node: "<< biggest << std::endl;
+
+        for (std::vector<std::tuple<int,int,double> > es : gEdges ) {
+            std::map<int, double> ce;
+            for (std::tuple<int,int,double> fe : es) {
+                int coarseNab = map_to_coarse_[std::get<1>(fe)];
+                double transVal = std::get<2>(fe);
+                double weight = edgeWeightMethod == 0 ? 1.0 : transVal;
```
michal-toth, line 272: What about logarithmic edge weights `logTransEdgeWgt`? :)
I also implemented them last in GraphOfGrid. I recommend throwing if the `edgeWeightsMethod` is not 0 or 1, it is better to get an exception then running an unexpected/irrelevant branch of the code.

opm/grid/GraphOfGridWrappers.cpp
```diff
@@ -152,6 +152,112 @@ void setGraphOfGridZoltanGraphFunctions(Zoltan_Struct *zz,
         Zoltan_Set_Edge_List_Multi_Fn(zz, getGraphOfGridEdgeList, pGraph);
     }
 }
+
+int getCoarseGraphNumVertices(void* pGraph, int *err)
```
michal-toth: The function is copy-pasted version for GraphOfGrid. Make it a template and let it accept either type.

opm/grid/GraphOfGridWrappers.cpp
```diff
@@ -152,6 +152,112 @@ void setGraphOfGridZoltanGraphFunctions(Zoltan_Struct *zz,
...
+                                ZOLTAN_ID_PTR gIDs,
+               [[maybe_unused]] ZOLTAN_ID_PTR lIDs,
+                                int weightDim,
+                                float *objWeights,
+                                int *err)
+{
+    assert(dimGlobalID==1); // ID is a single int
+    assert(weightDim==1); // vertex weight is a single float
+    const CoarseGraphOfGrid<Dune::CpGrid>& gog = *static_cast<const CoarseGraphOfGrid<Dune::CpGrid>*>(pGraph);
+    const std::vector<std::vector<int>> cnodes = gog.getCoarseNodes();
```
michal-toth: Reference!

opm/grid/GraphOfGridWrappers.cpp
```diff
@@ -152,6 +152,112 @@ void setGraphOfGridZoltanGraphFunctions(Zoltan_Struct *zz,
...
+    int i=0;
+    for (const auto& v : cnodes)
+    {
+        gIDs[i] = i;
```
michal-toth, line 151: Can you guarantee that `gIDs[i]` is always `i`? It was not the case with GraphOfGrid.

opm/grid/GraphOfGridWrappers.cpp
```diff
@@ -152,6 +152,112 @@ void setGraphOfGridZoltanGraphFunctions(Zoltan_Struct *zz,
...
+           [[maybe_unused]] int numCells,
+                            ZOLTAN_ID_PTR gIDs,
+           [[maybe_unused]] ZOLTAN_ID_PTR lIDs,
+                            int *numEdges,
+                            int *err)
+{
+    assert(dimGlobalID==1); // ID is a single int
+    const CoarseGraphOfGrid<Dune::CpGrid>& gog = *static_cast<const CoarseGraphOfGrid<Dune::CpGrid>*>(pGraph);
+
+    const std::vector<std::map<int, double> > edges = gog.getCoarseEdges();
```
michal-toth: reference

opm/grid/GraphOfGridWrappers.cpp
```diff
@@ -152,6 +152,112 @@ void setGraphOfGridZoltanGraphFunctions(Zoltan_Struct *zz,
...
+                            ZOLTAN_ID_PTR nborGIDs,
+                            int *nborProc,
+                            int weightDim,
+                            float *edgeWeights,
+                            int *err)
+{
+    assert(dimGlobalID==1); // ID is a single int
+    assert(weightDim==1); // edge weight is a single float
+    const CoarseGraphOfGrid<Dune::CpGrid>&  gog = *static_cast<const CoarseGraphOfGrid<Dune::CpGrid>*>(pGraph);
+    const std::vector<std::map<int, double> > edges = gog.getCoarseEdges();
```
michal-toth: reference

opm/grid/GraphOfGridWrappers.cpp
```diff
@@ -152,6 +152,112 @@ void setGraphOfGridZoltanGraphFunctions(Zoltan_Struct *zz,
...
+    int id=0;
+
+    for (const auto& node : edges)
+    {
+        for (const auto& edge : node) {
+            nborGIDs[id] = edge.first;
+            edgeWeights[id] = edge.second;
+            nborProc[id++] = 0;
```
michal-toth: Does this assume that the partitioning is done on rank 0? Please leave a comment in the code.

opm/grid/GraphOfGridWrappers.cpp
```diff
@@ -152,6 +152,112 @@ void setGraphOfGridZoltanGraphFunctions(Zoltan_Struct *zz,
...
+        }
+    }
+    
+    *err = ZOLTAN_OK;
+}
+
+template<typename Zoltan_Struct>
+void setCoarseGraphZoltanGraphFunctions(Zoltan_Struct *zz,
```
michal-toth: Identical to `setGraphOfGridZoltanGraphFunctions`. Template it instead of copy-paste.

opm/grid/GraphOfGridWrappers.cpp
```diff
@@ -301,6 +407,97 @@ extendRootExportList(const GraphOfGrid<Dune::CpGrid>& gog,
     return exportedCells;
 }
 
+std::vector<std::vector<int>>
+extendRootExportList(const CoarseGraphOfGrid<Dune::CpGrid>& gog,
```
michal-toth: Template is better than copy-pasting.

opm/grid/GraphOfGridWrappers.cpp
```diff
@@ -491,6 +688,128 @@ makeImportAndExportLists(const GraphOfGrid<Dune::CpGrid>& gog,
...
+        }
+    }
+    std::vector<int> newImportList;
+    int newNumImport;
+    if (cc.rank() == root) {
+        std::vector<MPI_Request> requestSize(2 * (cc.size() - 1));
+        
+        for (int i = 0; i < cc.size() - 1; ++i) {
+            int ii = i + (int)(i >= root); // ii takes values {0,...,mpisize-1} but skips root
+            int tag = 15; // a random number
```
michal-toth, line 658: I am happy to see that the random number was changed when the code was copied. (Btw it is the tag for the MPI to distinguish and pair correctly two communications with the same signature.)

opm/grid/GraphOfGridWrappers.cpp
```diff
@@ -491,6 +688,128 @@ makeImportAndExportLists(const GraphOfGrid<Dune::CpGrid>& gog,
...
+        }
+        else {
+            auto wellRanks = getWellRanks(gIDtoRank, wellConnections);
+            parallel_wells = wellsOnThisRank(*wells, wellRanks, cc, root);
+        }
+    }
+    else
+    {
+        std::sort(myExportList.begin(), myExportList.end());
+        std::sort(myImportList.begin(), myImportList.end());
```
michal-toth: std::ranges::sort is preferred

opm/grid/CoarseGraphOfGrid.cpp
```diff
@@ -0,0 +1,314 @@
...
+    const auto& rank = grid.comm().rank();
+
+    std::vector<bool> visited(N, false);
+    std::vector<int> hasWell(N,-1);
+    std::vector<std::vector<int>> wellPerf;
+    if (!allowDistributedWells) {
+        mergeWellCellsForCoarseGraph(hasWell, wellPerf, wellConn);
+    }
+    map_to_coarse_.resize(N, -1);
+    std::vector<int> c2f;
```
michal-toth: This variable is unused. Well, `v` is pushed into it but it is not utilized.

michal-toth (changes requested):
Regarding code:
There is also a lots of copy-pasting, so the code looks more like a draft with a pending refactorization. (Being a draft can make this review sound harsher, I am sorry.) I would also welcome more comments in the code. (E.g. method `dfsq` was fine but `createCoarseGraph` is completely commentless.)

For now, unit tests cover only uniform edge weights. Since the vertex grouping is done based on the greatest transmissibility (and all transmissibilities are equal), the priority queue `q` (in `dfsq`) is never filled and a big part of the code is not tested.

Regarding the algorithm:
I am looking forward to the new partitioning option. I did not test it yet but for now my feeling is that picking the right seed nodes for the cell grouping can make a big difference. I would welcome an option to use edge weights rescaled to the node instead of using transmissibilities. (E.g. by dividing each node's edge weights by its lowest edge weight. Big edge weight would then mean that two cells are coupled more strongly to themself than to their neighbors. The edge weights would not be symmetric anymore but it would be more fair in the graphs where one area has higher transmissibilities than other.)

opm/grid/GraphOfGridWrappers.cpp
andrthu: Yes, this is how it should work. Merging is done differently in CoarseGraphOfGrid and GraphOfGrid

andrthu:
I believe I have implemented most of the suggestions. The exceptions are logarithmic edge weights and rewriting the recursion to an iterative algorithm.

>For now, unit tests cover only uniform edge weights. Since the vertex grouping is done based on the greatest transmissibility (and all transmissibilities are equal), the priority queue q (in dfsq) is never filled and a big part of the code is not tested.

I am a bit confused about this. You are correct that all tests use uniform edge weights, but this is only relevant for setting the edge weights in the coarse graph. For creating the coarse graph by merging nodes, transmissibility is always used to determine strong connections.

opm/grid/CoarseGraphOfGrid.cpp
andrthu: You are right, dfsqw works as dfsq when the wellPerf is empty.

opm/grid/GraphOfGridWrappers.cpp
andrthu: It is now done on root.

michal-toth:
> You are correct that all tests use uniform edge weights, but this is only relevant for setting the edge weights in the coarse graph. For creating the coarse graph by merging nodes, transmissibility is always used to determine strong connections.

I assumed that the edge weights and transmissibilities are the same thing in this case, i.e. using uniform edge weights is equivalent to set all transmissibilities to 1. Consequently, the queue `q` in `dfsq` would never get longer because it could not find a stronger connection.

atgeirr:
> I assumed that the edge weights and transmissibilities are the same thing in this case

It is very easy to make that assumption! When discussing weights please make sure to distinguish between the two stages: coarsening (always transmissibility based) and subsequent partitioning (uniform or transmissibility based, weights between nodes are in all cases the sum of weights of connecting fine-level weights).

PR #1051, opened 2026-07-22
MinpvProcessor: guard merge branch against out-of-bounds zcorn access

hnil:
Bug. In MinpvProcessor::process, the mergeMinPVCells branch reads and writes getCellZcorn(ii, jj, kk_iter, ...) without the bottom-of-grid guard that the NNC branch already has. When a column of low-pore-volume cells runs off the grid bottom, the inner column walk exits with kk_iter == dims_[2] and the void-fill then accesses 8 doubles past the end of the zcorn array — both a read and a write. The merge is also performed when the receiving bottom cell is inactive.

Fix. Two early-outs mirroring the NNC branch: skip when the column ran off the bottom, and skip when the receiving cell is inactive.

Test. MergeColumnRunsOffGridBottom in tests/test_minpvprocessor.cpp places sentinel padding directly behind the zcorn data and asserts it is untouched, so the out-of-bounds write is detected deterministically rather than relying on a crash.

Verified on this branch: with the fix reverted the new case produces 5 failures (wrong removed_cells count plus 4 clobbered sentinels) while all 5 pre-existing cases still pass; with the fix, all 6 pass.

### DIFF
--- a/opm/grid/MinpvProcessor.cpp
+++ b/opm/grid/MinpvProcessor.cpp
@@ -228,16 +228,18 @@ MinpvProcessor::process(const std::vector<double>& thickness,
                         low_pv_active = pv[c_below] < minpvv[c_below] && active;
                     }
 
+                    // The column ran off the grid bottom: the loop above left the last
+                    // cell via the kk_iter == dims_[2] break, and that break only exits
+                    // the inner while, not this column's for-loop.  There is no cell
+                    // below to receive the void or to connect by an NNC, and c_below
+                    // is stale, so neither branch can proceed.  Without this guard the
+                    // merge branch reads and writes 8 doubles past the end of zcorn.
+                    if (kk_iter == dims_[2]) {
+                        break;
+                    }
+
                     // create nnc if false or merge the cells if true
                     if (mergeMinPVCells && c_low_pv_active) {
-                        // Column ran off the grid bottom: no cell below to receive
-                        // the void.  Without this guard the zcorn access at
-                        // kk_iter == dims_[2] reads and writes 8 doubles past the
-                        // end of the zcorn array.
-                        if (kk_iter == dims_[2]) {
-                            break;
-                        }
-
                         // Bottom cell inactive: leave the geometry untouched, no merge.
                         if (!actnum.empty() && !actnum[c_below]) {
                             kk = kk_iter;
@@ -256,8 +258,9 @@ MinpvProcessor::process(const std::vector<double>& thickness,
                     else
                     {
 
-                        // No top or bottom cell, so no nnc is created.
-                        if (kk == 0 || kk_iter == dims_[2]) {
+                        // No top cell, so no nnc is created.  (The no-bottom-cell case,
+                        // kk_iter == dims_[2], is handled by the guard above.)
+                        if (kk == 0) {
                             kk = kk_iter;
                             continue;
                         }
--- a/opm/grid/MinpvProcessor.cpp
+++ b/opm/grid/MinpvProcessor.cpp
@@ -235,8 +235,7 @@ MinpvProcessor::process(const std::vector<double>& thickness,
                         // kk_iter == dims_[2] reads and writes 8 doubles past the
                         // end of the zcorn array.
                         if (kk_iter == dims_[2]) {
-                            kk = kk_iter;
-                            continue;
+                            break;
                         }
 
                         // Bottom cell inactive: leave the geometry untouched, no merge.
--- a/opm/grid/MinpvProcessor.cpp
+++ b/opm/grid/MinpvProcessor.cpp
@@ -230,6 +230,21 @@ MinpvProcessor::process(const std::vector<double>& thickness,
 
                     // create nnc if false or merge the cells if true
                     if (mergeMinPVCells && c_low_pv_active) {
+                        // Column ran off the grid bottom: no cell below to receive
+                        // the void.  Without this guard the zcorn access at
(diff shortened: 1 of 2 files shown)


opm/grid/MinpvProcessor.cpp
```diff
@@ -230,6 +230,21 @@ MinpvProcessor::process(const std::vector<double>& thickness,
...
                     // create nnc if false or merge the cells if true
                     if (mergeMinPVCells && c_low_pv_active) {
+                        // Column ran off the grid bottom: no cell below to receive
+                        // the void.  Without this guard the zcorn access at
+                        // kk_iter == dims_[2] reads and writes 8 doubles past the
+                        // end of the zcorn array.
+                        if (kk_iter == dims_[2]) {
+                            kk = kk_iter;
+                            continue;
+                        }
```
blattms: Thanks a lot.

I think this code  should be moved outside of the surrounding if-statement to line 230.

You should also add a clear comment why this is needed: because the situation occurs when we reach the condition in line 221 and we only step outside of the inner while and not the surrounding for-loop.

I also think the code would be much clearer if this was just a 
```suggestion
                        // Column ran off the grid bottom: no cell below to receive
                        // the void.  Without this guard the zcorn access at
                        // kk_iter == dims_[2] reads and writes 8 doubles past the
                        // end of the zcorn array.
                        if (kk_iter == dims_[2]) {
                            break;
                        }
```

blattms (changes requested):

atgeirr:
Just a note on the commit message of the last commit: I think it is better to avoid using the github handle with the @ in the commit message, as it may trigger a lot of alerts for Markus whenever the commit is part of a sequence of events, so it may be better to just use the full name. (Based on something that happened years ago, so maybe github fixed it so that it is not an annoyance anymore.)

blattms:
FYI, only part of my suggestions were applied

blattms:
Please also change the label. This is a bug fix and not irrelevant (at least if you do not run with the default options).

hnil:
You were right, only the `break` had been applied. The guard is now hoisted out of the `if (mergeMinPVCells && c_low_pv_active)` so it covers the NNC branch too, with a comment explaining why the case is reachable: the inner `while` exits via its own `kk_iter == dims_[2]` break, and that leaves only the inner loop, not the enclosing column loop.

With the check hoisted, the NNC branch's `kk_iter == dims_[2]` disjunct is dead, so that condition reduces to `kk == 0`. This is behaviour-preserving: the outer loop is `for (int kk = 0; kk < dims_[2]; ++kk)`, so the branch's previous `kk = kk_iter; continue;` with `kk_iter == dims_[2]` already ended the loop -- exactly what the hoisted `break` now does.

Verified: all 6 cases in `test_minpvprocessor` pass; with the fix reverted the new case produces 5 failures (wrong `removed_cells` count plus 4 clobbered sentinels), so the sentinel padding really does catch the out-of-bounds write.

Label changed to `manual:bugfix` as requested.

On the commit message: good point, and it applies beyond this PR -- no GitHub handles in commit messages from now on. The commit is reworded to use the full name; the `Co-authored-by:` trailer is unchanged and the tree is byte-identical.

tests/test_minpvprocessor.cpp
```diff
@@ -388,3 +388,45 @@ BOOST_AUTO_TEST_CASE(Processing)
...
+                                  1, 1, 1, 1,
+                                  2, 2, 2, 2,
+                                  2, 2, 2, 2,
+                                  3, 3, 3, 3 };
+    const std::size_t ncorn = zcorn.size();
+    const double sentinel = -999.25;
+    std::vector<double> padded = zcorn;
+    padded.resize(ncorn + 8, sentinel); // guard region behind zcorn
+
+    std::vector<double> pv = { 0.1, 0.1, 0.1 };
```
blattms: Should we maybe do two more tests:
1. First cell stays
2. Second cell stays (maybe with 4 cells)

blattms (commented):
thanks a lot for the changes and sorry for the late comment:

hnil:
Both cases added — first cell retained, and a retained cell in the middle of a four-cell column, so the walk reaches the bottom from a different starting depth each time.

Checked they discriminate: with the guard removed, 3 of 8 cases fail; with it, all 8 pass.

tests/test_minpvprocessor.cpp
```diff
@@ -388,3 +390,129 @@ BOOST_AUTO_TEST_CASE(Processing)
...
+            for (std::size_t i = ncorn; i < padded.size(); ++i) {
+                BOOST_CHECK_EQUAL(padded[i], sentinel);
+            }
+        }
+
+        double sentinel;
+        std::vector<double> zcorn{};
+        std::size_t ncorn{};
+        std::vector<double> padded{};
+    };
```
blattms: "People" change their mind to often. Please use one way to add the padding in all tests. Adding multiple ways just increases the complexity for no reason.

blattms (changes requested):
The last commit added too much complexity. 

I also noted that the new tests only are tests for the bug that was found. I think this might not be good enough. They should actually test all of the expected outcome (other zcorn values, actnum, etc.)

hnil:
Fair — all three off-the-bottom tests now build the padding the same way, through one `PaddedColumn` helper taking the column height. The inline version in the first test is gone.

blattms (approved):

blattms:
jenkins build this please

PR #1054, opened 2026-07-30
bump required dune to 2.11

akva2:
Downstream of https://github.com/OPM/opm-common/pull/5255


CMakeLists.txt
```diff
@@ -43,14 +43,14 @@ macro(opm-grid_prereqs_hook)
     target_link_libraries(opmgrid PUBLIC MPI::MPI_C MPI::MPI_CXX)
   endif()
 
-  target_link_libraries(opmgrid PUBLIC dunecommon dunegrid)
+  target_link_libraries(opmgrid PUBLIC Dune::Common Dune::Grid)
```
SoilRos: At this point, `Dune::Grid` implies `Dune::Common`, if not let me know because something went wrong then.

akva2: Right. Not taken those things into account but sure I guess we can rely on the implicit.

PR #1056, opened 2026-07-31
CpGrid: build point communication interfaces from all points of a cell

hnil:
On a faulted corner-point grid the hanging nodes along the fault are not canonical corners of the neighbouring cells, so point interfaces built from the 8 canonical corners never include them: they receive no message and vertex ownership is not a partition of unity across ranks (breaks vertex-based fields, e.g. mechanics).

First commit adds a test that fails on master (np 2/4); second builds the interfaces from all points of a cell, after which np 1/2/4 pass and `distribution_test` is unchanged.

One invariant worth stating: the per-cell point rows are traversal-order deduplicated, deliberately not sorted — the interface pairs sender/receiver rows positionally across ranks, and local point indices differ per rank, so sort+unique would silently reintroduce the bug.


opm/grid/cpgrid/CpGridData.cpp
```diff
@@ -1697,10 +1702,40 @@ void CpGridData::computeCommunicationInterfaces([[maybe_unused]] int noExistingP
...
+        const int nf = faces.size();
+        for (int f = 0; f < nf; ++f) {
+            for (const auto& fv : face_to_point_[faces[f].index()]) {
+                if (std::find(points.begin(), points.end(), fv) == points.end()) {
+                    points.push_back(fv);
+                }
+            }
+        }
+        cell_to_allpoint_.appendRow(points.begin(), points.end());
+    }
```
blattms, line 1740: I think the best place to do this is during [`computeFace2Point`](https://github.com/hnil/opm-grid/blob/99aa1cc1f1761123ba5d75e3f94d4cde8d108ea9/opm/grid/cpgrid/CpGridData.cpp#L1241) namely during the last `grid.scatterData` which uses a  `FaceViaCellHandleWrapper<SparseTableDataHandle>`

One possibility would be to optionally make SparseTableDataHandle populate your container with the values during its scatter method.

opm/grid/cpgrid/CpGridData.cpp
```diff
@@ -1928,7 +1963,7 @@ bool CpGridData::preAdapt()
             if (local_empty)
                 mark_.resize(size(0));
         }
-
+
```
blattms: We should not add trailing whitespace

opm/grid/cpgrid/CpGridData.hpp
```diff
@@ -775,6 +775,11 @@ class CpGridData
     cpgrid::OrientedEntityTable<1, 0> face_to_cell_;
     /** @brief Container for the lookup of the points for each face. */
     Opm::SparseTable<int>             face_to_point_;
+    /** @brief All points of each cell, i.e. every point of every face of the
+     *   cell, not only the eight canonical corners.  On corner-point grids
+     *   with hanging nodes the extra points are essential for building
+     *   complete codim-3 communication interfaces. */
+    Opm::SparseTable<int>             cell_to_allpoint_;
```
blattms: This is not really used after the set up and should be a local variable in some function

opm/grid/cpgrid/DataHandleWrappers.hpp
```diff
@@ -148,7 +148,7 @@ template<class Handle>
 struct PointViaCellHandleWrapper : public PointViaCellWarner
 {
     using DataType = typename Handle::DataType;
-    using C2PTable = std::vector< std::array<int,8> >;
+    using C2PTable = Opm::SparseTable<int>;
```
blattms: I don't think we even use the PointViaCellHandle. So maybe remove it?

blattms (dismissed):
I need to think about this a little. Did not look at the tests. So this is a first assessement. More to follow,

It is nice to have this fixed. We never needed communication between points before, hence this surely needed some love.

I think we should make sure that we do not add permanent addtional memory usage and maybe we make this more efficient, even.

hnil:
Thanks — three of four done, and the memory objection is fully addressed.

- **No persistent table.** `cell_to_allpoint` is now a local in the function that builds the point interfaces; nothing is retained after set-up.
- **Cheaper, too.** The duplicate check is a per-cell marker array instead of `std::find`, so the rows build in linear rather than quadratic time in a cell's points. Traversal order is unchanged — that is the invariant the positional gather/scatter depends on.
- **`PointViaCellHandleWrapper` left alone.** You are right that nothing uses it — I checked, it has no callers anywhere in opm-grid. With the table local, my typedef change was unnecessary, so it is reverted and the struct is untouched. Removing it entirely seems like its own cleanup; happy to do that separately if you want it.
- Trailing whitespace gone.

Not done: building this during `computeFace2Point`'s last `scatterData`. That is the better home and I would rather you confirm the shape first — making `SparseTableDataHandle` optionally populate a cell-indexed container during scatter means threading that container through `computeFace2Point`'s signature, which touches the distribution path rather than just this feature. Say the word and I will do it.

Rebased onto master (after #1051). np 1/2/4 pass, full opm-grid suite 84/84.

hnil:
What do I do with this @blattms

opm/grid/cpgrid/DataHandleWrappers.hpp
hnil: Removed. Checked first that nothing uses it - not opm-grid, not opm-simulators, and not opm-flowgeomechanics - and `PointViaCellWarner` existed only as its base, so that and the translation unit holding the warning went too. 113 lines. All 84 opm-grid tests pass.

hnil:
jenkins build this please

opm/grid/cpgrid/CpGridData.cpp
blattms: This comment has not been adressed. As outlined there should be no need to create an extra large vector and we need to prevent this.

blattms (changes requested):

blattms:
Unfortunately, @bska has dismissed my old review. (Not sure since when we are doing that and what that even is). Please recover it as otherwise my comment is missing context.

blattms:
>  No persistent table. cell_to_allpoint is now a local in the function that builds the point interfaces; nothing is retained after set-up.

My point was not make this a local variable, but to prevent it using the other communication that I mentioned.

opm/grid/cpgrid/CpGridData.cpp
hnil: Is it written as you want now?

hnil: It is only tested on the tests. Not in real simulations with mechanics, but hopefully it is ok.

hnil: New version checked at on quite some mpi ranks on norne.

opm/grid/cpgrid/CpGridData.cpp
```diff
@@ -1697,10 +1759,12 @@ void CpGridData::computeCommunicationInterfaces([[maybe_unused]] int noExistingP
...
     std::vector<std::map<int,char> >().swap(face_attributes);
     */
+    // All points of each cell, not only its eight corners: with hanging nodes a face
+    // can hold points that are no corner of the neighbouring cell.
     std::vector<std::map<int,char> > point_attributes(noExistingPoints);
-    AttributeDataHandle<std::vector<std::array<int,8> > >
+    CellPointAttributeDataHandle
         point_handle(ccobj_.rank(), *partition_type_indicator_,
-                     point_attributes, cell_to_point_, *this);
+                     point_attributes, cell_to_face_, face_to_point_, *this);
```
blattms, line 1767: I have been looking at this for quite a while now.

The communication below (line 1771) is communicating from an undistributed grid (were there is only an owner partition) to distributed grid (that should actually have more partitions).

It is probably me, but I cannot see how the latter can be the case as the scatter method seems to just write the original value. What am I seeing wrong?

A point can only be interior on one process.

So my question is: Are there any points in the distributed grid that do not belong to the interior partition?

tests/cpgrid/partition_of_unity_test.cpp
```diff
@@ -0,0 +1,220 @@
...
+    grid.loadBalance(parts, /* ownersFirst = */ false,
+                     /* addCornerCells = */ true, /* overlapLayers = */ 1);
+
+    const int numVertices = grid.leafIndexSet().size(3);
+    std::vector<int> received(numVertices, 0);
+
+    VertexReachedHandle handle(grid, received, rank);
+    grid.communicate(handle, Dune::All_All_Interface,
+                     Dune::ForwardCommunication);
+
```
blattms, line 188: I think you should add other additional  test here as ALL_ALL_Interface ignores the partition type:
- Check that there are points in the distributed grid that are not interior
- add communication using Overlap_OverlapFront_Interface  and see whether this does what is expected.

BTW DUNE has a quite extensive test suite for this. If would finally be conformant we would get that for free.

tests/cpgrid/partition_of_unity_test.cpp
```diff
@@ -0,0 +1,220 @@
...
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include <config.h>
+
+#define NVERBOSE
```
blattms, line 22: What is this?

PR #1058, opened 2026-08-10
MinpvProcessor: opt-in thin-cell handling

hnil:
New defaulted parameter `thin_cells_as_minpv` (default `false`, behaviour unchanged). When enabled, active cells thinner than `z_tolerance` are treated like low-pore-volume cells throughout — removal, the column walk, NNC suppression and the gap check — and the merge branch runs for all removed cells, not only low-pv ones.

Every new condition reduces to the plain pore-volume comparison when the option is off, so the default path is unchanged rather than merely equivalent-looking.

`processEclipseFormat` wires it to `edge_conformal`. The requirement comes from the geomech/thermal stack: vertex-based discretizations need a conforming, connected grid.

**Stacked on OPM/opm-grid#1064**, which carries the `processEclipseFormat` half. That part is independent, needed for ordinary edge-conformal grids, and can be reviewed without this one. Until it merges the diff here includes it.

### DIFF
--- a/opm/grid/MinpvProcessor.cpp
+++ b/opm/grid/MinpvProcessor.cpp
@@ -77,7 +77,8 @@ MinpvProcessor::process(const std::vector<double>& thickness,
                         bool pinchOption4ALL,
                         const std::vector<double>& permz,
                         const std::function<double(int)>& multz,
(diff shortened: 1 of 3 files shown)


bska:
jenkins build this please

PR #1061, opened 2026-08-13
cpgrid: replace std::sort on 2-element arrays with explicit swap

akva2:
GCC 14 inlines std::__move_median_to_first (the introsort median-of-3 step) even when the runtime guard 'last - first > 16' is statically false, then fires -Warray-bounds after seeing a pointer at byte offset [8, ...] inside the 8-byte cells[2] array.

Replace both std::sort(cells, cells + N) calls (N <= 2) with a direct conditional swap.  This is semantically identical for N in {1, 2} and avoids instantiating the introsort machinery.

Also reinstate the assert(cellcount == 1 || cellcount == 2) inside the 'if (cellcount > 0)' guard, where it remains true; the original assertion was only commented out because cellcount == 0 became possible for periodic-extension boundary faces.

### DIFF
--- a/opm/grid/cpgrid/processEclipseFormat.cpp
+++ b/opm/grid/cpgrid/processEclipseFormat.cpp
@@ -58,8 +58,8 @@
 
 #include <algorithm>
 #include <array>
+#include <cassert>
 #include <cstddef>
-#include <fstream>
 #include <initializer_list>
 #include <iostream>
 #include <memory>
@@ -1083,7 +1083,9 @@ namespace cpgrid
                 const int c2 = global_to_local[nncpair.second];
                 cells[0].setValue(c1, true);
                 cells[1].setValue(c2, false);
-                std::sort(cells, cells + 2);
+                if (cells[1] < cells[0]) {
+                    std::swap(cells[0], cells[1]);
+                }
                 f2c.appendRow(cells, cells + 2);
                 face_to_output_face.push_back(cpgrid::NNCFace);
             }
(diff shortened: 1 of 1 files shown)


akva2:
jenkins build this please

opm/grid/cpgrid/processEclipseFormat.cpp
```diff
@@ -1160,11 +1162,12 @@ namespace cpgrid
                     }
                 }
 
-                // Assertation below is no longer true, due to periodic_extension etc.
-                // Instead, the appendRow() is put inside an if test.
-                // assert(cellcount == 1 || cellcount == 2);
+                // cellcount == 0 is possible (periodic_extension boundary faces).
                 if (cellcount > 0) {
-                    std::sort(cells, cells + cellcount);
+                    assert(cellcount == 1 || cellcount == 2);
```
bska, line 1167: Tiny nit since we're changing this anyway: Maybe add `<cassert>` for the `assert()` macro?

bska (approved):
Looks good to me.  Just one (very) small issue and then you can merge this.

opm/grid/cpgrid/processEclipseFormat.cpp
akva2: absolutely, thought the agent checked that. also removed the unused fstream include while at it.

akva2:
jenkins build this please

bska:
Thanks a lot for the updates.  I'll merge into master.

PR #1062, opened 2026-08-14
cpgrid: expose throw-only paths to CUDA diagnostics

akva2:
Hand holding for nvcc, see https://github.com/OPM/opm-common/pull/5293

### DIFF
--- a/opm/grid/cpgrid/CpGridData.hpp
+++ b/opm/grid/cpgrid/CpGridData.hpp
@@ -381,6 +381,7 @@ public:
         }
         else {
             OPM_THROW(std::logic_error, "Vertex has no history record.\n");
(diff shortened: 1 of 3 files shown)


opm/grid/cpgrid/CpGridData.hpp
```diff
@@ -381,6 +381,7 @@ class CpGridData
         }
         else {
             OPM_THROW(std::logic_error, "Vertex has no history record.\n");
+            __builtin_unreachable();
```
bska, line 384: Oh, how I long for C++23's [`std::unreachable()`](https://cppreference.com/cpp/utility/unreachable)...

multitalentloes (approved):

PR #1064, opened 2026-08-20
processEclipseFormat: merge removed cells on edge-conformal grids

hnil:
An edge-conformal grid has to stay topologically connected. Cells removed by MINPV must therefore be merged geometrically rather than bridged with NNCs, which would create faces with no geometric counterpart — vertex-based discretizations cannot use those.

So for `edge_conformal` grids this selects `mergeMinPVCells`, asserts that no MINPV, PINCH or explicit NNC survives, and treats the merged columns as active gaps.

Nothing changes unless `edge_conformal` is set.

Split out of #1058 so it can be reviewed on its own: that PR also adds an opt-in thin-cell mode to `MinpvProcessor`, which is a separate question and not needed for ordinary grids. #1058 is now stacked on this one.


bska:
jenkins build this please

bska (approved):

PR #1055, opened 2026-07-30
Return geometry by value (alternative 4)

SoilRos:
This is yet another version of how to return the geometry by value (see #885, #884, #891, and #334). Compared to the others, this should be the fastest one. The trick is made by not sharing the ownership of the corners, just a non-owning pointer. Additionally, since the `geometryInFather` needs to store its own data, I simply changed its type for `Dune::AxisAlignedCubeGeometry` which does the same job, but faster.

PR #1065, opened 2026-08-24
Add test collapsed-corners-cell does not map to entire unit cube 

aritorto:


atgeirr:
I am not sure we want this, I will try to explain why.

The dune geometry concepts assumes that the mapping from reference domain to the real domain is one-to-one, see https://www.dune-project.org/doxygen/master/dune-grid/a05970.html#details for example. The approach in CpGrid deviates from this: we may pretend to have a hexahedron, but due to possible degenerate cells such as in the case you bring here (a wedge-shaped cell with two adjacent degenerate corners) we cannot have a 1-1 mapping from the unit cube.

This has (so far) not been a problem: we do not use this mapping (i.e. local(), global(), the jacobianTransposed() method etc.) in our finite volume code. Anyone trying to use CpGrid with a finite element method that needs this will face a problem if the grid has degenerate cells (I know some things have been done in this direction but I am not familiar with details).

At a degenerate corner, it is not correct to call local(), the mapping does not exist. That is what this test checks though!

(Another somewhat unrelated note on global() that may be relevant to your work: it implements a trilinear mapping, which will not yield matching geometries along faults. A consistent mapping needs to map points on faces to the bilinear surface defined by e.g. the pillars and z-coordinates, as we have discussed earlier.)

aritorto:
@atgeirr A bit late to describe the goal of this PR, which is mainly to help illustrate some of the internal discussions I mentioned in OPM/opm-grid#1055. What OPM/opm-grid#1065 shows might be obvious to everyone, but I wanted to make the current situation a bit more concrete.

I'm very aware that we need to remove the assumption that a cell is a real hexahedron in order to be able to refine it. For now, however, this is still the case in `master`. See, for example, [[here](https://github.com/OPM/opm-grid/blob/e94a7386ca9d62104c329214d28e65f2e2da44dd/opm/grid/cpgrid/Geometry.hpp#L624)](https://github.com/OPM/opm-grid/blob/e94a7386ca9d62104c329214d28e65f2e2da44dd/opm/grid/cpgrid/Geometry.hpp#L624), where the 1-to-1 mapping with the unit cube is assumed when refining a `Geometry`.

I wasn't expecting this to be merged, so apologies for the delayed description!

